PR #51, opened 2014-06-19
support config file in XDG style

keis:
A quick and dirty patch that tries to load the config from the xdg configured location. This is not very useful on its own but I figured the best way to open the discussion on xdg in irssi is by putting some code out there.

More on XDG Base Directory http://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html

### DIFF
--- a/src/core/core.c
+++ b/src/core/core.c
@@ -196,9 +196,12 @@ void core_preinit(const char *path)
 		if (irssi_dir[len-1] == G_DIR_SEPARATOR)
 			irssi_dir[len-1] = '\0';
 	}
-	if (irssi_config_file == NULL)
-		irssi_config_file = g_strdup_printf("%s/"IRSSI_HOME_CONFIG, irssi_dir);
-	else {
+	if (irssi_config_file == NULL) {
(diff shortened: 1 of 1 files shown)


wsdjeg:
I also want irssi support `$XDG_CONFIG_HOME`, as you can see the differences between neovim and vim.

Managor:
Any progress on this issue yet?

josteink:
@Managor: It's probably not the answer you're looking for, but I decided to give [weechat](https://github.com/weechat/weechat) a try, and it supports quite a lot of things irssi doesn't, including this.

dequis:
@josteink in what way does it support this? I have a `~/.weechat` here and googling "weechat xdg" gets me this old issue http://savannah.nongnu.org/task/?10934 which went more or less the same way as every single xdg issue in every project.

FWIW there's always `irssi --home ~/whatever`

josteink:
Oh. Seems you're right and I'm 100% in the wrong. My bad.,

I did type the comment out of memory, and somehow I must have mixed something up.

Consider my comment withdrawn.

PR #57, opened 2014-06-23
Improve tab completion for /topic

vstakhov:
Currently, /topic works rather strange and sometimes it doesn't complete this 
command at all (/t however works fine in those cases). This patch enables the
following logic: complete topic by its first letters, for example if a channel
has a topic named `example` then something like `/topic ex<tab>` is completed
but not `/topic oth<tab>`.


ailin-nemui:
thanks for your contribution, but we need to fix that other issue instead

PR #64, opened 2014-06-26
Some updates for the irssi proxy module

hansnielsen:
I added a few creature comforts to the proxy module.

* Changed the signals slightly so that there is no "proxy client connected" until a user has authenticated with a password
* Made the console output include more useful details and go to the server windows
* Added a toggle to enable / disable the module; on by default to prevent existing user confusion
* Updated documentation

### DIFF
--- a/docs/help/in/proxy.in
+++ b/docs/help/in/proxy.in
@@ -1,0 +1,14 @@
+
+@SYNTAX:proxy@
+
+%9Description:%9
+
+    Displays the list of clients connected to the proxy.
+
+%9Examples:%9
+
+    /PROXY
+    /PROXY STATUS
+
+%9See also:%9 LOAD PROXY, SET irssiproxy
+
--- a/docs/proxy.txt
+++ b/docs/proxy.txt
@@ -12,6 +12,11 @@ In irssi, say:
 
   /LOAD proxy
 
+If you want the proxy to be loaded automatically at startup, add the
+load command to ~/.irssi/startup:
+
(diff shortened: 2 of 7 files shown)


dequis:
@hansnielsen poke, i believe this is just missing some /help stuff as geert pointed out.

hansnielsen:
Whoops, kinda forgot about this. Added the help file and rebased it against current master.

ailin-nemui:
I'd use formats for the proxy information output instead of hard-coded printfs. It would be possible to toggle the proxy by unloading the module, I guess a toggle is also fine. The command seems a bit misleading, maybe have just `/proxy` to list connected clients? I wonder if a `/proxy disconnect ` kind of command would be nice to have.

dequis:
The comments of #148 (which is about the non-bnc proxy feature, i.e. socks5 and similar) suggest a /proxy command, which would conflict with this one. Maybe rename this one to /irssiproxy to match the setting names?

GeertHauwaerts:
any update on this?

dequis:
Well, @hansnielsen seems to have disappeared from github (last public activity in february).

I wouldn't mind taking over and opening a new PR based on his work.

GeertHauwaerts:
That would be awesome, since this PR is almost finished and just requires some minor tweaking :)

dequis:
Did some stuff the other day, not completely done yet, but submitted #259 to keep track of it, at least.

ailin-nemui:
continued in #259

PR #65, opened 2014-06-26
forward alternate_nick to Irc::Server attributes

ailin-nemui:
add the missing alternate_nick in Irc::Server by making an additional
call to the Irc::Connect filler. this is not quite ideal but might
need bigger refactoring otherwise.


dequis:
So, now I understand this code a little bit more, but still don't get why you say "this is not quite ideal"

ailin-nemui:
because we're overwriting attributes once from there and then from here iirc. Maybe it would be better to manually patch it up

PR #91, opened 2014-07-10
Allow to specify the maximum length of a line with the new split lines feature

falsovsky:
This adds a new setting to irssi called "split_max_length" as it existed on splitlines.pl.

It allows the user to define the max length of the messages before they are split.
If it is set to 0 (the default) it will use the default split lines algorithm to find the max_length.

This is an addition to Pull Request #29.


madduck:
Why was this closed? Being able to export the max length setting to per-network stanzas seems like a very good addition to this feature. In my case, using Bitlbee, as well as another IRCd proxy run locally, both of which have no max lengths, this would be the perfect solution to the problem I otherwise have (not being able to send long commands)

PR #106, opened 2014-07-22
Some fixes for the Haiku port

mmuman:
- we currently have Links2 ported but not elinks nor lynx,
- socket() and friends are in libnetwork now,
  (btw, you'll want to check out AC_SEARCH_LIBS instead of nested AC_CHECK_LIB, cf. [here](https://www.flameeyes.eu/autotools-mythbuster/autoconf/finding.html))
- Haiku doesn't have /usr; and now with multi-arch package management, headers can be at different locations.


mmuman:
Oh, and now the ncurses test was dropped altogether ;-)

Issue #97, opened 2014-07-13
Right-aligned statusbars do not refresh / redraw properly.


cpbills:
This has been an issue in 0.8.15, and I have not seen mention of it in the existing bugs or the fix changelog for 0.8.16.

Example:

/statusbar topic add -alignment right topic
/j #somechannel
/topic long topic of stuff and junk and things
/j #someotherchannel
/topic shorter topic

Then switch between channels and see the longer topic not clear properly, for the short topic.

PR #109, opened 2014-07-23
Fix right aligned statusbar item redrawing

cpbills:
Fixes Github issue #97 https://github.com/irssi/irssi/issues/97

Fix proposed by ailin-nemui, built and tested on Debian Jessie using
0.8.15 source, tested by GeertHauwaerts as well.


ailin-nemui:
irssi-import/bugs.irssi.org#921

PR #115, opened 2014-07-28
Make ^ key and Ctrl+^ key usable with /BIND

ailin-nemui:
Fixes FS#721

This makes Ctrl+^ and ^ bindable again as different keys.

Please test.


dequis:
Tested in xterm/urxvt/vte/st/tmux/screen/linux tty and I guess there's no difference at all between those.

Patch looks good. The `^-` seemed a bit weird at first, but it's clever, and harmless.

ahf:
LGTdx, thanks!

Issue #105, opened 2014-07-20
Show git commit in CTCP VERSION.


KindOne-:
Currently the CTCP VERSION shows: ```irssi v0.8.17-head - running on Linux x86_64```
I am thinking something like this: ```irssi v0.8.17-6e68b83 - running on Linux x86_64```

The reason behind this suggestion is to easily find out the version of irssi someone is running, if they find some bug or have some issue.

KindOne-:
Thanks.

PR #124, opened 2014-08-09
Include commit and if the tree is dirty in builds of -head versions

dgl:
Implements request in #105


ahf:
Poke? :-)

dgl:
I don't think doing it in `configure.ac` makes sense for us, when we've done -rc releases we've explicitly marked them as such, so likely won't use this to mark dists.

The use I see for it is someone has a git checkout of Irssi and they don't know where it is built from. If we make it use the git revision when `./autogen.sh` is run there could still be doubt (e.g. git pull and then just re-run make without `./autogen.sh`).

dequis:
@ahf sup.

ahf:
Let's get it in! Thanks.

PR #126, opened 2014-08-10
Fix glib warnings when tab completing in weird places

### LINKED ISSUE #125: Some glib assertion failure warnings when tab completing
This is probably harmless, but may hint at some wrong behavior. It's been a ["known bug" for over 10 years][fs]. A few months ago I just walked into #irssi and wasted a few hours debugging this thing and coming up with a simple patch to fix the warnings, so now that the chances to get patches merged are non-zero, I'm posting this.

[Relevant IRC logs from 2014-03-19][logs] (may contain inaccurate information as we were figuring it out)

How to reproduce (just one way, there might be others):

* Be connected to at least one server
* Have one query window open
* In the query window, type write any number of spaces followed any number of non-spaces, but no spaces at the end of the message
    - For example `_a` works (replace _ with space)
* Move the cursor to the leftmost non-space character (that would be "`_|a`", underscore being space, pipe being cursor)
* Press tab
* Acquire warnings in status window:

```
01:12 -!- Irssi: critical g_string_erase: assertion 'pos <= string->len' failed
01:12 -!- Irssi: critical g_string_insert: assertion 'pos <= string->len' failed
01:12 -!- Irssi: critical g_string_insert_c: assertion 'pos <= string->len' failed
```

Sending my patch for this issue separately as a pull request, since as you can see in the logs i linked above, I wasn't completely sure if that was the right way to fix it.

[fs]: http://bugs.irssi.org/index.php?do=details&task_id=124
[logs]: https://gist.github.com/dequis/0673902b4dc216b79b5e

dequis:
See github issue #125 for more details (or [flyspray issue 124](http://bugs.irssi.org/index.php?do=details&task_id=124), submitted 10 years ago), and [these IRC logs linked from the other ticket](https://gist.github.com/dequis/0673902b4dc216b79b5e) for my comments while I was writing this thing.

It was months ago so I don't really remember how sure I was if this was the correct way to fix this issue, but it does get rid of the warnings and doesn't do anything too weird.

Like I said in the other ticket, submitting "now that the chances to get patches merged are non-zero".

Fixes #125.

### DIFF
--- a/src/fe-common/core/completion.c
+++ b/src/fe-common/core/completion.c
@@ -162,7 +162,8 @@ char *word_complete(WINDOW_REC *window, const char *line, int *pos, int erase, i
 		if (isseparator(*line)) {
 			/* empty space at the start of line */
 			if (wordstart == line)
-				wordstart += strlen(wordstart);
+				while (wordstart != '\0' && isseparator(wordstart))
+					wordstart++;
 		} else {
 			while (wordstart > line && isseparator(wordstart[-1]))
 				wordstart--;


dequis:
Closing this since the patch is wrong (first `wordstart` should be `*wordstart`) and honestly I should look more into those assertion warnings.

Might reopen later, or maybe submit a different PR.

PR #128, opened 2014-08-11
Add term_force_256colors setting.

mattvz:
This patch adds a term_force_256color option. I'd agree with requiring a proper TERM env variable, however this is the only way to get Konsole to display bold text without making it bright.
konsole-irssi

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -175,6 +175,15 @@ AC_ARG_ENABLE(true-color,
 	fi,
 	want_truecolor=no)
 
+AC_ARG_ENABLE(force-256color,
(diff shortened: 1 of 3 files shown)


dequis:
Poke? Can we decide if we want this patch or not, instead of leaving it here open forever?

dequis:
@ahf ^

ahf:
I agree with @dgl - I think we should document ourselves out of this, but have anyone tried to reach out to the projects where this is a problem? Like, say, konsole?

ailin-nemui:
this can be fixed through terminfo or terminal emulator

PR #148, opened 2014-10-03
[WIP] Native Proxy support for Irssi

hawken93:
See the changes in TODO. There are warnings produced while running it and there are also some usage of "\_\_attribute__((\__packed__));"
Nevertheless, I would like feedback :)

### DIFF
--- a/TODO
+++ b/TODO
@@ -1,3 +1,8 @@
+ - New proxy code crashes if an invalid proxy_type setting is used
+ - Remove old socks code
+ - Lots of warnings at least when using socks5
+ - Clean up coding style
+
 19:36 [IRCNet] [muzzy] more bugs in irssi, apparently the new version:  foo splits out,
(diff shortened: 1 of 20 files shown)


dgl:
Just mentioned on #irssi that it would be cool to have support for external proxies (ala ssh's `ProxyCommand`). That way we don't need to support every proxy type but people can still experiment. (To be clear this would be in addition to the proxy types here as those seem reasonable to have in core).

vague666:
Just an opinion, but -host shouldn't be used to specify server to connect to since it's used to specify the interface to bind to elsewhere in irssi(/server), no need to add to the confusion

ahf:
Good point, thanks :-)

dequis:
This lives in https://github.com/irssi/irssi/tree/orphaned/native-proxy now

PR #152, opened 2014-10-11
Try to split long lines on spaces

sebth:
I didn't add this in the first splitlong-PR because I thought it was a bad idea: it's dependent on the terminal encoding; but a couple of users said they want this and the most common encodings uses ASCII spaces anyway, so it should work properly for the majority. Spaces are left intact on the previous line, so if a user has an encoding where 0x20 isn't space, it won't get removed. It can be turned off with the option "split_line_on_space".


ahf:
LGTM - let's get it tested.

PR #169, opened 2014-11-05
New setting rejoin_channels_on_reconnect with fixes to comments in PR#77

vague666:


src/irc/core/irc-servers.c
```diff
@@ -61,6 +62,7 @@ void irc_servers_reconnect_init(void);
 void irc_servers_reconnect_deinit(void);
 
 static int cmd_tag;
+static char *rejoin_channels_mode;
```
dequis: Any particular reason this is in this scope and not inside irc_server_get_channels?

src/irc/core/irc-servers.c
```diff
@@ -586,33 +588,60 @@ char *irc_server_get_channels(IRC_SERVER_REC *server)
...
+	rejoin_channels_mode = g_strdup(settings_get_str("rejoin_channels_on_reconnect"));
+
 	int use_keys;
 
+	if (rejoin_channels_mode == NULL ||
+	    (g_ascii_strcasecmp(rejoin_channels_mode, "on") != 0 &&
+	     g_ascii_strcasecmp(rejoin_channels_mode, "off") != 0 &&
+	     g_ascii_strcasecmp(rejoin_channels_mode, "auto") != 0)) {
+		g_warning("Invalid value for 'rejoin_channels_on_reconnect', valid values are 'on', 'off', 'auto', using 'on' as default value.");
+		rejoin_channels_mode = g_strdup("on");
```
dequis: Leak here if it's not null

src/irc/core/irc-servers.c
```diff
@@ -586,33 +588,60 @@ char *irc_server_get_channels(IRC_SERVER_REC *server)
 	GSList *tmp;
 	GString *chans, *keys;
 	char *ret;
+	rejoin_channels_mode = g_strdup(settings_get_str("rejoin_channels_on_reconnect"));
```
dequis: Leak here the second time this function is called. Blame the scope of the variable, which as far as i can see isn't needed. Also I don't think the g_strdup is needed at all.

dequis:
Also, please use `git commit --amend` and `git push --force` to change the commit message to the current PR title. Also don't open yet another PR plzkthx<3

vague666:
@dequis 
patch updated :)

dequis:
Uhhh, you left two commits instead of one, the first one with the misleading title.

Also the leak when replacing rejoin_channels_mode with a new strdup'd "on" still applies

vague666:
Ah, I think I know what went wrong with the commit at least, I did a git commit -am "...", I assume you meant I should've done git commit --amend "..." instead? I was wondering what you meant by it
I've removed gstrdup now, so it's just rejoin_channels_mode = "on" now, which should be ok I hope :)
Can I assume I should do git commit --amend "..." now?
I have no deeper understanding of git so I apologize for my ignorance. I don't know if you saw the discussion in channel :) git is beyond me

dequis:
>I did a git commit -am "...", I assume you meant I should've done git commit --amend "..." instead? I was wondering what you meant by it

That's silly. Completely different commands.

>Can I assume I should do git commit --amend "..." now?

Not quite. That changes the last commit, and there are two commits now. Try, uh, "git rebase -i HEAD~2", and change the second line to squash

>I've removed gstrdup now, so it's just rejoin_channels_mode = "on" now, which should be ok I hope :)

Not quite. That still throws away the old strdup'd pointer, and replaces it with a string that isn't strdup'd, so if you g_free() that one, bad things happen. You need to keep the strdup and do a g_free() right before it, to avoid leaking the old pointer and put something else that can be freed in its place.

vague666:
Should be ok now, thanks for your patience dx :)

ailin-nemui:
@vague666 apparently there is still a g_free missing

dequis:
Went ahead and fixed the leak, changed commit message and did a few minor indentation changes resulting in 9586766d87820bdfa9e7cbe805f5ed4b7b0378e5

Issue #187, opened 2014-12-10
crash on loading config.


dajohi:
irssi crashes on startup with `irssi --config test-config`.  irssi should catch invalid syntax and not crash :)

Crash:
```
Program received signal SIGSEGV, Segmentation fault.
0x00001e008638d41d in config_node_first (list=0x1e02cac20560) at get.c:284
284                     if (node->type != NODE_TYPE_COMMENT)
(gdb) bt
#0  0x00001e008638d41d in config_node_first (list=0x1e02cac20560) at get.c:284
#1  0x00001e0086364cab in read_servers () at servers-setup.c:520
#2  0x00001e0086376540 in signal_emit_real (rec=0x1e0340fca8a0, params=Variable "params" is not available.
) at signals.c:242
#3  0x00001e0086377711 in signal_emit (signal=Variable "signal" is not available.
) at signals.c:286
#4  0x00001e00861faa8e in fe_common_core_finish_init () at fe-common-core.c:427
#5  0x00001e00861a79b5 in main (argc=3, argv=0x7f7ffffcdb48) at irssi.c:205
#6  0x00001e0086120de1 in _start () from /home/dhill/git/irssi/src/fe-text/irssi
#7  0x0000000000000000 in ?? ()
```

`test-config`
```
servers 5 (
  { address = "eu.irc6.net"; chatnet = "IRCnet"; port = "6667"; },
  { address = "opt.net"; chatnet = "IRCnet"; port = "6667"; },
);
```

dequis:
Close?

ailin-nemui:
@ahf ping

GeertHauwaerts:
This issue has been resolved in #191.

PR #191, opened 2014-12-16
Die if the wrong type of node is found when traversing config

dgl:
Fixes issue #187. It's a bit annoying this can't do anything other than
exit, however as there's no schema for the config it's only possible to
validate on use. This level of config can't be accessed from Perl so a
script can't cause Irssi to die (via this method at least).


ahf:
I hate out config system, but I agree it's better to crash early. LGTM, thanks! :-)

PR #199, opened 2015-01-07
Make config parser more robust

ailin-nemui:
This seems slightly more opinionated than #191

### DIFF
--- a/src/fe-text/irssi.c
+++ b/src/fe-text/irssi.c
@@ -310,6 +310,7 @@ int main(int argc, char **argv)
 		{ "version", 'v', 0, G_OPTION_ARG_NONE, &version, "Display irssi version", NULL },
 		{ NULL }
 	};
+	int loglev;
 
 #ifdef USE_GC
 	g_mem_set_vtable(&gc_mem_table);
@@ -352,6 +353,7 @@ int main(int argc, char **argv)
 	   you have to call setlocale(LC_ALL, "") */
 	setlocale(LC_ALL, "");
 
+	loglev = g_log_set_always_fatal(G_LOG_FATAL_MASK | G_LOG_LEVEL_CRITICAL);
 	textui_init();
 
 	if (!dummy && !term_init()) {
@@ -360,6 +362,7 @@ int main(int argc, char **argv)
 		return 1;
 	}
 
+	g_log_set_always_fatal(loglev);
 	textui_finish_init();
 	main_loop = g_main_new(TRUE);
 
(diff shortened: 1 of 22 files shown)


ahf:
Awesome. Let's get some eyes on this and get a lot of people to test this :-)

src/fe-text/irssi.c
```diff
@@ -352,6 +353,7 @@ int main(int argc, char **argv)
 	   you have to call setlocale(LC_ALL, "") */
 	setlocale(LC_ALL, "");
 
+	loglev = g_log_set_always_fatal(G_LOG_FATAL_MASK | G_LOG_LEVEL_CRITICAL);
```
dequis: Wouldn't this mean that glib "critical" errors such as the #125 assertion failures would result in irssi crashing?

Or for example stuff like "Source ID ### was not found when attempting to remove it" which i've seen in a lot of glib apps after glib 2.40 was released

dequis:
Reviewing this thing is hard without actually understanding irssi's weird config system first...

Took me a while to get past those `s/config_node_section/i&/`, maybe would have been better to split those in a separate commit to separate the diff noise more easily. Not that it matters a lot, I already realized I should just skim past those to get to the interesting stuff.

The changes to `src/lib-config/get.c`, are what needs some review by a real irssi dev.

Other than that, see my line comment.

src/fe-text/irssi.c
dequis: There's also some `s/g_error/g_critical/` - maybe could reverse those, have all of them as `g_error` and get rid of this `g_log_set_always_fatal`?

ailin-nemui: g_error is always fatal, so we turn it into g_critical. The commit #191 which @ahf merged and I'm not happy with, will crash your irssi instantly if you accidentally `/reload` a broken config file. By using g_critical, we will be able to inform the user gracefully on `/reload` but still refuse to load this config on initial start by temporarily raising the fatality here. It is re-set on line 365.

dequis: Ooooh, that's awesome. Right, I didn't realize that the return value of g_log_set_always_fatal is the old mask. Okay, I like this.

ahf: /me too.

dequis:
Thanks for splitting this! And by adding the commit logs it makes much more sense. First four commits are straightforward and look good to me. Still don't really understand fef25d6a35d23f885937b087ed7642b327279df5, @ahf should check that one.

ahf:
Let's see how many b0rked configs we'll see by getting this in.

PR #215, opened 2015-02-06
./configure has apparently been changed to ./autogen.sh but INSTALL ...

Lohhari:
...was not updated

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -11,7 +11,7 @@ To compile irssi you need:
 
 For most people, this should work just fine:
 
- ./configure
+ ./autogen.sh
  make
  su
  make install     (not _really_ required except for perl support)


ailin-nemui:
Not quite, INSTALL is for releases, which don't need autogen. After autogen once, you'll have configure. There should be better instructions for using irssi from git, but you'd need to check with a project admin where to put them best. I made some sort of a guide as part of https://github.com/shabble/irssi-docs/wiki/Irssi-0.8.17

Lohhari:
Thanks for the clarification ailin-nemui. Nice Wiki page. Another strangeish feature: the irssi executable is generated in src/fe-text/, not ./ or src/core/ for instance.

ailin-nemui:
yea that's how things work ;-)

PR #219, opened 2015-02-10
handle an already used nick different from the one we send

erdnaxeli:
This commit fixes the following bug.

If we set a nick with a space, like "abc d", the IRC server only understands the first part, here "abc". Now imagine the case where this nick ("abc") is already used. The server notifies Irssi, but Irssi think the nick is "abc d" so it tries "abc d_". The server only understands the "abc" part, and so on, so the connection stays stuck.

This commit fixes that by checking when receiving a "nick already used" message if the nick sent by the server is the same as the one we have registered. If not, we use the nick from the server, and the classic process (with underscores and numbers) continues.

### DIFF
--- a/src/irc/core/irc-nicklist.c
+++ b/src/irc/core/irc-nicklist.c
@@ -343,7 +343,7 @@ static void event_nick_invalid(IRC_SERVER_REC *server, const char *data)
 
 static void event_nick_in_use(IRC_SERVER_REC *server, const char *data)
 {
-	char *str, *cmd;
+	char *str, *cmd, *params, *nick;
 	int n;
 
 	g_return_if_fail(data != NULL);
@@ -352,6 +352,14 @@ static void event_nick_in_use(IRC_SERVER_REC *server, const char *data)
 		/* Already connected, no need to handle this anymore. */
 		return;
 	}
+
+	params = event_get_params(data, 2, NULL, &nick);
+	if (g_ascii_strcasecmp(server->nick, nick) != 0) {
+		/* the server uses a nick different from the one we send */
+		g_free(server->nick);
+		server->nick = g_strdup(nick);
+	}
+	g_free(params);
 
 	/* nick already in use - need to change it .. */
 	if (g_ascii_strcasecmp(server->nick, server->connrec->nick) == 0 &&


dequis:
>If we set a nick with a space, like "abc d"

I think you're fixing the wrong issue here...

erdnaxeli:
Can you be more specific?

On irc someone says to me that some protocols may accept a nick with a space, so we cannot reject it when it is set.

dequis:
IRC certainly doesn't, and this sounds like an IRC specific problem to me. Sure, it might not be correct to reject it in the UI side, but the irc protocol side shouldn't be attempting to send an invalid nick

erdnaxeli:
Ok. So maybe we can handle this in server_init() in irc/core/irc-servers.c?

ailin-nemui:
might be viable, what do you think @dequis

ailin-nemui:
I kinda like this idea, could we have some comment @ahf

ailin-nemui:
@erdnaxeli I would like to fix this issue, I see that there is the following issue: if you set the alternate nick to an invalid (space-containing) nick, then the code could -- based on the server reply -- re-set the nick to the alternate nick and again be trapped in a circle. do you agree?

erdnaxeli:
If the alternate nick contains a space too, irssi (with my commit) will still figure out that the nick send back by the server is different, and then the process with underscore will start.

Tested and validated.

erdnaxeli:
I've not spend any time on this issue till our discussion in September. Is the idea of handling this case in server_init() in irc/core/irc-servers.c still the right solution? We could check both nick and alternate_nick here for spaces.

I can take time this week-end to work on it.

erdnaxeli:
Ok, I tried to set conn->nick to a correct nick without space (I just takes the substring before the space), and a valid nick is sent. But then, if this nick is already taken, the alternate nick process tries with the wrong nick (containing space). It doesn't use the version set on server_init().

ailin-nemui:
hi @erdnaxeli , my original analysis was flawed:

if the nick (as per /set nick) contains spaces or otherwise the server uses a different nick, *then* the alt_nick is *not* tried by your code:

the comparison whether `nick EQ server->nick` AND `nick NOT EQ alt_nick` fails (because the server->nick was corrected)

we just need to move the comparison before the correction

timofonic:
@erdnaxeli @ailin-nemui @dequis @LemonBoy @coekie @ahf @GeertHauwaerts @falconindy What do you think about this? It seems @ailin-nemui corrected himself. Any news about this? Will it be possible to see it solved? Better nickname handling is always something good for any IRC/comm client :D

dequis:
Don't do that. It's rude and you're pinging people who aren't even involved in this.

PR #222, opened 2015-02-11
[RFC] CAP support

LemonBoy:
Take 2 on the problem :dancer: 
The (initial, simple, PLAIN only) SASL implementation you see it's just as a PoC to test how well the CAP system would work in a real-life scenario.
Please, review with care and pour your heart out :heart: 

For reference, here's the relevant specifications:
* [IRCv3.1](http://ircv3.net/specs/core/capability-negotiation-3.1.html)
* [IRCv3.2](http://ircv3.net/specs/core/capability-negotiation-3.2.html)

[Review on Reviewable](https://reviewable.io/reviews/irssi/irssi/222)

### DIFF
--- a/src/core/misc.c
+++ b/src/core/misc.c
@@ -211,6 +211,30 @@ void *gslist_foreach_find(GSList *list, FOREACH_FIND_FUNC func, const void *data
 	return NULL;
 }
 
+void gslist_free_full (GSList *list, GDestroyNotify free_func)
+{
+	GSList *tmp;
+
+	if (!list)
+		return;
+
+	for (tmp = list; tmp != NULL; tmp = tmp->next)
+		free_func(tmp->data);
+
+	g_slist_free(list);
+}
+
+GSList *gslist_remove_string (GSList *list, const char *str)
+{
+	GSList *l;
+
+	l = g_slist_find_custom(list, str, (GCompareFunc) g_strcmp0);
+	if (l != NULL)
+		return g_slist_remove_link(list, l);
+
+	return list;
+}
+
 /* `list' contains pointer to structure with a char* to string. */
 char *gslistptr_to_string(GSList *list, int offset, const char *delimiter)
 {
--- a/src/irc/core/Makefile.am
+++ b/src/irc/core/Makefile.am
@@ -26,6 +26,7 @@ libirc_core_a_SOURCES = \
         irc-servers-reconnect.c \
         irc-servers-setup.c \
         irc-session.c \
+        irc-cap.c \
         lag.c \
         massjoin.c \
         modes.c \
@@ -48,6 +49,7 @@ pkginc_irc_core_HEADERS = \
         irc-queries.h \
         irc-servers.h \
         irc-servers-setup.h \
+        irc-cap.h \
         modes.h \
         mode-lists.h \
 	module.h \
--- a/src/irc/core/irc-cap.c
+++ b/src/irc/core/irc-cap.c
@@ -1,0 +1,172 @@
+#include "module.h"
+#include "signals.h"
+#include "misc.h"
+
+#include "irc-cap.h"
+#include "irc-servers.h"
+
+int cap_toggle (IRC_SERVER_REC *server, char *cap, int enable)
+{
+	if (cap == NULL || *cap == '\0')
+		return FALSE;
+
+	/* If the negotiation hasn't been completed yet just queue the requests */
+	if (!server->cap_complete) {
+		if (enable && !gslist_find_string(server->cap_queue, cap)) {
+			server->cap_queue = g_slist_prepend(server->cap_queue, g_strdup(cap));
+			return TRUE;
+		}
+		else if (!enable && gslist_find_string(server->cap_queue, cap)) {
+			server->cap_queue = gslist_remove_string(server->cap_queue, cap);
+			return TRUE;
+		}
+
+		return FALSE;
+	}
+
+	if (enable && !gslist_find_string(server->cap_active, cap)) {
+		/* Make sure the required cap is supported by the server */
+		if (!gslist_find_string(server->cap_supported, cap))
+			return FALSE;
+
+		irc_send_cmdv(server, "CAP REQ %s", cap);
+		return TRUE;
+	}
+	else if (!enable && gslist_find_string(server->cap_active, cap)) {
+		irc_send_cmdv(server, "CAP REQ -%s", cap);
+		return TRUE;
+	}
+
+	return FALSE;
+}
+
+void cap_finish_negotiation (IRC_SERVER_REC *server)
+{
+	if (server->cap_complete)
+		return;
+
+	server->cap_complete = TRUE;
+	irc_send_cmd_now(server, "CAP END");
+
+	signal_emit("server cap end", 1, server);
+}
+
+static void cap_emit_signal (IRC_SERVER_REC *server, char *cmd, char *args)
+{
+	char *signal_name;
+
+	signal_name = g_strdup_printf("server cap %s %s", cmd, args? args: "");
+	signal_emit(signal_name, 1, server);
+	g_free(signal_name);
+}
+
+static void event_cap (IRC_SERVER_REC *server, char *args, char *nick, char *address)
+{
+	GSList *tmp;
+	GString *cmd;
+	char *params, *evt, *list, **caps;
+	int i, caps_length, disable, avail_caps;
+
+	params = event_get_params(args, 3, NULL, &evt, &list);
+	if (params == NULL)
+		return;
+
+	/* Strip the trailing whitespaces before splitting the string, some servers send responses with
+	 * superfluous whitespaces that g_strsplit the interprets as tokens */
+	caps = g_strsplit(g_strchomp(list), " ", -1);
+	caps_length = g_strv_length(caps);
+
+	if (!g_strcmp0(evt, "LS")) {
+		/* Create a list of the supported caps */
+		for (i = 0; i < caps_length; i++)
+			server->cap_supported = g_slist_prepend(server->cap_supported, g_strdup(caps[i]));
+
+		/* Request the required caps, if any */
+		if (server->cap_queue == NULL) {
+			cap_finish_negotiation(server);
+		}
+		else {
+			cmd = g_string_new("CAP REQ :");
+
+			avail_caps = 0;
+
+			/* Check whether the cap is supported by the server */
+			for (tmp = server->cap_queue; tmp != NULL; tmp = tmp->next) {
+				if (gslist_find_string(server->cap_supported, tmp->data)) {
+					g_string_append_c(cmd, ' ');
+					g_string_append(cmd, tmp->data);
+
+					avail_caps++;
+				}
+			}
+
+			/* Clear the queue here */
+			gslist_free_full(server->cap_queue, (GDestroyNotify) g_free);
+			server->cap_queue = NULL;
+
+			/* If the server doesn't support any cap we requested close the negotiation here */
+			if (avail_caps > 0)
+				irc_send_cmd_now(server, cmd->str);
+			else
+				cap_finish_negotiation(server);
+
+			g_string_free(cmd, TRUE);
+		}
+	}
+	else if (!g_strcmp0(evt, "ACK")) {
+		int got_sasl = FALSE;
+
+		/* Emit a signal for every ack'd cap */
+		for (i = 0; i < caps_length; i++) {
+			disable = (*caps[i] == '-');
+
+			if (disable)
+				server->cap_active = gslist_remove_string(server->cap_active, caps[i] + 1);
+			else
+				server->cap_active = g_slist_prepend(server->cap_active, g_strdup(caps[i]));
+
+			if (!g_strcmp0(caps[i], "sasl"))
+				got_sasl = TRUE;
+
+			cap_emit_signal(server, "ack", caps[i]);
+		}
+
+		/* Hopefully the server has ack'd all the caps requested and we're ready to terminate the
+		 * negotiation, unless sasl was requested. In this case we must not terminate the negotiation
+		 * until the sasl handshake is over. */
+		if (got_sasl == FALSE)
+			cap_finish_negotiation(server);
+	}
+	else if (!g_strcmp0(evt, "NAK")) {
+		g_warning("The server answered with a NAK to our CAP request, this should not happen");
+
+		/* A NAK'd request means that a required cap can't be enabled or disabled, don't update the
+		 * list of active caps and notify the listeners. */
+		for (i = 0; i < caps_length; i++)
+			cap_emit_signal(server, "nak", caps[i]);
+	}
+
+	g_strfreev(caps);
+	g_free(params);
+}
+
+static void event_invalid_cap (IRC_SERVER_REC *server, const char *data, const char *from)
+{
+	/* The server didn't understand one (or more) requested caps, terminate the negotiation.
+	 * This could be handled in a graceful way but since it shouldn't really ever happen this seems a
+	 * good way to deal with 410 errors. */
+	server->cap_complete = FALSE;
+	irc_send_cmd_now(server, "CAP END");
+}
+
+void cap_init (void)
+{
+	signal_add_first("event cap", (SIGNAL_FUNC) event_cap);
+	signal_add_first("event 410", (SIGNAL_FUNC) event_invalid_cap);
+}
+
+void cap_deinit (void)
+{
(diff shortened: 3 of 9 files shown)


src/irc/core/irc-servers.c
```diff
@@ -968,6 +985,46 @@ void irc_server_init_isupport(IRC_SERVER_REC *server)
...
+		/* abort the authentication */
+		irc_send_cmd_now(server, "AUTHENTICATE *");
+		cap_register_done(server);
+		return;
+	}
+
+	blob = g_strdup_printf("%s-%s-%s",
+												 conn->username,
+												 conn->username,
+												 conn->password);
```
dequis: jesus christ, those tabs, what. do you have tabs set to 1 space or something?

src/irc/core/irc-servers.c
```diff
@@ -968,6 +985,46 @@ void irc_server_init_isupport(IRC_SERVER_REC *server)
...
+	len = strlen(blob);
+
+	blob[strlen(conn->username)] = '\0';
+	blob[strlen(conn->username) * 2 + 1] = '\0';
```
dequis: Since you're dealing with strings that have NULs in them, maybe it would be saner to use [GString][]s?

[Gstring]: https://developer.gnome.org/glib/stable/glib-Strings.html

src/irc/core/irc-servers.c
```diff
@@ -409,7 +423,10 @@ static void sig_disconnected(IRC_SERVER_REC *server)
                         server_redirect_destroy(tmp->next->data);
 	}
 	g_slist_free(server->cmdqueue);
-        server->cmdqueue = NULL;
+	server->cmdqueue = NULL;
+
+	g_slist_free_full(server->cap, g_free);
```
dequis: I don't know what's the current glib version requirement of irssi (configure.ac says 2.6.0 but that's obviously not true) but `g_slist_free_full()` was added in 2.28

dequis: >that's obviously not true

I take that back. I can't find any symbols from newer glib releases. This also means that using `g_base64_encode()` below would bump it up to 2.12 (released 2006, included in centos 5, debian lenny 5.0, ubuntu 8.04, so i'm pretty sure nobody is going to complain about that one)

`g_slist_free_full()` is definitely too recent though, and it's not too hard to replace it with a loop.

src/irc/core/irc-cap.c
```diff
@@ -0,0 +1,82 @@
...
+{
+	/* We managed to get all the caps we requested */ 
+	if (++server->cap_got == g_hash_table_size(server->cap_req)) {
+		server->cap_complete = 1;
+		irc_send_cmd_now(server, "CAP END");
+	}
+}
+
+#define foreach_cap(t,l) \
+	for (t = strtok(l, " "); t; t = strtok(NULL, " "))
```
dequis: strtok is kinda ugly (and very thread unsafe, not that irssi uses threads but still ugly).

And not sure if this macro is even needed. `g_strsplit()`?

src/irc/core/Makefile.am
```diff
@@ -26,6 +26,7 @@ libirc_core_a_SOURCES = \
         irc-servers-reconnect.c \
         irc-servers-setup.c \
         irc-session.c \
+				irc-cap.c \
```
dequis: nit, should be spaces instead of tabs

src/irc/core/irc-cap.c
```diff
@@ -0,0 +1,96 @@
...
+				got_sasl = TRUE;
+
+			sig_name = g_strdup_printf("server cap %s", caps[i]);
+			signal_emit(sig_name, 1, server);
+			g_free(sig_name);
+		}
+
+		/* the negotiation is now over, unless 'sasl' is requested.
+		 * in that case the negotiation will be closed after the login succeeds */
+		if (got_sasl == FALSE) {
```
dequis: Not really sure if it's okay to do this here. Maybe there could be other ircv3 extensions that require delaying registration? Maybe could pass a pointer to a variable like ``signal_emit(sig_name, 2, &continue_registration)` and the handler of the "server cap sasl" would set that to true?

*(removed old post and reposted in the right place)*

src/irc/core/irc-servers.c
```diff
@@ -968,6 +979,43 @@ void irc_server_init_isupport(IRC_SERVER_REC *server)
 	}
 }
 
+void event_auth_start (IRC_SERVER_REC *server)
+{
+	irc_send_cmd_now(server, "AUTHENTICATE PLAIN");
+}
+
+void event_auth_end (IRC_SERVER_REC *server) {
```
dequis: nit, `{` below

dequis:
Regarding 4903acaeba9ee5e652f3fb795059085cef1d4ae6, indeed does look much better. It's easier to understand now, I think.

This still should get some review from someone with more experience with the irssi codebase.

dequis:
Comment from #ircv3:

```
08:34 < mquin> skimming those comits it looks like they're heading in the right direction, having a sensible cap implemention in the core will make hooking in new extensions a lot easier than the current mess
```

LemonBoy:
Right now the code breaks the CAP negotiation if a requested CAP isn't supported by the server, I feel this isn't the best move.
@ahf

ahf:
Can that be fixed easily? :-)

LemonBoy:
Huh, sure. But I wonder if it has to signal that error somehow (how ?)

LemonBoy:
SURPRISE MOTHERFUCKER!
I've made some changes here and there, the plan is to handle SASL by interleaving it with the CAP requests. Plus it avoids the need of having two separate lists to hold the capacities.

ahf:
DAYUM. I'll review this in the weekend. Awesome! There's a merge failure though :-)

src/irc/core/irc-cap.c
```diff
@@ -0,0 +1,127 @@
...
+
+	params = event_get_params(args, 3, NULL, &evt, &list);
+	caps = g_strsplit(list, " ", -1);
+
+	g_warning("EVENT %s", evt);
+
+	if (!strcmp(evt, "LS")) {
+		m = 0;
+
+		for (int i = 0; caps[i] != NULL; i++)
```
auscompgeek: https://travis-ci.org/irssi/irssi/jobs/58005857#L1551

src/irc/core/irc-cap.c
```diff
@@ -0,0 +1,128 @@
...
+
+	g_warning("cap queue %x", server->cap_queue);
+
+	/* the negotiation is over */
+	if (server->cap_queue == CAP_NONE) {
+		irc_send_cmd_now(server, "CAP END");
+		return;
+	}
+
+	zeros = __builtin_ctz(server->cap_queue);
```
ahf: This stands out for me as being odd.

ahf: This looks odd.

ahf:
Need to remove the `g_warning()` calls used for debugging as well. We have switched to use `g_strcmp0()` instead of `strcmp()` :-)

Would it be possible to squash the patches together to one single patch?

dequis:
Hmmm, the weird __builtin_ctz made me look at how you were doing that bitmask stuff, and... while it's cute and probably very efficient, it also means it's not extensible at all. For instance, someone might want to implement znc-specific capabilities in a script (vendor-prefixed, `znc.in/whatever`), and you aren't going to see those in the official ircv3 specs.

The second version of this PR (4903acaeba9ee5e652f3fb795059085cef1d4ae6) wasn't limited to a defined set of capabilities, and even seemed to be decently usable from scripts.

Another thing regarding the current version (0adc450219dbb11dfee45fd727798b35654508b5): there's that concept of queue in `cap_request_next` which i'm not sure if it's a good idea. It sends caps one at a time with CAP REQ, and that command supports sending a space separated list just fine

dequis:
Ah dammit you had me confused for a while trying to figure out what happened with the stuff I complained about in the last comment, turns out you just rebased removing those two commits completely.

Real commit messages would be appreciated, or at least comments in here, or something more than just random irc pings saying "are you there?" while i'm afk, and quitting afterwards. Aaaaanyway, time to read the code and figure out what you did.

src/irc/core/irc-cap.c
```diff
@@ -0,0 +1,229 @@
...
+	 * The whole request is then encoded in base64. */
+
+	req = g_string_new(NULL);
+
+	g_string_append_c(req, '\0');
+	g_string_append(req, conn->username);
+	g_string_append_c(req, '\0');
+	g_string_append(req, conn->password);
+
+	enc_req = g_base64_encode((const guchar *)req->str, req->len);
```
dequis: Added in glib 2.12, current requirement is 2.6.

IMO this is a good reason to bump the requirement. @ahf?

RHEL 5, the oldest still-supported distro version i know of, has glib 2.12, coincidentally, so i imagine it would cost nothing to bump it

src/irc/core/irc-servers.c
```diff
@@ -201,6 +202,10 @@ static void server_init(IRC_SERVER_REC *server)
 
 	conn = server->connrec;
 
+	server->cap_active = g_slist_append(server->cap_active, "identify-msg");
```
dequis:
```
08:09 < Nei> I'd remove that cap
08:09 < Nei> it's bullshitcrap
```

src/irc/core/irc-servers.h
```diff
@@ -69,6 +69,10 @@ struct _IRC_SERVER_REC {
 	int max_whois_in_cmd; /* max. number of nicks in one /WHOIS command */
 	int max_msgs_in_cmd; /* max. number of targets in one /MSG */
 
+	GSList *cap_supported; /* A list of caps supported by the server */
+	GSList *cap_active; /* A list of the caps active for this session */
```
dequis: So, hm, this is a bit misleading, since you actually add sasl here before it's been acknowledged, and remove it on NAK. So I guess it's more like a list of requested-and-probably-active caps.

src/irc/core/irc-cap.c
```diff
@@ -0,0 +1,229 @@
...
+#include "irc-cap.h"
+#include "irc-servers.h"
+
+int cap_toggle (IRC_SERVER_REC *server, char *cap, int enable) {
+	/* This is meant to be used after the inital negotiation is done */
+	if (!server->cap_complete)
+		return FALSE;
+
+	if (enable && g_slist_find(server->cap_supported, cap)) {
+		irc_send_cmdv(server, "CAP REQ %s", cap);
```
dequis: Shouldn't this also add to the requested cap to `server->cap_active`? Since receiving the ACK doesn't do that atm

src/irc/core/irc-cap.c
```diff
@@ -0,0 +1,229 @@
...
+	int i, caps_length;
+
+	params = event_get_params(args, 3, NULL, &evt, &list);
+	if (params == NULL)
+		return;
+
+	caps = g_strsplit(list, " ", -1);
+	caps_length = g_strv_length(caps);
+
+	if (!strcmp(evt, "LS")) {
```
dequis: g_strcmp0 everywhere!

src/irc/core/irc-cap.c
```diff
@@ -0,0 +1,229 @@
...
+static void sasl_step (IRC_SERVER_REC *server, const char *data, const char *from)
+{
+	IRC_SERVER_CONNECT_REC *conn;
+	GString *req;
+	char *enc_req;
+
+	conn = server->connrec;
+
+	/* Stop the timer */
+	g_source_remove_by_funcs_user_data((GSourceFunc) sasl_timeout, NULL);
```
dequis: I think the last parameter should match the passed `user_data` in the original `g_timeout_add`, so, `server` instead of `NULL`

dequis:
Minor details aside, I think I like this.

@LemonBoy :heart:

LemonBoy:
@dx Thanks for the review, you made me realized that I was using some glib's functions the wrong way, but now everything should be fine.

ahf:
I'm OK with requiring Glib 2.12 as long as you have checked the various big "distro's" and it sounds like you have :-)

dequis:
I wonder how often @dx gets pinged instead of me and i don't even notice.

@ahf neat.

So that would be [this line in configure.ac](https://github.com/irssi/irssi/blob/c122a2a226363114e11f32072395e2f06ad6a450/configure.ac#L273) and maybe [this one too in INSTALL](https://github.com/irssi/irssi/blob/c122a2a226363114e11f32072395e2f06ad6a450/INSTALL#L7)

@LemonBoy looks much better!

I think some more review of the scriptability of this design would be needed, at least to confirm that perl bindings can be added on top of this stuff

src/irc/core/irc-cap.c
```diff
@@ -0,0 +1,241 @@
...
+	if (params == NULL)
+		return;
+
+	caps = g_strsplit(list, " ", -1);
+	caps_length = g_strv_length(caps);
+
+	if (!g_strcmp0(evt, "LS")) {
+		/* Create a list of the supported caps */
+		for (i = 0; i < caps_length; i++)
+			server->cap_supported = g_slist_append(server->cap_supported, g_strdup(caps[i]));
```
dequis: >Note that `g_slist_append()` has to traverse the entire list to find the end, which is inefficient when adding multiple elements. A common idiom to avoid the inefficiency is to prepend the elements and reverse the list when all elements have been added

Since order doesn't really matter, replacing these calls with `g_slist_prepend()` should work.

src/irc/core/irc-servers.c
```diff
@@ -201,6 +202,10 @@ static void server_init(IRC_SERVER_REC *server)
 
 	conn = server->connrec;
 
+	// server->cap_queue = g_slist_append(server->cap_active, "identify-msg");
+	server->cap_queue = g_slist_append(server->cap_active, "sasl");
```
dequis: `s/cap_active/cap_queue/`

Also I think `g_strdup()`-ing the values might be saner.

Also there should be a way to allow scripts to request caps before registration is finished

src/irc/core/irc-cap.c
LemonBoy: It was mainly for consistency sake, most of the code uses g_slist_append and the lists we're talking about are short.

src/irc/core/irc-servers.c
LemonBoy: The latest iteration provides a better cap_queue that should be nice enough to expose to scripts

ahf:
Add -sasl_username, -sasl_password, -sasl_timeout and -sasl_mechanism to /NETWORK. Fix documentation for NETWORK too.

Split the patches up: one CAP + one SASL.

Dx:
Im getting used to :P

dequis:
There are a few minor mentions of sasl still, mainly in the handler of CAP ACK, but i think it's not worth wasting time splitting that off, given how CAP, SASL and the registration flow in general are closely related to each other.

I was going to complain about the usage of gslist_free_full (glib 2.28) but i see it's defined in misc.c.

:+1:

@ahf this seems complete to me

ailin-nemui:
can someone check the interoperation of CAP with HTTP PROXY?

dequis:
```
CAP LS
CONNECT irc.freenode.net:6667 HTTP/1.0


NICK dx
USER dx dx irc.freenode.net :dequis
```

welp

I'm blaming irssi's terrible http proxy implementation for being terrible (asking the user to do `/EVAL SET proxy_string CONNECT %s:%d HTTP/1.0\n\n` is pretty bad)

Worth noting that the #irssi user in question was using [cap_sasl.pl][], not this pull request.

[cap_sasl.pl]: https://github.com/irssi/scripts.irssi.org/blob/370c39219fe59c48825ca7c6fa71da80cd61a43f/scripts/cap_sasl.pl#L31

dequis:
@ahf @GeertHauwaerts the issue mentioned above is fixed, i believe this is ready to be merged

src/core/misc.c
```diff
@@ -211,6 +211,30 @@ void *gslist_foreach_find(GSList *list, FOREACH_FIND_FUNC func, const void *data
 	return NULL;
 }
 
+void gslist_free_full (GSList *list, GDestroyNotify free_func)
+{
+	GSList *tmp;
+
+	if (!list)
```
ahf: Nit: `if (list == NULL)`

src/irc/core/irc-cap.c
```diff
@@ -0,0 +1,172 @@
+#include "module.h"
```
ahf: You're missing a licence header here :-)

ahf:
LGTM. The two minor issues can be resolved in another patch. Thanks a lot!

PR #116, opened 2014-07-30
install more recent gcc from a PPA for travis

ailin-nemui:

PR #225, opened 2015-02-16
Silence unused value warnings in perl binding code

dequis:
sed -i 's/hv_store/(void) &/'

This only results in a warning in older gcc versions, but that includes
the one used in the Travis CI environment by default

Replaces PR #116. Quoting @ahf:

>I personally much prefer being explicit in the code (with a type-cast to void) over disabling a warning option.


ahf:
I hate it, but I much prefer it over the other option. Let's get it in! Thanks

PR #226, opened 2015-02-20
Sync from scripts: Fix scriptassist broken on unload

ailin-nemui:
Sync https://github.com/irssi/scripts.irssi.org/pull/135


dequis:
The #134 in the commit message links to the issue in this repo, not in the other one, you might want to clarify it's irssi/scripts.irssi.org#134

Other than that, this is a change already approved elsewhere, so yeah, good.

ailin-nemui:
yea let's just remove that from this copy of the commit

dequis:
@ahf trivial merge right here

PR #234, opened 2015-04-07
Add SNI support

Adam-:


dgl:
LGTM.

BTW, does any server/bouncer care about this or use this for anything (yet)? Just wondering what plans/ideas if any you have for this.

ahf:
Thanks!

Adam-:
Plexus 4 does as of the other day and I will be using it to serve the correct certificate (eg either irc.example.com if you connect to the RR or the servername.example.com). mIRC recently added support too but I'm not aware of any other servers using it.

dgl:
Seems like that sort of thing would be better served by a cert with subjectAltNames, but okay; I wasn't aware of anything else IRC related supporting it.

Adam-:
Based on my understanding that would require re-issuing the certificate every time you had a server change (new server comes/old server goes). And it would allow servers to pretend to be other servers. Unless you issued per-server certificates maybe.

PR #235, opened 2015-04-08
Change all strcmp() to g_strcmp0() to handle nulls gracefully

dequis:
Just a string replacement (but i did check every one of them)

    sed -i 's/strcmp(/g_strcmp0(/g' **/*.c

-----------

Also added a backwards compatibility macro since centos 5 is supported until 2017 and it's stuck with glib 2.12


ahf:
It's a minor style issue I commented on. Let's get this in.

Issue #231, opened 2015-04-07
RFC: enhancement: native RobustIRC support


stapelberg:
I’m one of the developers of http://robustirc.net/. See http://robustirc.net/docs.html for more details, but in a nutshell, it’s an IRC network written on top of https://raftconsensus.github.io/ which doesn’t suffer from netsplits. To not be affected by network issues between the client and the IRC servers, clients should use the RobustSession protocol which wraps IRC in a connection-independent HTTPS/JSON-based transport, see http://robustirc.net/docs/robustsession.html for details.

Currently, we offer a program called the RobustIRC bridge, which you can configure as a network/proxy server in irssi. The bridge then does the conversion (bridging) from IRC to IRC in RobustSession.

I intend to submit a pull request for irssi to talk RobustSession natively, i.e. without the bridge. Would such a pull request be accepted?

Thank you.

ailin-nemui:
I wonder how feasible it would be to contain your connection in an external module instead, similar to the [ICB/SILC/Quassel](https://github.com/shabble/irssi-docs/wiki/Modules) ones?

dgl:
Yeah, I think a module is the way to go to start with, it also means users of existing versions should be able to use your module and therefore it is more useful in the near term. Then if robustirc sees reasonable adoption we can think about including it with Irssi itself.

ahf:
I love what you guys are working on, but I think this belongs in a module. We cannot take ownership of this code as it is right now, but we'll happily help you if you have a need for it when you start working on your module :-)

stapelberg:
Thanks for the advice, everyone. I’ll investigate the feasibility of putting this into a module over the weekend.

stapelberg:
I’ve pushed the initial commit of a RobustIRC module to https://github.com/robustirc/irssi-robustirc. At the moment, it requires that you patch in pull request #237.

I’m happy to hear any feedback you have, especially with regards to the architecture of the module. The basic functionality works, but I haven’t tested this extensively.

kaniini:
Please be aware that RobustIRC is not robust in any way, and only trades availability for consistency tolerance.  If a RobustIRC user is connected to the partitioned side of the network, they cannot use the network at all until the partition is resolved.  This makes it a robustness downgrade in all cases as availability is the primary quality of an IRC network.

stapelberg:
[...]

PR #237, opened 2015-04-12
Introduce send_data_func for interception my modules.

stapelberg:
This is for the RobustIRC module we talked about in #231,
which now lives at https://github.com/robustirc/irssi-robustirc

### DIFF
--- a/src/core/server-rec.h
+++ b/src/core/server-rec.h
@@ -71,5 +71,6 @@ QUERY_REC *(*query_find_func)(SERVER_REC *server, const char *nick);
 int (*mask_match_func)(const char *mask, const char *data);
 /* returns true if `msg' was meant for `nick' */
 int (*nick_match_msg)(const char *nick, const char *msg);
+void (*send_data_func)(SERVER_REC *server, const char *data, int len);
 
 #undef STRUCT_SERVER_CONNECT_REC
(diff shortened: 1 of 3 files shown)


ailin-nemui:
not to inflict too much additional pain on you; I understand that you want to reuse the existing IRC protocol implementation. I'm not sure about the best strategy because your whole project feels like a hack to me. On the other hand it would be cool if irssi can be the place for such a playground.

I haven't looked into it, but how feasible would it be to do the redirection on a g_io_channel level? Then you shouldn't need this special case in the IRC parser here

stapelberg:
@ailin-nemui I think http://web.mit.edu/ghudson/dev/nokrb/third/yelp/src/yelp-io-channel.c is an example of how this might work. I’ll take a look.

Note that while it might work, https://developer.gnome.org/glib/stable/glib-IO-Channels.html#GIOChannel forbids us from doing that, though: “A data structure representing an IO Channel. The fields should be considered private and should only be accessed with the following functions.”

So, merging this pull request might be the cleaner solution after all :).

dequis:
I'm surprised they even document stuff like GIOFuncs when the only way to set those is to access private members.

ailin-nemui:
I suspect the documentation might be talking about consumers of GIOChannels, not about implementors

stapelberg:
With https://github.com/robustirc/irssi-robustirc/commit/a176acafe4724c6205e3599f3a1ba690e96c3729, this pull request becomes obsolete.

Thanks for the hint, @ailin-nemui!

PR #240, opened 2015-04-18
trivial: Minor cosmetic changes fixing docs as per review from ahf.

hloeung:
Trivial fix for missing period/dots in docs.

Thanks @dgl and @ahf for reviewing my previous pull request.


ahf:
Thanks for fixing this :-)

PR #247, opened 2015-05-14
lastlog doc fix

FreeFull:
Only the filename can come right after `-file`. Having `-force` in that position causes an
`Irssi: Could not open lastlog: No such file or directory` error.


ailin-nemui:
@ahf merge

PR #248, opened 2015-05-15
Implement CHANTYPES support

LemonBoy:
It's almost a matter of replacing stuff here and there, I tested it for a couple of minutes and found no problems (doesn't crash and allows me to join channels).

### DIFF
--- a/src/fe-common/irc/fe-events.c
+++ b/src/fe-common/irc/fe-events.c
@@ -42,9 +42,6 @@
 #include "fe-windows.h"
 #include "fe-irc-server.h"
 
-int fe_channel_is_opchannel(IRC_SERVER_REC *server, const char *target);
-const char *fe_channel_skip_prefix(IRC_SERVER_REC *server, const char *target);
-
 static void event_privmsg(IRC_SERVER_REC *server, const char *data,
 			  const char *nick, const char *addr)
 {
--- a/src/fe-common/irc/fe-events.c
+++ b/src/fe-common/irc/fe-events.c
@@ -42,6 +42,9 @@
 #include "fe-windows.h"
 #include "fe-irc-server.h"
 
+int fe_channel_is_opchannel(IRC_SERVER_REC *server, const char *target);
+const char *fe_channel_skip_prefix(IRC_SERVER_REC *server, const char *target);
+
 static void event_privmsg(IRC_SERVER_REC *server, const char *data,
 			  const char *nick, const char *addr)
 {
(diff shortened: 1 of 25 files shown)


src/irc/core/irc-servers.c
```diff
@@ -71,13 +71,20 @@ static int isnickflag_func(SERVER_REC *server, char flag)
...
-	return ischannel(*data);
+	IRC_SERVER_REC *irc_server = (IRC_SERVER_REC *) server;
+	char *chantypes;
+
+	chantypes = g_hash_table_lookup(irc_server->isupport, "chantypes");
+	if (chantypes == NULL)
+		chantypes = "#&!+"; /* normal, local, secure, modeless */
+
+	/* @#channel, @+#channel */
+	if (data[0] == '@' && data[1] == '+')
```
dequis: You got review'd from a different ticket: https://github.com/irssi/irssi/issues/253#issuecomment-106934788

>the logic introduced at [this line] does not account for the fact that an arbitrary number of channel types can be prefixed to the "#channel", including duplicates

ahf:
This looks good to me - I want someone else to review it as well. I like anything that kills these odd defaults.

src/irc/proxy/listen.c
```diff
@@ -262,7 +262,7 @@ static void handle_client_cmd(CLIENT_REC *client, char *cmd, char *args,
 
 		ignore_next = TRUE;
 		if (*msg != '\001' || msg[strlen(msg)-1] != '\001') {
-	        	signal_emit(ischannel(*target) ?
+	        	signal_emit(server_ischannel(SERVER(client->server), target) ?
```
dequis: nit: spaces mixed with tabs

dequis:
Also looks good to me, tested with a modified bitlbee that used `=` and `*` channels

ahf:
Fix the nit and I'll merge it :-)

src/fe-common/irc/fe-events.c
```diff
@@ -42,6 +42,9 @@
 #include "fe-windows.h"
 #include "fe-irc-server.h"
 
+int fe_channel_is_opchannel(IRC_SERVER_REC *server, const char *target);
+const char *fe_channel_skip_prefix(IRC_SERVER_REC *server, const char *target);
```
dequis: Would be nice to put these in some header instead

(reposting comment from diff view, thanks github!)

src/fe-common/irc/fe-irc-messages.c
```diff
@@ -37,28 +37,47 @@
...
 {
-	int i = 0;
-	const char *val, *chars;
+	const char *statusmsg;
+
+	/* Quick check */
+	if (server == NULL || server->prefix[(int)(unsigned char)*target] == 0)
+		return FALSE;
+
+	statusmsg = g_hash_table_lookup(server->isupport, "statusmsg");
```
dequis: Even though this is case insensitive, i'm not really sold on having it as lowercase, given that it's looking up an ISUPPORT value that is uppercase. But yeah. Nitpick.

dequis:
Did some more testing with the last commit, looks good to me.

Found some weirdness with formats, but it's not this patch's fault. Well, neither was the STATUSMSG issue, but...



(Screenshot from the point of view of `dx__`)

Actions with target have the target bolded, own actions with target (or normal messages) don't have it bolded.

Oddly enough it doesn't look like that in my main irssi, which uses almost exactly the same theme as the default. It is a mystery. After spending 5 minutes looking at this issue I decided it's not worth spending more time on. Mentioning it here just because.

But yeah, looks good to me. Would be nice to have @ahf's opinion on `fe_channel_is_opchannel` and `fe_channel_skip_prefix` (name and placement), check yesterday's irc logs for context.

dequis:
@ahf beep

src/fe-common/irc/fe-irc-channels.c
```diff
@@ -31,6 +31,51 @@
...
+
+	return strchr(statusmsg, *target) != NULL;
+}
+
+const char *fe_channel_skip_prefix(IRC_SERVER_REC *server, const char *target)
+{
+	const char *statusmsg;
+
+	/* Quick check */
+	if (server == NULL || server->prefix[(int)(unsigned char)*target] == 0)
```
ahf: The cast's here deserves a comment in the code. Make a follow up patch which cleans that up a bit please :-)

ahf:
01:36:20 <           dx> that's from the old code, anyway, not added by lemon

ahf:
LGTM, let's get it in.

PR #250, opened 2015-05-22
Add one more elif clause to autogen for w3m support.

Polyatomic:


ailin-nemui:
if I remember correctly w3m cannot make a legend of URLs to visible links when dumping. probably the whole system should be revamped and just deliver markdown or something.

Polyatomic:
Here is the dump of docs/faq.html  http://dpaste.com/36KBB1R, is that ok.

ailin-nemui:
Here's how it should look:
```
Q: Will there be /DETACH-like feature?
A: Maybe. Detach code already is there, attach is just missing :) But I
   don't have much interest in coding it, [2]screen and [3]dtach can be used
   to do it just fine.

....

   References

   Visible links
   ....
   2. http://www.gnu.org/software/screen/screen.html
   3. http://dtach.sf.net/
....
```

Polyatomic:
Ahh, now I see where w3m craps out. Cheers for raising my awareness on that man. I'll try harder next time.

PR #252, opened 2015-05-29
Update sb_search.pl

ailin-nemui:
make it easier to bind sb search to a key


dequis:
:+1:

Commit log message isn't great but uh, whatever. Important part is:

>Without arguments, the last search is repeated

Issue #245, opened 2015-05-03
Centos 5: `doc_DATA' is used but `docdir' is undefined


dequis:
Installed centos 5 in a docker container to see if this thing works at all, and nope.

#232 broke it.

```
[root@fa7e7b1af53f irssi]# ./autogen.sh
Creating help files...
Creating ChangeLog...
Documentation: html -> txt...
**Warning**: I am going to run `configure' with no arguments.
If you wish to pass any to it, please specify them on the
`./autogen.sh' command line.

Running autoreconf ...
Putting files in AC_CONFIG_AUX_DIR, `build-aux'.
autoheader: WARNING: Using auxiliary files such as `acconfig.h', `config.h.bot'
autoheader: WARNING: and `config.h.top', to define templates for `config.h.in'
autoheader: WARNING: is deprecated and discouraged.
autoheader:
autoheader: WARNING: Using the third argument of `AC_DEFINE' and
autoheader: WARNING: `AC_DEFINE_UNQUOTED' allows to define a template without
autoheader: WARNING: `acconfig.h':
autoheader:
autoheader: WARNING:   AC_DEFINE([NEED_FUNC_MAIN], 1,
autoheader:             [Define if a function `main' is needed.])
autoheader:
autoheader: WARNING: More sophisticated templates can also be produced, see the
autoheader: WARNING: documentation.
configure.ac: installing `build-aux/install-sh'
configure.ac: installing `build-aux/missing'
docs/Makefile.am:4: `doc_DATA' is used but `docdir' is undefined
src/core/Makefile.am: installing `build-aux/depcomp'
autoreconf: automake failed with exit status: 1
```

Versions:

* autoconf 2.59
* automake 1.9.6
* make 3.81

Autoconf proudly claims:

    Copyright (C) 2003 Free Software Foundation, Inc.

This works in other boxes with newer versions (tested with autoconf 2.69 and automake 1.15)

Apparently docdir is defined by newer autoconf versions (there are relevant changelog entries around 2005-2007)

Adding `docdir ?= ${datarootdir}/doc/irssi` didn't help.

Wat do.

GeertHauwaerts:
I confirm the issue and will revert #232.

Issue #230, opened 2015-04-06
docdir is ignored during installation


dimkr:
doc/Makefile.in always installs the documentation to $datarootdir/doc/irssi.

Here's a fix:
```
diff -rup irssi-0.8.17-orig/docs/Makefile.in irssi-0.8.17/docs/Makefile.in
--- irssi-0.8.17-orig/docs/Makefile.in	2015-04-06 19:09:36.907777873 +0300
+++ irssi-0.8.17/docs/Makefile.in	2015-04-06 19:12:24.083781846 +0300
@@ -315,7 +315,7 @@ build_vendor = @build_vendor@
 builddir = @builddir@
 datadir = @datadir@
 datarootdir = @datarootdir@
-docdir = $(datadir)/doc/irssi
+docdir = @docdir@
 dvidir = @dvidir@
 exec_prefix = @exec_prefix@
 host = @host@
```

ailin-nemui:
you should change `Makefile.am` instead, and make a Pull Request once you confirmed that default operation is not changed

ailin-nemui:
@ahf close

ailin-nemui:
do we really need to support CentOS 5? Is there any way to make this work on both? @dequis @dimkr

dequis:
The only reason I wanted to test with centos 5 is that i wanted to ensure that we don't use any glib functions introduced after the oldest glib version we claim to support. I don't care a lot about it personally, but it's still covered by redhat's absurdly long 10 year support cycle (one year left, iirc), so some people do use it.

In the case of bitlbee I decided to drop support for centos 5 because I wanted to use some features introduced in glib 2.16 (hash table iterators, GChecksum for md5/sha1, g_strcmp0). Irssi uses g_strcmp0 and provides a macro implementation of it for older glib versions - i suspect that macro implementation was broken by recent patches that pass g_strcmp0 as a parameter, but didn't get around to confirm that.

No one complained so far but I guess we can't really expect people who use dinosaurs to test bleeding edge versions of the client of the past.

Anyway, dropping centos 5 support should be considered - this ticket might be the first decent reason to do it.

dimkr:
@ailin-nemui I need this change to control docdir on a distro that uses a non-standard directory, not for CentOS 5.

ailin-nemui:
I know, @GeertHauwaerts removed your change because it does *not* work on CentOS5 :-(

GeertHauwaerts:
Then I guess we drop the support for CentOS5.

[...]

PR #258, opened 2015-06-12
Revert "Bug fix - docdir is ignored during installation."

GeertHauwaerts:
Reverts irssi/irssi#232; see irssi/irssi#245, irssi/irssi#245, irssi/irssi#230


dequis:
#230 should probably be reopened, then.

PR #259, opened 2015-06-16
Updates for irssiproxy (continuation of PR #64)

dequis:
As I mentioned in #64, i'm taking over and finishing a few details to get this merged.

- [x] Fix conflicts
- [x] Renamed /proxy to /irssiproxy
- [x] Fix some rare bug that I found while testing
- ~~Formats instead of hardcoded printfs~~
- ~~Consider what to do about /irssiproxy and /irssiproxy status being the same thing~~
- ~~Add /irssiproxy disconnect?~~
- Left for a future PR: SSL support?

Closes #64


GeertHauwaerts:
Those checkboxes are awesome :D

dequis:
So, while trying to figure out where the hell to put module-formats.c and module-formats.h (which are only present in the modules under `fe-*`), I found this:

https://github.com/dequis/irssi/blob/irssiproxy/src/irc/proxy/listen.c#L32

```c
#include "fe-common/core/printtext.h" /* FIXME: evil. need to do fe-proxy */
```

Added by db4a0e50 (2002-02-20).

I feel like i'm opening a can of worms with the format stuff. Can I not?

miniCruzer:
:+1: 

SSL is nice too if you're not ssh-tunneling into irssiproxy, and convenience for using a mobile client.

dequis:
Earlier today (was on my phone back then)

> < ahf> dx: https://github.com/irssi/irssi/pull/259 <- want to iterate more on this or?
> < dx> ahf: i've been considering leaving it as-is and leave ssl for a separate pull request, since that ssl patch is big enough
> < ahf> dx: oki, but i see some renaming of commands from proxy to irssiproxy
> < ahf> but the docs isn't in sync with that

@ahf I don't see what part of the docs isn't in sync with the command name change.

The signal names and the module name remain as they were before since I don't want to break that (at least not in this PR)

Anyway, declaring this thing ready, no point in blocking it anymore.

ahf:
Odd. Looks right, yeah. Not sure what I saw last night then.

PR #263, opened 2015-07-09
Introduce check for git commands in autogen

ailin-nemui:
This avoids a misleading compile error when users try to `make` an
incomplete (or non-git) check-out.

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -18,6 +18,10 @@ perl syntax.pl
 
 echo "Creating ChangeLog..."
 git log > $srcdir/ChangeLog
+if test "$?" -ne 0; then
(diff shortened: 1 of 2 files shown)


ailin-nemui:
The error

```
expandos.c
expandos.c: In function ‘expando_releasedate’:
expandos.c:407:49: error: expected expression before ‘)’ token
  return g_strdup_printf("%d", IRSSI_VERSION_DATE);
                                                 ^
expandos.c:408:1: warning: control reaches end of non-void function [-Wreturn-type]
 }
 ^
```

could be shown to people who used the "Download ZIP" button on github (apparently not so uncommon). It is much better to show them a meaningful warning instead.

ahf:
Neat.

PR #267, opened 2015-07-19
Add expandos for hostname

kyak:
See http://bugs.irssi.org/index.php?do=details&task_id=829

### DIFF
--- a/src/irc/core/irc-expandos.c
+++ b/src/irc/core/irc-expandos.c
@@ -82,13 +82,16 @@ static char *expando_hostname(SERVER_REC *server, void *item, int *free_ret)
 	IRC_SERVER_REC *ircserver;
 	char hostname[100];
 	char **list;
+	char *hostname_split;
 
 	ircserver = IRC_SERVER(server);
 
 	/* prefer the _real_ /userhost reply */
 	if (ircserver != NULL && ircserver->userhost != NULL) {
 		list = g_strsplit(ircserver->userhost, "@", -1);
-		return list[1];
+		hostname_split = g_strdup(list[1]);
+		g_strfreev(list);
+		return hostname_split;
 	}
 
 	/* haven't received userhost reply yet. guess something */
@@ -96,7 +99,7 @@ static char *expando_hostname(SERVER_REC *server, void *item, int *free_ret)
 
 	if (gethostname(hostname, sizeof(hostname)) != 0 || *hostname == '\0')
 		strcpy(hostname, "??");
-	return g_strconcat(hostname, NULL);
+	return g_strdup(hostname);
 }
 
 /* user mode in active server */
--- a/src/irc/core/irc-expandos.c
+++ b/src/irc/core/irc-expandos.c
@@ -76,6 +76,29 @@ static char *expando_userhost(SERVER_REC *server, void *item, int *free_ret)
 	return g_strconcat(username, "@", hostname, NULL);;
 }
 
+/* your hostname address (host) */
+static char *expando_hostname(SERVER_REC *server, void *item, int *free_ret)
+{
+	IRC_SERVER_REC *ircserver;
+	char hostname[100];
+	char **list;
+
+	ircserver = IRC_SERVER(server);
+
+	/* prefer the _real_ /userhost reply */
+	if (ircserver != NULL && ircserver->userhost != NULL) {
+		list = g_strsplit(ircserver->userhost, "@", -1);
+		return list[1];
+	}
+
+	/* haven't received userhost reply yet. guess something */
+	*free_ret = TRUE;
+
+	if (gethostname(hostname, sizeof(hostname)) != 0 || *hostname == '\0')
+		strcpy(hostname, "??");
+	return g_strconcat(hostname, NULL);
+}
+
 /* user mode in active server */
 static char *expando_usermode(SERVER_REC *server, void *item, int *free_ret)
 {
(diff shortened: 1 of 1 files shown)


src/irc/core/irc-expandos.c
```diff
@@ -76,6 +76,29 @@ static char *expando_userhost(SERVER_REC *server, void *item, int *free_ret)
 	return g_strconcat(username, "@", hostname, NULL);;
 }
 
+/* your hostname address (host) */
+static char *expando_hostname(SERVER_REC *server, void *item, int *free_ret)
+{
+	IRC_SERVER_REC *ircserver;
+	char hostname[100];
```
ahf: Hostnames should be up to `HOST_NAME_MAX` long on POSIX systems.

src/irc/core/irc-expandos.c
```diff
@@ -76,6 +76,29 @@ static char *expando_userhost(SERVER_REC *server, void *item, int *free_ret)
...
+	char **list;
+
+	ircserver = IRC_SERVER(server);
+
+	/* prefer the _real_ /userhost reply */
+	if (ircserver != NULL && ircserver->userhost != NULL) {
+		list = g_strsplit(ircserver->userhost, "@", -1);
+		return list[1];
```
ahf: No need to clean-up here? You need to duplicate the value and clean-up the vector afterwards.

src/irc/core/irc-expandos.c
```diff
@@ -76,6 +76,29 @@ static char *expando_userhost(SERVER_REC *server, void *item, int *free_ret)
...
+	}
+
+	/* haven't received userhost reply yet. guess something */
+	*free_ret = TRUE;
+
+	if (gethostname(hostname, sizeof(hostname)) != 0 || *hostname == '\0')
+		strcpy(hostname, "??");
```
ahf: I'd use something like `(none)` instead.

src/irc/core/irc-expandos.c
```diff
@@ -76,6 +76,29 @@ static char *expando_userhost(SERVER_REC *server, void *item, int *free_ret)
...
+	return g_strconcat(hostname, NULL);
```
ahf: `g_strdup()`.

kyak:
I mostly just copied the expando_userhost function. I think it is better to leave expando_hostname as it is or change both the new expando_hostname and the existing expando_userhost functions for consistency.

Basically you just reviewed the existing code. I'm not into changing the existing code.

dequis:
You're still leaking the `g_strsplit(ircserver->userhost, "@", -1)`

And `g_strconcat(hostname, NULL);` is pointless (it makes sense the way it's used in `expando_userhost`, with several strings to concatenate)

kyak:
1) Not sure i understood what exactly is leaking. Can you please elaborate?

2) Originally in expando_userhost the NULL is appended to "user@host". I don't know the reason for such implementation, but I just mimic the same behaviour.

Seriously, i didn't expect this trivial patch cause such a ripple effect on neighbouring code :) Making enhancements to surrounding functions was not in my plans.

dequis:
1) [Quoting the docs][1]

>Returns

>a newly-allocated NULL-terminated array of strings. Use [g_strfreev()][2] to free it

You are only returning list[1]. Assuming a string with a single `@`, you're leaking the string in `list[0]` and the two-element array of pointers to char that is `list`. Like ahf said,

>You need to duplicate the value and clean-up the vector afterwards.

That's a g_strdup(list[1]), g_strfreev(list), and return the dup'd string.

2) It's not appending NULL to the string, rather, that null is a sentinel value that must always be the last argument to `g_strconcat()` ([docs][3]) so that it knows when to stop reading the argument list.

So in the case of `expando_userhost`, it does `g_strconcat(username, "@", hostname, NULL)`, which concatenates the first three parameters, stops there, and returns a newly allocated string. Also worth nothing that those three parameters are null-terminated already, just like all C string literals.

In your case, you're "concatenating" hostname and nothing else, which returns a newly allocated string. That fills the requirement for these functions to return pointers to newly-allocated strings in the heap (you can't return "hostname" directly, as that is part of the stack), but is just silly since you're not concatenating anything. `g_strdup()` will do just fine.

------

(And you can ignore the changes that also apply to other functions, like the `HOST_NAME_MAX` or the `strcpy` of `??`)


[1]: https://developer.gnome.org/glib/2.42/glib-String-Utility-Functions.html#g-strsplit
[2]: https://developer.gnome.org/glib/2.42/glib-String-Utility-Functions.html#g-strfreev
[3]: https://developer.gnome.org/glib/2.42/glib-String-Utility-Functions.html#g-strconcat

kyak:
Thanks, guys. I hope it looks better now.

ahf:
You're still missing two of the things I pointed out :-)

ahf:
But the other fixes looks good!

kyak:
Hm, i thought we agreed not to touch the other two things...

ahf:
I think it should still be fixed, but it can be done in another patch where both your newly added code gets fixed AND the old code path above it. Not sure if the "??" is used elsewhere in our code.

src/irc/core/irc-expandos.c
```diff
@@ -76,6 +76,32 @@ static char *expando_userhost(SERVER_REC *server, void *item, int *free_ret)
...
+	char *hostname_split;
+
+	ircserver = IRC_SERVER(server);
+
+	/* prefer the _real_ /userhost reply */
+	if (ircserver != NULL && ircserver->userhost != NULL) {
+		list = g_strsplit(ircserver->userhost, "@", -1);
+		hostname_split = g_strdup(list[1]);
+		g_strfreev(list);
+		return hostname_split;
```
ahf: `*free_ret` must also be set to `TRUE` in this case, given that you return some newly initialised memory from your call to `g_strdup()`.

kyak:
Does it look better now?

ailin-nemui:
@kyak I believe there is a line note waiting for you to address

dequis:
Nah, that's addressed already, it just doesn't go away because the comment was elsewhere. If you view the full diff, there's a "free_ret" a few lines above the comment

ailin-nemui:
*** github

ailin-nemui:
@ahf care to take a final look

PR #269, opened 2015-07-25
Replace \n with \r\n in irssiproxy source, fixes bug

blha303:
Fixes [FS#780](http://bugs.irssi.org/index.php?do=details&task_id=780)

Bug confirmed by connecting to an unpatched irssi setup with Palaver, an iOS IRC client, which disconnected immediately after connection. After modifying the files as below and rebuilding irssi, Palaver could connect successfully. Other IRC clients would likely experience the same issue.


GeertHauwaerts:
LGTM

PR #274, opened 2015-08-30
Strip trailing whitespace from cmd args to avoid poor autocompletion UI

lamby:
On my Android phone the keyboard will "helpfully" always add a trailing
space when I select one of the autocomplete suggestions anticipating that
I will be adding another word. This seems a sensible default option in a
phone environment.

However, a command such as "/win 4 " is not accepted by irssi, thus when
using the auto-completion I have to manually remove this trailing space if
I want it to work. This patch therefore removes the trailing whitespaces
from command arguments, improving the mobile user experience.

See also:

 * http://bugs.irssi.org/index.php?do=details&task_id=727
 * http://bugs.irssi.org/index.php?do=details&task_id=769

Signed-off-by: Chris Lamb <chris@chris-lamb.co.uk>

### DIFF
--- a/src/core/commands.c
+++ b/src/core/commands.c
@@ -857,7 +857,12 @@ static void parse_command(const char *command, int expand_aliases,
 
 	cmd = orig = g_strconcat("command ", command, NULL);
 	args = strchr(cmd+8, ' ');
-	if (args != NULL) *args++ = '\0'; else args = "";
+	if (args == NULL)
+		args = "";
+	else {
+		*args++ = '\0';
+		strchomp(args);
+	}
 
 	/* check if there's an alias for command. Don't allow
 	   recursive aliases */
--- a/src/core/misc.c
+++ b/src/core/misc.c
@@ -1019,3 +1019,12 @@ char **strsplit_len(const char *str, int len, gboolean onspace)
 
 	return ret;
 }
+
+void strchomp(char *str) {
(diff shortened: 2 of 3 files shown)


jillest:
This patch strips off all whitespace from command arguments. For example, a command like `/window goto #irssi` now causes an error message `-!- Irssi: Unknown command: window goto#irssi`.

I agree that "/window 7 " should work, but this cure is worse than the disease. The right fix should be somewhere deeper in the parsing. Also, trailing whitespace should be preserved in commands like `/msg`, `/me` and `/action` that take text to send.

lamby:
> This patch strips off all whitespace from command argument

Oh, of course - how silly of me.

> trailing whitespace should be preserved in commands

ACK.

Will prepare another patch.

KindOne-:
I use [trigger.pl](https://github.com/irssi/scripts.irssi.org/blob/gh-pages/scripts/trigger.pl) with ```-send_command -regexp '\s+$' -replace ''``` to strip all the trailing whitespaces from commands.

lamby:
> I use trigger.pl [..]

Neat :) So, given that this is available in most distros and appears to work perfectly, am closing this PR given the above complications wrt only stripping whitespace for some commands (requiring an ugly whitelist/blacklist) I'm going to close this PR.

Thanks all :)

PR #275, opened 2015-09-02
Fix two minor issues outlined in the PR#222

LemonBoy:
irc-cap.c has now a licence header.
A minor style fix in misc.c


ahf:
Thanks The Lemon! :-)

PR #276, opened 2015-09-06
Send smkx/rmkx sequence at terminal init

LemonBoy:
Enter the "application" mode when setting up the terminal, this improves the
compatiblity with the standards; as a side effect now DEL key works ootb when
irssi is run in the suckless's st terminal.

Some more infos can be found here [1].

Successfully tested with : st, urxvt, vte-based terms, xterm (with metaSendsEscape set to True)

[1] http://git.suckless.org/st/tree/FAQ#n25

### DIFF
--- a/src/fe-text/terminfo-core.c
+++ b/src/fe-text/terminfo-core.c
@@ -50,62 +50,66 @@ TERM_REC *current_term;
 
 /* Define only what we might need */
 static TERMINFO_REC tcaps[] = {
-        /* Terminal size */
-	{ "cols",	"co",	CAP_TYPE_INT,	G_STRUCT_OFFSET(TERM_REC, width) },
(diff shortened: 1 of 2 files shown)


dequis:
Could you please not realign the whole struct?

I never understood st's problems with backspace/del. I use it, made a few slight changes to config.h and don't have problems... most of the time. Irssi certainly works. I should try again wtih a config that is more like the stock one.

They have also changed their stance on the issue a few times, and probably also the default behavior.

This change sounds like a Good Thing™ though.

LemonBoy:
Let's just stash the realign-job in the diff, I should've made another commit for that but my ocd kicked in.
From what I understand this is the right way™ to initialize the terminal and handle all the special keys and at the end of the day it doesn't harm to use those sequences (at least in the terminals I've tried with)

ahf:
Thanks

PR #277, opened 2015-09-09
Implement paste detection via the bracketed paste mode

LemonBoy:
Disabled by default as there's no way to check whether the terminal supports the bracketed paste mode, anyway `readline` approximates the detection in some way that might work quite well and we could be interested in that.

ping @dequis
ticket #9

### DIFF
--- a/src/fe-text/gui-readline.c
+++ b/src/fe-text/gui-readline.c
@@ -37,6 +37,7 @@
 #include "gui-windows.h"
 #include "utf8.h"
 
+#include <string.h>
 #include <signal.h>
 
 typedef void (*ENTRY_REDIRECT_KEY_FUNC) (int key, void *data, SERVER_REC *server, WI_ITEM_REC *item);
@@ -65,6 +66,13 @@ static char *paste_old_prompt;
 static int paste_prompt, paste_line_count;
 static int paste_join_multiline;
 static int paste_timeout_id;
+static int paste_bracketed_mode;
+
+/* Terminal sequences that surround the input when the terminal has the
+ * bracketed paste mode active. Fror more details see
+ * https://cirw.in/blog/bracketed-paste */
+static const unichar bp_start[] = { 0x1b, '[', '2', '0', '0', '~' };
+static const unichar bp_end[]   = { 0x1b, '[', '2', '0', '1', '~' };
 
 static void sig_input(void);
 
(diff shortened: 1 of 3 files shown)


dequis:
The patch looks good.

>anyway readline approximates the detection in some way that might work quite well and we could be interested in that

Wait, what?

LemonBoy:
See the 'rl_bracketed_paste_probably_supported_p' function in [1]. It's not rocket proof but should allow us to enable this by default if the terminal passes the test.

[1]https://lists.gnu.org/archive/html/bug-bash/2014-10/msg00211.html

ailin-nemui:
I think the proper way to do it is to emit signals from the sasl.c and turn them into messages from another fe-sasl.c

dequis:
This looks good to me but I wrote the last part so of course it looks good to me. Although I guess it counts as reviewed by lemon, too. Wouldn't mind another pair of eyes though.

Pending some testing with laggy connections and mosh, although i think it shouldn't really matter.

ailin-nemui:
Can you submit the fix for the line garbling separately?

src/fe-text/gui-readline.c
```diff
@@ -647,21 +683,47 @@ static void sig_input(void)
...
 			paste_timeout_id = g_timeout_add(paste_detect_time, paste_timeout, NULL);
-		} else {
+		} else if (!paste_bracketed_mode) {
 			int i;
 
 			for (i = 0; i < paste_buffer->len; i++) {
 				unichar key = g_array_index(paste_buffer, unichar, i);
 				signal_emit("gui key pressed", 1, GINT_TO_POINTER(key));
+
+				if (paste_bracketed_mode) {
```
ailin-nemui: is this possible to be true when the if on 715 says !?

src/fe-text/gui-readline.c
```diff
@@ -647,21 +683,47 @@ static void sig_input(void)
...
+				return;
+			}
+
+			for (i = 0; i <= len; i++, ptr++) {
+				if (ptr[0] == bp_end[0] && !memcmp(ptr, bp_end, sizeof(bp_end))) {
+					paste_bracketed_end(i, i != len);
+					break;
+				}
+			}
+		}
```
ailin-nemui: no \n

src/fe-text/gui-readline.c
```diff
@@ -647,21 +683,47 @@ static void sig_input(void)
...
+			int i;
+			int len = paste_buffer->len - G_N_ELEMENTS(bp_end);
+			unichar *ptr = (unichar *) paste_buffer->data;
+
+			if (len <= 0) {
+				return;
+			}
+
+			for (i = 0; i <= len; i++, ptr++) {
+				if (ptr[0] == bp_end[0] && !memcmp(ptr, bp_end, sizeof(bp_end))) {
```
ailin-nemui: iirc irssi coding style demands explicit == 0 everywhere

dequis:
Cherry picked last commit and submitted #292, this one needs rebasing so that the commit doesn't appear twice in the history.

src/fe-text/gui-readline.c
dequis: Yes, see the comment right below `/* just enabled by the signal, remove what was processed so far */`

That is, it was false before, but the signal enabled it.

dequis:
Submitted https://github.com/LemonBoy/irssi/pull/2 - fixes the style issues mentioned above, broken rest length calculation, also several other edge cases. Copypasting:

- Last line not ending with a newline
- Small pastes that end in the same sig_input() call where they started
- Paste start marker right after the end marker, which is a st bug: http://lists.suckless.org/dev/1509/27328.html
- Empty pastes and sequences of those (just don't do anything)

dequis:
Replacing this with #306 (as requested by lemon)

PR #278, opened 2015-09-10
SASL support

LemonBoy:
:dancer: 

The EXTERNAL mechanism is untested as of now and we might need some documentation (do we ?)


ailin-nemui:
+1

src/irc/core/sasl.c
```diff
@@ -0,0 +1,134 @@
...
+
+static gboolean sasl_timeout (IRC_SERVER_REC *server)
+{
+	/* The authentication timed out, we can't do much beside terminating it */
+	g_critical("The authentication timed out, try increasing the timeout and check your connection "
+	           "to the network.");
+	irc_send_cmd_now(server, "AUTHENTICATE *");
+	cap_finish_negotiation(server);
+
+	return G_SOURCE_REMOVE;
```
dequis: This was added in glib 2.32 and is defined to `FALSE`. Just use `FALSE`.

src/irc/core/irc-chatnets.c
```diff
@@ -48,6 +48,10 @@ static void sig_chatnet_read(IRC_CHATNET_REC *rec, CONFIG_NODE *node)
 	rec->max_msgs = config_node_get_int(node, "max_msgs", 0);
 	rec->max_modes = config_node_get_int(node, "max_modes", 0);
 	rec->max_whois = config_node_get_int(node, "max_whois", 0);
+
+	rec->sasl_mechanism = config_node_get_str(node, "sasl_mechanism", NULL);
+	rec->sasl_username = config_node_get_str(node, "sasl_username", NULL);
+	rec->sasl_password = config_node_get_str(node, "sasl_password", NULL);
 }
```
ahf: You need to handle saving these values in `sig_chatnet_saved` as well.

src/irc/core/sasl.c
```diff
@@ -0,0 +1,134 @@
...
+			irc_send_cmd_now(server, "AUTHENTICATE EXTERNAL");
+			break;
+	}
+	server->sasl_timeout = g_timeout_add(SASL_TIMEOUT, (GSourceFunc) sasl_timeout, server);
+}
+
+static void sasl_fail (IRC_SERVER_REC *server, const char *data, const char *from)
+{
+	/* Stop any pending timeout, if any */
+	g_source_remove(server->sasl_timeout);
```
dequis:
```
if (server->sasl_timeout != -1) {
        g_source_remove(server->sasl_timeout);
        server->sasl_timeout = -1;
}
```

For every g_source_remove.

src/irc/core/sasl.c
dequis: Also, `server->sasl_timeout = -1;` before this line

dequis:
Same thing as CAP: this definitely should be extensible through scripts. But maybe in a later PR.

`PLAIN` and `EXTERNAL` are probably more than enough for most users atm. AFAIK, the only other non-deprecated mechanism is `ECDSA-NIST256P-CHALLENGE` and who knows how long it will last until someone decides it should be deprecated because it uses NIST curves or something.

So yeah. Scripts can take care of keeping up with that

src/irc/core/sasl.c
```diff
@@ -0,0 +1,148 @@
...
+	server->sasl_timeout = g_timeout_add(SASL_TIMEOUT, (GSourceFunc) sasl_timeout, server);
+}
+
+static void sasl_fail (IRC_SERVER_REC *server, const char *data, const char *from)
+{
+	/* Stop any pending timeout, if any */
+  if (server->sasl_timeout != -1) {
+    g_source_remove(server->sasl_timeout);
+    server->sasl_timeout = -1;
+  }
```
dequis: Suddenly, two spaces of indentation, for some reason.

ahf:
**F***** NIST.

dequis:
You're sending `base64("\x00username\x00password")`, it should be `base64("username\x00username\x00password")`. Yes, the username twice. Technically they are slightly different, in practice they are the same.

Relevant #ircv3 snippet: 

>< frogdr> what's the difference between authentication identity and auth identity?
>< TingPing> "The
>< TingPing>    client presents the authorization identity (identity to act as),
>< TingPing>    followed by a NUL (U+0000) character, followed by the authentication
>< TingPing>    identity (identity whose password will be used)"
>< awilfox> on IRC, there is no difference
>< TingPing> in terms of irc ive never seen them be different
>< frogdr> oh
>< frogdr> ok
>< awilfox> but that would be useful for say 'sudo'
>< Jobe> <awilfox> on IRC, there is no difference <- strictly speaking there can be a difference, just very few services provide such a feature
>< Jobe> in part for security concerns
>< Jobe> but I do know of one package that does
>< Jobe> and the reason it does is to provide its network with a means for its official LDAP backed ZNC to auth as a user's own account without storing their password in plain text
>< Jobe> basically its not a feature an end user's client would use
>< Jobe> but it has its uses where network provided ZNC services are concerned
>< grawity> awilfox: well I implemented that for Atheme
>< grawity> frogdr: authentication id is the one your password / credentials are checked against
>< grawity> frogdr: authorization id is the one whose privileges / restrictions you get

dequis:
I have donated one (1) sasl implementation to bitlbee, matching lemon's donation of one (1) sasl implementation to irssi. And by that I mean I just implemented it in bitlbee to test it alongside this irssi client implementation. I should probably submit a PR in the bitlbee repo so you guys can help me review that stuff too.

Anyway, after fixing the username stuff, I tried to auth to it with sasl user `dx|prosody`, while `dx|skypeweb` as nick.  My implementation replies with numeric 902 and message `Your SASL username does not match your nickname` (other implementations don't care, i just don't want to deal with the asymmetry)

```
<< CAP LS
<< NICK dx|skypeweb
<< USER dx dx localhost :dequis
>> :localhost.localdomain NOTICE * :BitlBee-IRCd initialized, please go on
>> :localhost.localdomain CAP * LS :sasl
<< CAP REQ : sasl
>> :localhost.localdomain CAP dx|skypeweb ACK : sasl
<< AUTHENTICATE PLAIN
>> AUTHENTICATE +
<< AUTHENTICATE ZHh8cHJvc29keQBkeHxwcm9zb2R5AGFzZA==
>> :localhost.localdomain 902 dx|skypeweb :Your SASL username does not match your nickname
<< CAP END
```

But irssi only displays a generic error message:

>-!- Irssi: critical Authentication failed, make sure your credentials are correct and that the mechanism you have selected is supported by this server.

Should probably display the server message in that case.

Also, not an issue with this PR, but should be fixed: `CAP REQ : sasl`, there's a space after the colon.

Also, copyright headers.

dequis:
Another case, a server not supporting PLAIN, replying with 908 `RPL_SASLMECHS`, and following up with 906 `ERR_SASLABORTED` (somewhat incorrect response from the server, should be 904 failed)

```
<< AUTHENTICATE PLAIN
>> :localhost.localdomain 908 dx FOO,BAR :are available SASL mechanisms
>> :localhost.localdomain 906 dx :SASL authentication aborted
```

```
12:50:36 -!- FOO,BAR are available SASL mechanisms
12:50:36 -!- SASL authentication aborted
[20 seconds pass]
12:50:56 -!- Irssi: critical The authentication timed out, try increasing the timeout and check your connection to the network.
```

908 and 906 aren't handled, and show their messages as server crap. For 908 this is a good thing. 906 should be attached to `sasl_fail` too. And `sasl_fail` should display the server error messages.

You should probably handle 907 `ERR_SASLALREADY` too, since it might result in a 20 second timeout too. Not sure how you'd handle it, though.

src/irc/core/sasl.c
```diff
@@ -0,0 +1,155 @@
...
+
+static void sasl_success (IRC_SERVER_REC *server, const char *data, const char *from)
+{
+	if (server->sasl_timeout != -1) {
+		g_source_remove(server->sasl_timeout);
+		server->sasl_timeout = -1;
+	}
+
+	/* The authentication succeeded, time to finish the CAP negotiation */
+	g_warning("SASL authentication succeeded");
```
dequis: This shows `warning` in red scary text.

dequis:
Not really sure if these `g_log()` functions are the best way to show log messages, in general

dequis:
@LemonBoy You can now have some sweet revenge by reviewing my side of the implementation!

https://github.com/bitlbee/bitlbee/pull/46

ailin-nemui:
I think the proper way to do it is to emit signals from the sasl.c and turn them into messages from another fe-sasl.c

LemonBoy:
Do we really need another layer of abstraction for no good reason ?

dequis:
Of course, needless layers of abstraction are the irssi way /s

I dunno, you'd need signals to make the mechanism part pluggable, but i don't really know about the rest of the events that need to print messages. Unless there's something else plugins may want to do there that requires knowing the outcome of processing the numerics?

(I think it sucks that this is being held due to the way it prints messages because that normally would be a nitpicky complaint, but at the same time I think it's a significant inconsistency to leave it like this ;_;)

src/irc/core/irc-servers-setup.c
```diff
@@ -79,6 +80,29 @@ static void sig_server_setup_fill_chatnet(IRC_SERVER_CONNECT_REC *conn,
...
+	if (ircnet->sasl_mechanism != NULL) {
+		if (!g_ascii_strcasecmp(ircnet->sasl_mechanism, "plain")) {
+			/* The PLAIN method needs both the username and the password */
+			if (ircnet->sasl_username != NULL && *ircnet->sasl_username &&
+			    ircnet->sasl_password != NULL && *ircnet->sasl_password) {
+				conn->sasl_mechanism = SASL_MECHANISM_PLAIN;
+				conn->sasl_username = ircnet->sasl_username;
+				conn->sasl_password = ircnet->sasl_password;
+			} else
+				g_warning("The fields sasl_username and sasl_password are either undefined or empty");
```
ahf: Are either missing or empty. Can come as a follow up patch.

grawity:
Signals? Not sure what abstraction level you're aiming for, but I think start(server, mech_name), step(server, in_data), and abort(server) would be good... Perhaps also provide a `$server->send_sasl_reply(out_data)` and such, so that the mechanism could be reused by other protocol plugins like XMPP.

PR #280, opened 2015-09-13
Remove seemingly unused variables

maxdeliso:
Hi. I could be missing something here, but I saw a bunch of variables in the code that were allocated seemingly for no reason, with a pattern similar to the following:

```C
char *foo, *bar, ...

foo = bar = g_strdup ....

/* ignore foo */

g_free(foo);
```

I removed all the instances of this that I could find, as it seemed pointless to allocate this memory and then do nothing with it.

In order to allow me to make this change in more places, I changed the event parameter decoder, ```event_get_params```,  to return void, instead of a ```char*```, as every path except one that called that function was unconditionally freeing the pointer it received without examining its value, and the only one that did check the value just compared it to NULL, which was easy to fix.

I ran the client for a bit with my changes and it didn't crash, but didn't go to any great length to ensure that my changes didn't have any unwanted side effects; I couldn't find any test cases, nor any documentation on automated testing.

Let me know if this change is a :lemon:, or how to test it better.

Thanks,
--max

### DIFF
--- a/src/core/commands.c
+++ b/src/core/commands.c
@@ -297,7 +297,7 @@ void command_runsub(const char *cmd, const char *data,
 		    void *server, void *item)
 {
 	const char *newcmd;
-	char *orig, *subcmd, *defcmd, *args;
+	char *subcmd, *defcmd, *args;
 
 	g_return_if_fail(data != NULL);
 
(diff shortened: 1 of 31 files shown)


maxdeliso:
Hmm. My branch is full of fail. Let me see if I can get Travis to build this...

maxdeliso:
Will re-open if I can get it to build using the command line that Travis uses.

maxdeliso:
See https://github.com/irssi/irssi/pull/281 for a slightly less failure-laden version of this one.

PR #281, opened 2015-09-13
Remove unused variables

maxdeliso:
This is another attempt of https://github.com/irssi/irssi/pull/280

### DIFF
--- a/src/core/commands.c
+++ b/src/core/commands.c
@@ -297,7 +297,7 @@ void command_runsub(const char *cmd, const char *data,
 		    void *server, void *item)
 {
 	const char *newcmd;
-	char *orig, *subcmd, *defcmd, *args;
+	char *subcmd, *defcmd, *args;
 
 	g_return_if_fail(data != NULL);
 
@@ -310,7 +310,7 @@ void command_runsub(const char *cmd, const char *data,
 	}
 
 	/* get command.. */
-	orig = subcmd = g_strdup_printf("command %s %s", cmd, data);
+	subcmd = g_strdup_printf("command %s %s", cmd, data);
 	args = strchr(subcmd+8 + strlen(cmd)+1, ' ');
 	if (args != NULL) *args++ = '\0'; else args = "";
 	while (*args == ' ') args++;
@@ -319,7 +319,6 @@ void command_runsub(const char *cmd, const char *data,
 	newcmd = command_expand(subcmd+8);
 	if (newcmd == NULL) {
                 /* ambiguous command */
-		g_free(orig);
 		return;
 	}
 
@@ -336,7 +335,6 @@ void command_runsub(const char *cmd, const char *data,
 	}
 
 	g_free(subcmd);
-	g_free(orig);
 }
 
 static GSList *optlist_find(GSList *optlist, const char *option)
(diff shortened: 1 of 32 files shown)


dequis:
So, here's command_runsub:

```c
 void command_runsub(const char *cmd, const char *data,
                    void *server, void *item)
 {
        const char *newcmd;
-       char *orig, *subcmd, *defcmd, *args;
+       char *subcmd, *defcmd, *args;
 
 [...]
 
        /* get command.. */
-       orig = subcmd = g_strdup_printf("command %s %s", cmd, data);
+       subcmd = g_strdup_printf("command %s %s", cmd, data);
        args = strchr(subcmd+8 + strlen(cmd)+1, ' ');
        if (args != NULL) *args++ = '\0'; else args = "";
        while (*args == ' ') args++;
 
        /* check if this command can be expanded */
        newcmd = command_expand(subcmd+8);
        if (newcmd == NULL) {
                 /* ambiguous command */
-               g_free(orig);
                return;
        }
 
        subcmd = g_strconcat("command ", newcmd, NULL);
 
 [...]
 
        g_free(subcmd);
-       g_free(orig);
 }
```

`g_strdup_printf()` allocates new memory and sets the pointer to orig and subcmd:

```c
       orig = subcmd = g_strdup_printf("command %s %s", cmd, data);
```

`command_expand()` takes `subcmd+8`. Sometimes it returns an expansion, sometimes it returns null, sometimes it return the same thing it received (if there's no expansion)

```c
        newcmd = command_expand(subcmd+8);
```

`g_strconcat()` allocates new memory and sets the pointer to subcmd, overwriting the old value of subcmd.

```c
        subcmd = g_strconcat("command ", newcmd, NULL);
```

If you don't keep the old value of subcmd around, that's a leak.

If you free the old value of subcmd before calling `g_strconcat()`, that's use-after-free in the case where `command_expand()` returns the same value it was passed.

Don't do that.

In general, if you see code that is doing something weird and apparently pointless, but more complex than the natural solution, there is probably a reason for it.

Also, you might want to look into valgrind. It's a tool that can track memory management bugs, like leaks or use-after-free. It's definitely recommended to ensure that you aren't introducing new bugs, but fixing existing bugs requires special care (it's easy to break things by attempting to fix a memory leak)

maxdeliso:
Hi @dequis.

Thanks for the explanation. I can see now that I did introduce several memory allocation bugs into the code with this PR. That pattern of:

```C
foo = bar = allocation()

bar = possible_reallocation()
```

Means that you have to keep both ```foo``` and ```bar``` to be able to free everything.

I will go through my diffs later and remove the instances of that.

There are also a bunch of instances of a different kind, like the following:

```C
 static void sig_whois_event_not_found(IRC_SERVER_REC *server, const char *data)
 {
-	char *params, *nick;
+	char *nick;
 
-	params = event_get_params(data, 2, NULL, &nick);
+	event_get_params(data, 2, NULL, &nick);
 	printformat(server, nick, MSGLEVEL_CRAP, IRCTXT_WHOIS_NOT_FOUND, nick);
-	g_free(params);
 }
```

Where the ```params``` variable is never referenced. Is removing params here also introducing a memory problem, is there some hidden reason for it?

I will run valgrind in order to test the remaining changes, but for large changes wouldn't it make sense to have some kind of test runner which exercises the majority of this code's branches automatically? I only mention it because I touched so many branches with my change that it seems inordinately time consuming to test each affected branch manually with valgrind.

dequis:
Read the code of event_get_params. Don't blindly change code without understanding it. And don't submit pull requests with changes full of branches you don't know how to reach :P

If you want to write a test suite for irssi and maximize coverage, be my guest. There is nothing automatic in the process of writing that test suite, unfortunately, but if you're willing to work on that, it would be awesome.

dequis:
@ahf beep boop

ahf:
Hello,

Thanks for the contribution, but as @dequis have already stated, there are a couple of problems with this set of patches that means we cannot accept them. We are not a high volume project with lots of volunteers, which means we do expect contributors to ensure that the branches they have modified have also been tested - especially when they are cleaning up our codebase :-) Working on much smaller chunks at a time might make it easier.

Working on automated coverage scans and test suites would be highly appreciated as already stated by @dequis :-)

Closing for now.

PR #283, opened 2015-09-16
Clear the ICRNL flag

LemonBoy:
Separate the ^J behaviour from the ^M / ENTER one


ailin-nemui:
that key_bind shouldn't be there

dequis:
Then where?

Hmmm, there's this thing above:

```c
 	key_bind("key", NULL, "^M", "return", (SIGNAL_FUNC) key_combo);
 	key_bind("key", NULL, "^J", "return", (SIGNAL_FUNC) key_combo);
```

So... both ^M and ^J are "bound" to the "return" key?

```c
 	key_bind("multi", NULL, "return", "check_replaces;send_line", NULL);
+	key_bind("multi", NULL, "^J", "check_replaces;send_line", NULL);
```

Which makes that addition redundant, probably

LemonBoy:
Modified and commit amended

PR #289, opened 2015-09-21
Move new stuff at the end of IRC_SERVER_REC to attempt to unbreak ABI

dequis:
This normally wouldn't matter at all but I just spent a couple of hours
figuring out a bug that turned out to be Server->isupport() from a
0.8.17 ABI Irc.so reading from the splits hash table instead of the
isupport hash table, and both happen to be 32 bytes away from each
other, which is the same size of the newly added struct items. So it
failed in a very elegant and subtle way (it would probably crash and
burn in 32bit architectures, though)

This commit is just an attempt to get something useful out of that time.

It also moves the cap_complete flag to the bitfield above so it can be
properly packed (taking one bit instead of 4 bytes like it did in the
other position). That's 4 bytes less per network! 60 bytes total for me!
Amazing.

### DIFF
--- a/src/irc/core/irc-servers.h
+++ b/src/irc/core/irc-servers.h
@@ -63,18 +63,12 @@ struct _IRC_SERVER_REC {
 	unsigned int nick_collision:1; /* We're just now being killed because of nick collision */
 	unsigned int motd_got:1; /* We've received MOTD */
 	unsigned int isupport_sent:1; /* Server has sent us an isupport reply */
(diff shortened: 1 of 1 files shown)


ailin-nemui:
I'd rather introduce some version checks in the ABI to catch these fuckups early

dequis:
Closing since I don't think this is going to happen. I might do the cap_complete bitfield change separately. Maybe. Maybe not.

Issue #190, opened 2014-12-15
IPv6 and IPv4 addressing order (RFC 6724) not respected


thiagomacieira:
irssi seems to default to IPv4 addresses before IPv6. That's wrong: IPv6 should always be tried first, to help with the conversion to IPv6.

In fact, irssi should just respect the order provided by getaddrinfo(3). It is already configured to use IPv6 first if both sides have native IPv6, but use IPv4 first if both sides are tunnelling. See RFC 6724 (the new version of RFC 3484), which is implemented in most libc.

sc68cal:
The problem with setting `resolve_prefer_ipv6` is that if you're on a laptop and travel a lot, some networks will have IPv6 and some will not. Then you end up just passing `-4` for your `/connect` command.

eevee:
I recently discovered that `resolve_prefer_ipv6` will try IPv6 not just when you have no v6 address, but _even if_ IPv6 is disabled system-wide (via `net.ipv6.conf.all.disable_ipv6`).  Despite being named "prefer", the setting seems to make irssi never try IPv4 at all.

amckinlay:
Please remove this setting.

thiagomacieira:
Right, it needs to respect the order from getaddrinfo. And it needs to try multiple servers returned from that function, not just the first one.

thiagomacieira:
So, when are you guys going to fix this? It's been nearly 6 years.

ailin-nemui:
I understand the frustration about long open bugs. Unfortunately we don't have any "guy" who is capable of fixing the bugs. :-(

thiagomacieira:
Proper networking support should be at the basis of any networking application. I understand that irssi was not properly architected when it was first written, but we're in 2020 and IPv4 addresses have begun running out for over a decade. This issue makes an unpatched irssi unusable on an IPv6-enabled network.

ailin-nemui:
thanks for your comment, could you please open a pull request with your patch?

thiagomacieira:
#290 already exists. It's been reverted.

ailin-nemui:
apparently because there were issues with it? it's been continuediin #299 but was never finished. I see that you did add some comments on that pull request as well, so the question remains if you have a working patch?

thiagomacieira:
#290 has been working for me for 6 years. In this time, I have not seen any issues that didn't exist before with irssi.

[...]

PR #290, opened 2015-09-21
Network and IPv{4,6} related changes

LemonBoy:
First take at fixing #190.
WORKSFORME on IPv4 networks, with and without `resolve_reverse_lookup`.

### DIFF
--- a/docs/help/in/toggle.in
+++ b/docs/help/in/toggle.in
@@ -15,7 +15,7 @@
 
 %9Examples:%9
 
-    /TOGGLE resolve_prefer_ipv6
+    /TOGGLE resolve_reverse_lookup
     /TOGGLE channels_rejoin_unavailable ON
 
 %9See also:%9 SET
--- a/docs/manual.txt
+++ b/docs/manual.txt
@@ -429,9 +429,6 @@
 	After connected to server, Irssi can automatically change your user
 	mode. You can set it with /SET usermode <mode>, default is +i.
 
-	/SET resolve_prefer_ipv6 - If ON, prefer IPv6 for hosts that
-	     have both v4 and v6 addresses.
-
  5.5 Automatic reconnecting
 
 	If you get disconnected from server, Irssi will try to reconnect
--- a/src/core/chat-commands.c
+++ b/src/core/chat-commands.c
@@ -128,10 +128,10 @@ static SERVER_CONNECT_REC *get_server_connect(const char *data, int *plus_addr,
 
         host = g_hash_table_lookup(optlist, "host");
 	if (host != NULL && *host != '\0') {
-		IPADDR ip4, ip6;
+		IPADDR ip;
 
-		if (net_gethostbyname(host, &ip4, &ip6) == 0)
-                        server_connect_own_ip_save(conn, &ip4, &ip6);
+		if (net_gethostbyname(host, &ip) == 0)
+                        server_connect_own_ip_save(conn, &ip);
 	}
 
 	cmd_params_free(free_arg);
--- a/src/core/chatnet-rec.h
+++ b/src/core/chatnet-rec.h
@@ -9,4 +9,4 @@ char *realname;
 
 char *own_host; /* address to use when connecting this server */
 char *autosendcmd; /* command to send after connecting to this ircnet */
-IPADDR *own_ip4, *own_ip6; /* resolved own_address if not NULL */
+IPADDR *own_ip; /* resolved own_address if not NULL */
--- a/src/core/net-nonblock.c
+++ b/src/core/net-nonblock.c
@@ -103,15 +103,12 @@ int net_gethostbyname_nonblock(const char *addr, GIOChannel *pipe,
 	srand(time(NULL));
 
         memset(&rec, 0, sizeof(rec));
-	rec.error = net_gethostbyname(addr, &rec.ip4, &rec.ip6);
+	rec.error = net_gethostbyname(addr, &rec.ip);
 	if (rec.error == 0) {
 		errorstr = NULL;
 		if (reverse_lookup) {
 			/* reverse lookup the IP, ignore any error */
-			if (rec.ip4.family != 0)
-				net_gethostbyaddr(&rec.ip4, &rec.host4);
-			if (rec.ip6.family != 0)
-				net_gethostbyaddr(&rec.ip6, &rec.host6);
+			net_gethostbyaddr(&rec.ip, &rec.host);
 		}
 	} else {
 		errorstr = net_gethosterror(rec.error);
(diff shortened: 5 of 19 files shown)


ailin-nemui:
+1

dequis:
Paging @thiagomacieira from #190

thiagomacieira:
Will test soon.

thiagomacieira:
Test #1: network with no IPv6:
```
-> /server Freenode
18:44 -!- Irssi: Connecting to chat.freenode.net [84.240.3.129] port 7000
18:44 -!- Irssi: Connection to chat.freenode.net established
```

thiagomacieira:
Test 2: network with native IPv6:
```
20:46 -!- Irssi: Looking up chat.freenode.net
20:46 -!- Irssi: Connecting to chat.freenode.net [2610:150:4b0f::] port 7000
20:46 -!- Irssi: Connection to chat.freenode.net established
```

Thanks!

thiagomacieira:
Need to test 3 now: network with IPv6 ULA address but not global.

dequis:
@thiagomacieira does that third test cover the common case of users who don't actually have ipv6 support from their ISP, but still get ipv6 addresses before ipv4 in the first result of DNS resolution functions?

I've seen issues like that plenty of times (not in irssi) and my solution has been to tell users to disable ipv6 support at the OS level.

I wonder if there's a proper way to just avoid that kind of failures.

thiagomacieira:
The third test involves the complex rules of RFC 6724. The default rules from that RFC say that if you have an Unique Local IPv6 address, it's considered global and therefore the IPv6 addresses should be sorted first. The Linux default implementation is slightly different:

```
$ ip addrlabel
prefix ::1/128 label 0 
prefix ::/96 label 3 
prefix ::ffff:0.0.0.0/96 label 4 
prefix 2001::/32 label 6 
prefix 2001:10::/28 label 7 
prefix 3ffe::/16 label 12 
prefix 2002::/16 label 2 
prefix fec0::/10 label 11 
prefix fc00::/7 label 5 
prefix ::/0 label 1 
```

As you can see, there's an extra rule for fc00::/7, so there's a label difference beteen the ULA addresses and the global addresses.

I **think** the default behaviour on Linux will be to sort IPv4 before IPv6 if the only addresses are ULA. I will also need to test FreeBSD and OS X.

thiagomacieira:
A quick test on Linux shows that getaddrinfo() sorts IPv4 before IPv6 if I only have ULA addresses, regardless of whether that fc00::/7 rule is there.

thiagomacieira:
I don't have OS X or FreeBSD on an IPv6-enabled network, so I tried to simulate by adding the addresses to the interfaces. The behaviour is the same as on Linux.

dequis:
The issue i mentioned was exclusively in linux anyway. I'm fairly sure that's a misconfiguration in their setups, anyway, and a relatively rare issue at that (judging by how often i hear complaints about it). It went "I can't connect to X", "change the server address to this IP", "oh thanks it works!", "add this to sysctl to disable ipv6".

Just slightly worried that irssi users with that kind of broken ipv6 setup were immune to that problem before this patch, and they will start having problems with the next release. I wish I could understand this stuff more.

So yeah, mostly speaking out of ignorance here, don't consider this fear of mine a valid reason to block this PR, unless someone has something more concrete.

thiagomacieira:
Most of those issues were caused by broken DNS servers failing to answer the AAAA query in the first place. Those were common 15 years ago, but are quite rare nowadays. And besides, with AI_ADDRCONFIG, getaddrinfo() should not send an AAAA query unless there's at least one non-link-local IPv6 address assigned to an interface.

The other possible scenario is a routing problem: that is, if you do have an IPv6 default route but it goes to the black hole (no ICMPv6 "net unreachable" reply). Those are very, very rare.

Also, consider modern browsers like Firefox and Chromium do have IPv6 enabled by default and use getaddrinfo().

ailin-nemui:
so to summarise, is this working properly for you?

thiagomacieira:
Yes.

ailin-nemui:
I jumped the gun on this, sorry. `/connect -4 chat.freenode.net` will crash due to assertion failure

dequis:
This continues in #299

thiagomacieira:
Hmm... I've run into an issue now with this patch: it always tries to connect to the same address, even if that address is pointing to a server that is unreachable. This is a side effect of RFC 6724's Rule 9 "Use longest matching prefix", the sort order out of getaddrinfo is somewhat predictable. For example, these are the IPv6 addresses out of chat.freenode.net, sorted numerically:

chat.freenode.net has IPv6 address 2001:6b0:e:2a18:5054:ff:fe01:8119
chat.freenode.net has IPv6 address 2001:778:627f::1:0:49
chat.freenode.net has IPv6 address 2610:150:4b0f::
chat.freenode.net has IPv6 address 2a00:1a28:1100:11::42
chat.freenode.net has IPv6 address 2a00:1a28:1100:11::44
chat.freenode.net has IPv6 address 2a01:270:0:666f::1
chat.freenode.net has IPv6 address 2a01:4f8:100:92f1::2
chat.freenode.net has IPv6 address 2a02:2498:1:3a3:6ef0:49ff:fe44:bc07
chat.freenode.net has IPv6 address 2a02:2f0c:8000:19:62:231:75:133

Given that my IP address starts with 2601 (Comcast's block), that third address starting with 2610 is always sorted first, followed by the two starting with 2001, followed by the rest.

Not a problem, until that that server is unreachable. Now irssi will keep trying to connect there, every time:

```
00:01 -!- Irssi: Looking up chat.freenode.net
00:01 -!- Irssi: Reconnecting to chat.freenode.net [2610:150:4b0f::] port 7000 - use /RMRECONNS to abort
00:01 -!- Irssi: warning SSL handshake failed: No route to host
00:01 -!- Irssi: Connection lost to chat.freenode.net
00:11 -!- Irssi: Removed reconnection to server chat.freenode.net port 7000
00:11 -!- Irssi: Looking up chat.freenode.net
00:11 -!- Irssi: Reconnecting to chat.freenode.net [2610:150:4b0f::] port 7000 - use /RMRECONNS to abort
00:11 -!- Irssi: warning SSL handshake failed: No route to host
00:11 -!- Irssi: Connection lost to chat.freenode.net
00:20 -!- Irssi: Removed reconnection to server chat.freenode.net port 7000
00:20 -!- Irssi: Looking up chat.freenode.net
00:20 -!- Irssi: Reconnecting to chat.freenode.net [2610:150:4b0f::] port 7000 - use /RMRECONNS to abort
00:20 -!- Irssi: warning SSL handshake failed: No route to host
```

getaddrinfo() returned a list of 27 addresses, but irssi always tries the first one only. The best solution is that net_gethostbyname return multiple addresses and that irssi try to connect them in order. There's no need to wait 10 minutes to connect, even.

If that is not possible, a stop-gap solution is to make net_gethostbyname stateful. If the lookup is the same as the previous one, return the top of the stack in addrinfo results and remove it.

PR #292, opened 2015-09-21
Fix FS#905, mangled text when pasted line length exceeds 400

dequis:
http://bugs.irssi.org/index.php?do=details&task_id=905

Not using the patch from that ticket, the issue turned out to be that
(dest - last_lf_pos) returned number of unichr, not bytes, so that's 4
times less than what the size parameter of memmove() should be.

-----

Cherry picked from #277, that one needs rebasing so that the commit doesn't appear twice in the history.


GeertHauwaerts:
FS#905 is now closed.

PR #293, opened 2015-09-22
wip/config include

steven-jackson:
This is still a work in progress but would like some feed back and maybe some insight in to possible edge cases. It provides a way to include config files within config files, the main motivation was to be able to put the config in source control without exposing passwords.

I'm not entirely happy with how I've prevented circular includes, but it works - I keep a hash (acting as a set) of include names in the root CONFIG_REC and look up whether it's already been included. I'll think about this tomorrow.

I don't prevent two idential files with different paths from being included as there's no good portable way to do it and it doesn't provide much benefit, each file will only be included once anyway. GLib provides g_file_equal but it doesn't handle symbolic links.

Known issue - with the following set up the b alias isn't added, but the freenode server is.

config:
        include "aliases.config";
        include "servers.config";
        ...
 
aliases.config:
        aliases = {
                a = "abc;
                include "more_aliases.config";
        };
 
more_aliases.config:
        b = "def";
 
servers.config:
        servers = ( { include "server.freenode.config"; } );
 
server.freenode.config:
        address = "holmes.freenode.net";
        ...

### DIFF
--- a/src/lib-config/get.c
+++ b/src/lib-config/get.c
@@ -33,6 +33,14 @@ CONFIG_NODE *config_node_find(CONFIG_NODE *parent, const char *key)
 
 		if (node->key != NULL && g_ascii_strcasecmp(node->key, key) == 0)
 			return node;
+
+		/* the primary children of an include node should also be searched */
+		if (node->type == NODE_TYPE_INCLUDE) {
+			CONFIG_INCLUDE *inc = node->value;
+			node = config_node_find(inc->rec->mainnode, key);
+			if (node)
+				return node;
+		}
 	}
 
 	return NULL;
--- a/src/lib-config/get.c
+++ b/src/lib-config/get.c
@@ -20,15 +20,15 @@
 
 #include "module.h"
 
-CONFIG_NODE *config_node_find(CONFIG_NODE *node, const char *key)
+CONFIG_NODE *config_node_find(CONFIG_NODE *parent, const char *key)
 {
 	GSList *tmp;
 
-	g_return_val_if_fail(node != NULL, NULL);
+	g_return_val_if_fail(parent != NULL, NULL);
 	g_return_val_if_fail(key != NULL, NULL);
-	g_return_val_if_fail(is_node_list(node), NULL);
+	g_return_val_if_fail(is_node_list(parent), NULL);
 
-	for (tmp = node->value; tmp != NULL; tmp = tmp->next) {
+	for (tmp = parent->value; tmp != NULL; tmp = tmp->next) {
 		CONFIG_NODE *node = tmp->data;
 
 		if (node->key != NULL && g_ascii_strcasecmp(node->key, key) == 0)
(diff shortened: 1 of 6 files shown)


ailin-nemui:
is it because you only keep a single include, or because the parser only allows a single key?

steven-jackson:
I can't think of a reasonable use case for including the same file twice, maybe I'm wrong. It prevents the following from ruining the stack with recursion:

config:
include "2.config"

2.config:
include "config"

ailin-nemui:
ok slowly for me, why can you include files inside a server config but not inside aliases?

steven-jackson:
That's something I'll be trying to fix tonight, it isn't a design decision, it's a bug.

ailin-nemui:
I could imagine some hypothetical scenario where you use the same passwords on multiple servers and do 'include 'mypass.config'" several times. As long as the include structure is a tree it could be acceptable (but harder to verify I guess)

steven-jackson:
That's a good use case. I think if I just prevent child configs from including parents, that would allow the same password file to be used on multiple servers.

steven-jackson:
I've managed to support detecting child configs being modified on disk in order to prevent overwriting manual changes. This was done by moving the file checksumming and checks to config_write/config_open.

There's now a bit of a problem though, in order to fix the issue mentioned in the pull request (list items being contained in an include) some fairly horrible stuff is going to have to be tracked. Every node will have to keep track of its parent list. I've done the traversal code:

346 GSList *config_node_next(GSList *list)                                                                                                                                                                                                                                                                                                                            
347 {                                                                                                                                                                                                                                                                                                                                                                 
348     CONFIG_NODE *node;                                                                                                                                                                                                                                                                                                                                            
349                                                                                                                                                                                                                                                                                                                                                                   
350     if (list->next) {                                                                                                                                                                                                                                                                                                                                             
351         node = list->next->data;                                                                                                                                                                                                                                                                                                                                  
352         if (node->type == NODE_TYPE_INCLUDE) {                                                                                                                                                                                                                                                                                                                    
353             CONFIG_INCLUDE *inc = node->value;                                                                                                                                                                                                                                                                                                                    
354             if (inc && inc->rec && inc->rec->mainnode)                                                                                                                                                                                                                                                                                                            
355                 list->next = inc->rec->mainnode->value;                                                                                                                                                                                                                                                                                                           
356         } else                                                                                                                                                                                                                                                                                                                                                    
357             list = list->next;                                                                                                                                                                                                                                                                                                                                    
358     } else {                                                                                                                                                                                                                                                                                                                                                      
359         node = list->data;                                                                                                                                                                                                                                                                                                                                        
360         if (node->parent_list) {                                                                                                                                                                                                                                                                                                                                  
361             list = node->parent_list->next;                                                                                                                                                                                                                                                                                                                       
362         }                                                                                                                                                                                                                                                                                                                                                         
363     }                                                                                                                                                                                                                                                                                                                                                             
364                                                                                                                                                                                                                                                                                                                                                                   
365     return config_node_first(list);

Actually setting a node's parent list is next, but this is starting to get pretty ugly.

ailin-nemui:
I can't say if this mightn't turn too ugly... :-(

steven-jackson:
I'm not happy with how complex this has become. I'll implement another solution when I have some time.

ailin-nemui:
thanks for the effort!

PR #298, opened 2015-09-23
Add multi-prefix to list of capabilities to request

dequis:
Turns out event_names_list() in irc-nicklist.c already handles this.

event_who() just ignores it, which is probably a good idea since some of
the irc servers I tested this with have a bug that results in sending
multiple prefixes in the NAMES reply but not in the WHO one (they were
forks of ircd-hybrid before 7.3.0)

And NAMES always happens, anyway. WHO is omitted sometimes for huge
channels.

--------------------

Absurdly easy commit, the hardest part was digging in the svn history of ircd-hybrid to figure out if they ever fixed it. Which isn't really a part of making this commit. Thanks nei for pointing out that it might already be implemented.


ahf:
Hah, cute. Good job!

Issue #220, opened 2015-02-11
DCC GET fails when saving to CIFS location


Revenent:
If the download path (dcc_download_path) is set to a location that is mounted with the CIFS filesystem, the GET will fail with a busy error.

According to the bug report @ Debian (https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=696963), this is due to temphandle in sig_dccget_connected (dcc-get.c) being left open while other operations are attempted against the file.

The Debian bug report has a patch written by Robert C Jensen at the bottom.

Hoping this gets in soon, since this would be really useful on a Raspberry Pi w/ a backing Windows machine as the NAS.

dequis:
Also references http://bugs.irssi.org/index.php?do=details&task_id=867

Patch:

```diff
Description: Call close() on temporary file descriptor so we don't get ETXTBSY
 when dcc_download_path is an ntfs mount
Author: Robert C Jensen <jensen.r@husky.neu.edu>
--- a/src/irc/dcc/dcc-get.c
+++ b/src/irc/dcc/dcc-get.c
@@ -223,6 +223,8 @@
 		else
 			ret = fchmod(temphandle, dcc_file_create_mode);
 
+		close(temphandle);
+
 		if (ret != -1) {
 			ret = link(tempfname, dcc->file);
 
@@ -246,7 +248,6 @@
 
 		/* close/remove the temp file */
 		ret_errno = errno;
-		close(temphandle);
 		unlink(tempfname);
 		g_free(tempfname);
 
```

GeertHauwaerts:
The patch is applied and FS#867 is closed as well.

PR #301, opened 2015-09-24
dcc-get: close() the temp fd so we don't get ETXTBSY in ntfs mounts

dequis:
Patch from [debian bug 696963][1]

Fixes bug #220 and [flyspray bug 867][2]

[1]: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=696963
[2]: http://bugs.irssi.org/index.php?do=details&task_id=867

-----------------

The bugs talk about ntfs and cifs mounts but it's just the windows filesystem limitation of only allowing a fd to be open by a single thing at the same time. I couldn't actually reproduce that with a local ntfs filesystem or a cifs mount from a remote ext4 partition. I'm lacking some windows here.

Either way the patch is sensible and straightforward and the original author claims it fixes the issue so i'll trust him on that.

### DIFF
--- a/src/irc/dcc/dcc-get.c
+++ b/src/irc/dcc/dcc-get.c
@@ -226,6 +226,8 @@ void sig_dccget_connected(GET_DCC_REC *dcc)
 		else
 			ret = fchmod(temphandle, dcc_file_create_mode);
 
+		close(temphandle);
+
 		if (ret != -1) {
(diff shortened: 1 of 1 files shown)


GeertHauwaerts:
FS#867 has been closed.

dequis:
Also thanks @erg0dic for the original patch (github figured out the name from the email, didn't see it before)

PR #303, opened 2015-09-24
Applied patch from fs#275 to make /hilight -mask -line work properly

vague666:


GeertHauwaerts:
Did you test this, or just imported from FS?

vague666:
I have tested it and it appears to work as intended
/jari

src/fe-common/core/hilight-text.h
```diff
@@ -7,7 +7,7 @@
 
 #include "formats.h"
 
-typedef struct _HILIGHT_REC HILIGHT_REC;
+//typedef struct _HILIGHT_REC HILIGHT_REC;
```
dequis: what

vague666: typedef struct _HILIGHT_REC HILIGHT_REC is defined in formats.h so that HILIGHT_REC can be used in _TEXT_DEST_REC, maybe the author of the patch should've removed that line all together but all I did was lift the patch into github and create a PR for it :)

ailin-nemui:
this seems sane to me, apart from removing the fixme comments what changes are needed here?

ailin-nemui:
@vague666 please remove the comments left in source code so we can get this in

PR #306, opened 2015-09-27
Implement paste detection via the bracketed paste mode 2: bracket pasterer

### LINKED ISSUE #9: Bracketed paste mode (alternative to time based paste detection)
Some xterm compatible terminal emulators support reporting usage of their 'paste' feature through special control codes. This is more reliable than time based paste detection, which can fail in laggy connections.

Short description in the xfree86 docs: http://www.xfree86.org/current/ctlseqs.html#Bracketed%20Paste%20Mode

Introduction and simple explanation on how to implement it: http://cirw.in/blog/bracketed-paste

Weechat implementation details for reference: http://dev.weechat.org/post/2012/03/10/Bracketed-paste-mode

Weechat initial implementation commit for reference: https://github.com/weechat/weechat/commit/152394689a5fe8d43e3e13ca37ffec5e65a44e84#diff-21 (GPLv3)

dequis:
Follow up of #277 since i kinda took over, including https://github.com/LemonBoy/irssi/pull/2

Rebased to current master to remove the commit that turned into #292.

Closes #9


dequis:
beep

src/fe-text/gui-readline.c
```diff
@@ -647,21 +725,37 @@ static void sig_input(void)
...
 			paste_timeout_id = g_timeout_add(paste_detect_time, paste_timeout, NULL);
-		} else {
+		} else if (!paste_bracketed_mode) {
 			int i;
 
 			for (i = 0; i < paste_buffer->len; i++) {
 				unichar key = g_array_index(paste_buffer, unichar, i);
 				signal_emit("gui key pressed", 1, GINT_TO_POINTER(key));
+
+				if (paste_bracketed_mode) {
```
ailin-nemui: at which point does paste_bracketed_mode get TRUE after the check on line 744?

src/fe-text/gui-readline.c
```diff
@@ -306,7 +315,14 @@ static void paste_send(void)
 		}
 	}
 
-	gui_entry_set_text(active_entry, str->str);
+	if (paste_was_bracketed_mode) {
+		/* the text before the bracket end should be sent along with the rest */
```
ailin-nemui: why's that?

src/fe-text/gui-readline.c
```diff
@@ -634,6 +658,60 @@ static gboolean paste_timeout(gpointer data)
...
+	/* remove the rest, including the trailing sequence chars */
+	g_array_set_size(paste_buffer, i);
+
+	/* decide what to do with the buffer */
+	paste_timeout(NULL);
+
+	paste_bracketed_mode = FALSE;
+}
+
+static void paste_bracketed_middle()
```
ailin-nemui: *this function has not been reviewed

src/fe-text/gui-readline.c
dequis: During the `gui key pressed` signal, `key_paste_start` function specifically.

>/* just enabled by the signal, remove what was processed so far */

src/fe-text/gui-readline.c
dequis: if you have paste something like this:

     \x1b[200~text1
    text2 text3 text4
    text5 text6 text7
    text8 text9\x1b[201~

you wouldn't want to answer the confirmation message, send the first three lines to the channel, and be left with `text8 text9` in the input box

ailin-nemui:
I just tried this patch and if I paste something with bracketed mode enabled, my input line looks like this: hi <pasted>~xx and input is blocked

dequis:
Managed to reproduce that behavior by having paste_detect_time set to nonzero, and pasting one line only. With more than one line the markers are visible but the input isn't blocked.

dequis:
Fixed the issues mentioned above just by avoiding timeout-based paste detection, or, rather, preferring bracketed mode if the user enabled that.

I was working on a [more complex diff][1] that sorta unifies the way `paste_timeout()` emits key signals, which worked fine for single-line pastes, but when i got to `paste_send()` i found myself writing code to strip the bracketed paste markers from the paste buffer, because the timeout based code likes to include raw control codes in its mess.

And that's where I realized the whole thing is dumb and I shouldn't give be giving priority to the timeout based code, requiring the users to set `paste_detect_time` to 0 to be able to use this feature at all.

[1]: http://dump.dequis.org/ElXSR.txt

dequis:
Also pushed another commit that fixes the main issue i've found while testing this in 'production' - if you paste `a\nb` (as opposed to `a\nb\n`) the line count said 1, not 2, even though two lines are sent.

This only applies to the bracketed paste code, which considers `a\nb` a single thing and won't leave the last part in the input box

src/fe-text/gui-readline.c
```diff
@@ -322,6 +338,12 @@ static void paste_flush(int send)
 		paste_send();
 	g_array_set_size(paste_buffer, 0);
 
+	/* re-add anything that may have been after the bracketed paste end */
+	if (paste_buffer_rest->len) {
```
ahf: I'd write `if (paste_buffer_rest->len > 0)` here. Nitpick though.

PR #309, opened 2015-09-29
Use _POSIX_HOST_NAME_MAX instead of HOST_NAME_MAX.

ahf:
Fixes compilation on OS X.

### DIFF
--- a/src/irc/core/irc-expandos.c
+++ b/src/irc/core/irc-expandos.c
@@ -27,6 +27,8 @@
 #include "irc-channels.h"
 #include "nicklist.h"
 
+#include <limits.h>
+
 static char *last_join;
 
 /* last person to join a channel you are on */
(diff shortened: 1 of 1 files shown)


jillest:
Using the _POSIX_ versions of limits is in some sense wrong since they are the minimum (in this case) or maximum limit permitted by the standard. In that sense it is similar to hard-coding 255 (or 256, see below). In practice, though, it will be enough except in pathological situations where people try to break things.

Also, {HOST_NAME_MAX} does not include the terminating NUL, so 1 should be added for a buffer size.

ahf:
I'll hardcode it to 256 - all the functions requires a size boundary anyway. Thanks for the review, @jillest

dequis:
```
#ifndef HOST_NAME_MAX
#define HOST_NAME_MAX 256
#endif
```

?

PR #311, opened 2015-09-29
[WIP] TLS enhancements

ahf:
WIP, don't merge - only reviews. Please nitpick.


src/core/network-openssl.c
```diff
@@ -610,18 +588,167 @@ int irssi_ssl_handshake(GIOChannel *handle)
...
+				tls_cert_rec_append_issuer_entry(cert_rec, tls_cert_entry_rec);
+			}
+
+			tls_rec_append_cert(tls_rec, cert_rec);
+		}
+	}
+
+	BIO *b;
+	char buffer[128];
+	size_t len;
```
dequis: These should be at the top of the scope they are defined in

src/core/network-openssl.c
```diff
@@ -610,18 +588,167 @@ int irssi_ssl_handshake(GIOChannel *handle)
...
+	}
+
+	tls_rec_set_public_key_size(tls_rec, EVP_PKEY_bits(pubkey));
+
+	chain = SSL_get_peer_cert_chain(chan->ssl);
+	if (chain != NULL) {
+		int i;
+		TLS_CERT_REC *cert_rec = NULL;
+		for (i = 0; i < sk_X509_num(chain); i++) {
+			cert_rec = tls_cert_create_rec();
```
dequis: This should be below the variable declarations

dequis:
irssi_ssl_handshake is just way too big. Seriously consider splitting it. Also, 

>man SSL_get_peer_certificate
>
>The reference count of the X509 object is incremented by one, so that it will not be destroyed when the session containing the peer certificate is freed. The X509 object must be explicitly freed using X509_free()

There's a lot of new branches that return without X509_free().

src/fe-common/core/fe-tls.c
```diff
@@ -0,0 +1,85 @@
...
+			printformat(server, NULL, MSGLEVEL_CLIENTNOTICE, TXT_TLS_CERT_NAMED_ENTRY, issuer_data->name, issuer_data->value);
+		}
+	}
+
+	printformat(server, NULL, MSGLEVEL_CLIENTNOTICE, TXT_TLS_PROTOCOL_VERSION, tls->protocol_version);
+	printformat(server, NULL, MSGLEVEL_CLIENTNOTICE, TXT_TLS_CIPHER, tls->cipher_size, tls->cipher);
+
+	printformat(server, NULL, MSGLEVEL_CLIENTNOTICE, TXT_TLS_VALIDATION_PERIOD, tls->not_before, tls->not_after);
+
+	printformat(server, NULL, MSGLEVEL_CLIENTNOTICE, TXT_TLS_PUBLIC_KEY, tls->public_key_size, tls->public_key_algorithm);
```
dequis: Not sure if it's worth doing anything about this, but all these `printformat(server, NULL, MSGLEVEL_CLIENTNOTICE,` are a bit too noisy visually.

Wrapping the lines so that i don't have to scroll horizontally to read the important part would be nice too, i guess.

src/core/tls.c
```diff
@@ -0,0 +1,200 @@
...
+	tls_cert_rec->issuer = g_slist_append(tls_cert_rec->issuer, tls_cert_entry_rec);
+}
+
+void tls_cert_rec_free(TLS_CERT_REC *tls_cert_rec)
+{
+	if (tls_cert_rec == NULL)
+		return;
+
+	if (tls_cert_rec->subject != NULL) {
+		g_slist_free_full(tls_cert_rec->subject, (GDestroyNotify)tls_cert_entry_rec_free);
```
dequis: `g_slist_free_full()` was introduced in glib 2.28. Current requirement is 2.16. Bumping it would mean dropping support for debian 6.0 (oldstable, squeeze, still supported). I suggest just iterating over the list.

dequis:
Might be cool to land the first five commits to master since they are trivial changes that add noise to the diff. Something like `git merge b64444b` should do.

src/core/network-openssl.c
```diff
@@ -561,6 +529,186 @@ static GIOChannel *irssi_ssl_get_iochannel(GIOChannel *handle, int port, SERVER_
...
+
+			case EVP_PKEY_EC:
+			{
+				EC_KEY *ec = EVP_PKEY_get1_EC_KEY(ephemeral_key);
+				int nid;
+				const char *cname;
+				nid = EC_GROUP_get_curve_name(EC_KEY_get0_group(ec));
+				EC_KEY_free(ec);
+				cname = OBJ_nid2sn(nid);
+				char *ephemeral_key_algorithm = g_strdup_printf("ECDH: %s", cname);
```
dequis: Another variable declaration in the middle of a scope

src/fe-common/core/fe-tls.c
```diff
@@ -0,0 +1,81 @@
...
+
+		GSList *subject = NULL;
+		for (subject = tls_cert_rec->subject; subject != NULL; subject = subject->next) {
+			TLS_CERT_ENTRY_REC *subject_data = subject->data;
+			printformat(server, NULL, MSGLEVEL_CLIENTNOTICE, TXT_TLS_CERT_NAMED_ENTRY, subject_data->name, subject_data->value);
+		}
+
+		printformat(server, NULL, MSGLEVEL_CLIENTNOTICE, TXT_TLS_CERT_ISSUER_HEADER);
+
+		GSList *issuer = NULL;
```
dequis: Another variable declaration in the middle of a scope, same thing with subject above.

dequis:
Regarding SSL_get_server_tmp_key, i was wondering "how does this ifdef even work?" (answer: it's a macro, duh) and while searching I got this

https://boringssl.googlesource.com/boringssl/+/7d97fd3a91cbd044cea5bb433a0e9cbb2c6e71ad%5E!/

Commit log:

>Remove SSL_get_server_tmp_key.
>
>It's never used and is partially broken right now; EVP_PKEY_DH doesn't work.

Review comment:

>Might want some API like this eventually (like that Opera CL), but since this incarnation is broken and unused, it seems simpler to go our own way there rather than make EVP_PKEY_DH work again

It's not openssl, but it's the serious business fork.

I don't know enough to tell the implications of this, but uhm, you wrote that thing that uses it, you probably do.

dequis:
https://www.imperialviolet.org/2011/05/04/pinning.html

>Why public key hashes, not certificate hashes?
>
>In general, hashing certificates is the obvious solution, but the wrong one. The problem is that CA certificates are often reissued: there are multiple certificates with the same public key, subject name etc but different extensions or expiry dates

It suggests hashing SubjectPublicKeyInfo (which includes the public key itself) instead of the whole certificate.

Also part of their proposed standard: https://tools.ietf.org/html/rfc7469#section-2.4

Not all of this applies since our implementation is mainly to handle self-signed certificates, and chrome's seems like a way to enhance CA-issued certificate validation.

ahf:
This is now being tracked in #557

PR #312, opened 2015-09-29
Switch to modern Travis CI infrastructure

### LINKED ISSUE #310: Switch to modern Travis CI infrastructure
We are currently using the old, slow, Travis CI infrastructure.

We need to do the following to get on the new one:

- Explicitly list packages we depend upon from apt. It doesn't look like there is an option for doing build dependencies easily like we do now. See: http://docs.travis-ci.com/user/migrating-from-legacy/#Adding-APT-Packages
- Use `--prefix=$HOME/irssi-build` (since we don't have `sudo` we cannot do `make install` to `/`).
- Add `sudo: false` in `.travis.yml`.  See: http://docs.travis-ci.com/user/migrating-from-legacy/

While you are at it, please enable `--enable-silent-rules` for the call to `autogen.sh` in `.travis.yml`. This should make warnings stand out more clearly.

Jellyfrog:
Had to switch from `lynx` to `elinks` due to; https://github.com/travis-ci/apt-package-whitelist/issues/630

Fixes #310

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+sudo: false
 language: perl
 perl:
     - "5.20-shrplib"
@@ -7,19 +8,22 @@ env:
     - CC=clang
     - CC=gcc
 
+addons:
+    apt:
+        packages:
+            - libperl-dev
+            - elinks
+
 before_install:
-    - sudo apt-get update -qq
     - perl -V
-    - sudo apt-get build-dep -qq irssi
-    - sudo apt-get install -qq lynx
 
 install: true
 
 script:
-    - ./autogen.sh --with-proxy --with-bot --with-perl=module
+    - ./autogen.sh --with-proxy --with-bot --with-perl=module --prefix=$HOME/irssi-build --enable-silent-rules
     - cat config.log
     - make
-    - sudo make install
+    - make install
 
 notifications:
     irc:


.travis.yml
```diff
@@ -7,19 +8,27 @@ env:
     - CC=clang
     - CC=gcc
 
+addons:
+    apt:
+        packages:
+            - libc6
+            - libglib2.0-0
+            - libssl1.0.0
+            - libtinfo5
```
dequis: Pretty sure these first four deps can be nuked right away.

Jellyfrog: I simply copied the list from ubuntu's deps.
I'll update and we see.

dequis:
>checking for working Perl support... error linking with perl libraries, building without Perl

In the system-perl ones.

https://travis-ci.org/irssi/irssi/jobs/82764669#L451

>Error linking with perl libraries: -Wl,-E -fstack-protector -L/usr/local/lib -L/usr/lib/perl/5.14/CORE -lperl -ldl -lm -lpthread -lcrypt: /usr/bin/ld: cannot find -lperl
>
>clang: error: linker command failed with exit code 1 (use -v to see invocation)
>
>Try: sudo apt-get install libperl-dev

Jellyfrog:
Fixed now.
Maybe enable true color and DANE?

dequis:
Hmm, no benefit in enabling that here IMO, I'm just going to merge this as-is.

Thanks for your contribution!

PR #313, opened 2015-10-01
Fix #45 Make it easy to delete default channels, servers and networks

### LINKED ISSUE #45: make it easy to delete default channels, servers and networks
irssi comes with a lot of networks and servers in the default config. They are there to make it easier for new users to get started. 

Please make it easier to remove all networks and servers to get a "clean slate".

isundil:
Removing network will now also remove all attached servers

### DIFF
--- a/src/core/servers-setup.c
+++ b/src/core/servers-setup.c
@@ -333,6 +333,24 @@ server_create_conn(int chat_type, const char *dest, int port,
 				chatnet, password, nick);
 }
 
+GSList *server_setup_find_chatnet(const char *chatnet)
+{
+	GSList *servers;
+	GSList *tmp;
+
+	g_return_val_if_fail(chatnet != NULL, NULL);
+
+	servers = NULL;
+	for (tmp = setupservers; tmp != NULL; tmp = tmp->next) {
+		SERVER_SETUP_REC *rec = tmp->data;
+
+		if (g_ascii_strcasecmp(rec->chatnet, chatnet) == 0)
+			servers = g_slist_append(servers, rec);
+	}
+
+	return servers;
+}
+
 /* Find matching server from setup. Try to find record with a same port,
    but fallback to any server with the same address. */
 SERVER_SETUP_REC *server_setup_find(const char *address, int port,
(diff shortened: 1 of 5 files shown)


ailin-nemui:
what about attached channels?

dequis:
You're iterating over `server_setup_find_chatnet` with a for loop, building a new linked list with only some entries, returning that linked list, iterating over the returned linked list to remove each element, and not freeing the resulting linked list at any point.

You should iterate in a similar way to what `server_remove_channels` in src/core/servers.c does, keeping a `GSList *next` pointer and doing `tmp = next` instead of `tmp = tmp->next`

isundil:
I added changes with your considerations.
Channels are also removed now.

dequis:
Not sure if i like the fact that both functions look copypasted, but, different types. I dunno. Probably okay.

isundil:
I know it look copypasted, but since the servers and channels are both linked-list their removal function should look alike.
The `channel_setup_find' and `server_setup_find' functions from these files are quite similar too.

dequis:
This looks good, but if anyone else can have a look and merge it, would be cool.

src/core/servers-setup.c
```diff
@@ -505,6 +505,21 @@ void server_setup_add(SERVER_SETUP_REC *rec)
 	signal_emit("server setup updated", 1, rec);
 }
 
+void server_setup_remove_chatnet(const char *chatnet)
+{
+	GSList *tmp, *next;
+
+	g_return_if_fail(chatnet != NULL);
+
+	for (tmp = setupservers; tmp != NULL; tmp = next) {
```
ailin-nemui: nit: why not the usual `tmp = tmp->next`?

dequis: See my comment above, iterating that way would result in use after free of tmp

isundil: Because I think tmp may be unallocated by `g_slist_remove` (at line 180)

ailin-nemui: I see

PR #315, opened 2015-10-02
Don't set the usermode field if blank

LemonBoy:
Fixes FS#919


dequis:
Proper link to the bug: http://bugs.irssi.org/index.php?do=details&task_id=919

The patch attached to the original bug is just terrible. This PR is way better. I could nitpick about that `char` vs `const char` but i think it's just fine.

GeertHauwaerts:
The [FS#919](http://bugs.irssi.org/index.php?do=details&task_id=919) has been closed.

PR #318, opened 2015-10-02
Fix the display of utf8 sequences in the gui

LemonBoy:
term_addstr() had a long-standing fixme that suggested it didn't
take into account the string encoding when calculating the string
length.
The BIG5 code path is untested.

### DIFF
--- a/src/fe-text/term-terminfo.c
+++ b/src/fe-text/term-terminfo.c
@@ -522,15 +522,60 @@ void term_add_unichar(TERM_WINDOW *window, unichar chr)
 	}
 }
 
-void term_addstr(TERM_WINDOW *window, const char *str)
+int term_addstr(TERM_WINDOW *window, const char *str)
 {
-	int len;
+	int i, len, raw_len;
+	unichar *tmp;
+	const char *ptr;
 
 	if (vcmove) term_move_real();
-	len = strlen(str); /* FIXME utf8 or big5 */
+
+	raw_len = strlen(str);
+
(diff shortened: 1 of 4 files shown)


src/fe-text/term-terminfo.c
```diff
@@ -522,15 +522,36 @@ void term_add_unichar(TERM_WINDOW *window, unichar chr)
...
+
+	len = 0;
+	raw_len = strlen(str);
+
+	/* The string length depends on the terminal encoding */
+
+	ptr = str;
+
+	if (term_type != TERM_TYPE_BIG5) {
+	    while (*ptr != '\0') {
```
dequis: 4 space indents

dequis:
I heard on irc that one issue with this was that it didn't use wcwidth. Looks like it changed since then, this looks good to me.

src/fe-text/term-terminfo.c
```diff
@@ -522,15 +522,36 @@ void term_add_unichar(TERM_WINDOW *window, unichar chr)
...
-	len = strlen(str); /* FIXME utf8 or big5 */
+
+	len = 0;
+	raw_len = strlen(str);
+
+	/* The string length depends on the terminal encoding */
+
+	ptr = str;
+
+	if (term_type != TERM_TYPE_BIG5) {
```
ailin-nemui: you must check for term_type == TERM_TYPE_UTF8 instead

ailin-nemui:
lets see how this goeas

PR #319, opened 2015-10-02
Flush the dirty buffer to disk

LemonBoy:
Original FS ticket [1]
Original patch [2]

[1] http://bugs.irssi.org/index.php?do=details&task_id=781
[2] http://bugs.irssi.org/index.php?getfile=238


GeertHauwaerts:
The [FS#781](http://bugs.irssi.org/index.php?do=details&task_id=781) has been closed.

PR #320, opened 2015-10-02
Fix warnings

ahf:
Fix some of the warnings we have left before being able to enable -Werror for Travis.


dequis:
Maybe the `type` variable in `cmd_toggle` should be SettingType instead of int (the diff doesn't reach that part to comment on it). They are compatible types so this is just a nitpick because I felt forced to try really hard to find something to complain about. Otherwise, looks good.

dequis:
Could get this one out of the way and merge it.

PR #330, opened 2015-10-06
Add xterm's keypad enter, meta-O-M to "key return" bindings

### LINKED ISSUE #327: Numeric keypad "Enter" key stopped working
Well, the issue title summarises it well: you can't send messages using the Enter key that there is in the numeric keypad since some time ago.

dequis:
From the 'kent' terminfo entry. Also applies to putty.

Fixes #327

(edited title, was meta2-M before)

### DIFF
--- a/src/fe-text/gui-readline.c
+++ b/src/fe-text/gui-readline.c
@@ -1020,6 +1020,8 @@ void gui_readline_init(void)
 	key_bind("key", NULL, "meta2-5F", "cend", (SIGNAL_FUNC) key_combo);
 	key_bind("key", NULL, "meta2-1;5F", "cend", (SIGNAL_FUNC) key_combo);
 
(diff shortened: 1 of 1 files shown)


ailin-nemui:
conflicts with mouse (any script using xterm mouse)

dequis:
Thanks.

>Irssi::command("/^bind meta-[M /mouse_xterm");

hm. This one is meta-OM.

dequis:
Updated, had to use meta-O-M instead of meta-OM.

-----

Debugging note to self: breakpoints in each `g_free(combo);` of `key_pressed()` of keyboard.c, running `p combo` every time they are hit. The first one is hit with unknown key combos, the second one is hit with valid ones.

Not quite sure why `/bind meta-[M` works but `/bind meta-OM` doesn't, both are seen respectively as `^[-[-M` and `^[-O-M` in those breakpoints.

PR #331, opened 2015-10-06
[WIP] Irssiproxy SSL support

dequis:
Based off a patch from "Christian Sachs", http://bugs.irssi.org/index.php?do=details&task_id=645

I'd like to credit the first commit properly, but couldn't find an email address.

Quoting the ticket:

>I've created this patch (on SVN/r5021) which adds SSL encryption to irssiproxy. (The proxy then acts as a SSL server, the client used to connect to it will need to support SSL.)
>It's been tested on Linux and Mac OS X. Unless someone stumbles upon bugs I've missed, maybe the patch could be merged.
>I know that adding a small wrapper might not be the most elegant solution, but taking a look at the existing irssi SSL code makes it appear pretty client centric;
>I assumed it might be more adequate to add this by changing the proxy code alone, not the basic network code.

[...]

>It's enabled by compiling irssi SSL enabled; encryption then needs a server certificate and can be enabled network-wise.
>One can switch SSL on by adding `:/path/to/certkey.pem` to the `irssiproxy_ports` setting.
>e.g. `/set irssiproxy_ports QuakeNet=7777:/home/user/.irssi/servercert.pem`
>The certificate/key file must contain both cert&key, and be in PEM format. If one is familiar with PKI, obviously a proper cert can be used.
>For basic security, a simple cert/key can be made using the following command: `openssl req -new -x509 -nodes -newkey rsa:2048 -keyout yourcert.pem -out yourcert.pem -days 365`
>(Obviously checking the cert for trustability must be disabled in this case on the client side.)

I just rebased, applied a bunch of style fixes, removed all the openssl ifdefs, simplified code slightly, and fixed the blatantly wrong usage of `SSLv3_server_method()`. Thanks openssl! So intuitive.

Just adding it here to not forget about it (most of this stuff is from a few months ago), and to review it some more.

Not ~~significantly~~ tested, not even significantly reviewed by myself either. The fixes I applied are all low hanging fruit.

Review comment from coekie:

>It seems strange to me that the path to the certificate is per network, and not just one global setting though. Is there really a use for having different certificates for different proxy ports; or only having ssl enabled for some of them?

### DIFF
--- a/src/irc/proxy/dump.c
+++ b/src/irc/proxy/dump.c
@@ -33,37 +33,34 @@
 
 void proxy_send(CLIENT_REC *client, char *d, int l)
 {
-#ifdef HAVE_OPENSSL
 	if(client->listen->use_ssl) {
 		SSL_write(client->ssl, d, l);
-	} else 
-#endif
-		net_sendbuffer_send(client->handle, d, l);
+		return;
+	}
+	net_sendbuffer_send(client->handle, d, l);
 }
 
 int proxy_readline(CLIENT_REC *client, char **str)
 {
-#ifdef HAVE_OPENSSL
 	if(client->listen->use_ssl) {
 		char tmpbuf[2048];
 		int recvlen = 0;
-        
+
 		recvlen = SSL_read(client->ssl, tmpbuf, sizeof(tmpbuf));
 		if(recvlen > 0) {
 			return line_split(tmpbuf, recvlen, str, &client->handle->readbuffer);
 		} else {
 			int err;
 			err = SSL_get_error(client->ssl, recvlen);
-			/* READ/WRITE are not really errors, they just indicate that atm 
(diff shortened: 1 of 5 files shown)


ailin-nemui:
@dequis are you instrested in working on this?

dequis:
Not really, at least not for now, i'd just like to complete it for the sake of completing it, but I have bigger priorities now. Feel free to take over, or ask me again in a month.

ailin-nemui:
- [bdo#655311](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=655311)

dequis:
This lives in https://github.com/irssi/irssi/tree/orphaned/irssiproxy-ssl now

csachs:
Oh, wow, didn't expect my patch would make it into … or nearly into irssi after 6 years :D
I'd offer help but I'm not an active IRC user anymore … is anything particular missing/not right yet?

PR #332, opened 2015-10-06
Make /hilight print -line for hilights where it's specified

vague666:


ailin-nemui:
please create a PR with this code:

```
	if (rec->nick && rec->word) { /* default case, no option */ }
	else if (rec->nick)
		g_string_append(options, "-nick ");
	else if (rec->word)
		g_string_append(options, "-word ");
	else
		g_string_append(options, "-line ");

PR #338, opened 2015-10-10
Add MODIFY to /network, /server and /channel

vague666:
Since people are confused about ADD working as modify, lets make it more clear


dequis:
Dunno if i like this idea, but at least document that ADD does the same thing if the network already exists, instead of removing that line completely.

Also MODIFY should fail if the network/server/channel doesn't exist instead of being exactly the same thing as add

vague666:
Yeah, I was thinking about that, but since it is just an alias to make it more clear to users I thought the duplication of code would be more work than warranted
But sure, I can add cmd_server_modify & co if you want that.
Or I can amend the help files with info that modify is just an alias for add

ahf:
I'm not fully convinced about this. I think this is a documentation issue.

vague666:
It is a documentation issue also, but adding modify doesn't add much code and it would make it more clear to users what to do. This PR is mostly help modifications but it's easy to overlook documentation when you are looking for a particular thing.

ailin-nemui:
I'm not very much in favour of this unless you actually modify `MODIFY` to disable its `ADD` functionality. (I'd keep ADD as fully functional modify for backward compat)

ailin-nemui:
I like the direction your code is taking, however you will need to refactor the common code parts into static methods instead of using copypasta

ailin-nemui:
hi @vague666 , I really like the change you're proposing even if it is conceptually minor. however there is still work left before I can accept it. You know what I've been thinking about may be better instead:

- take the existing _add functions, rename them to from cmd_..._add to ..._add_modify with a final argument gboolean add
- only slightly modify them each, changing `if (rec == NULL) {` to `if (rec == null && !add) { error out } else if (rec == NULL && add) { original code goes on`
- mask each line that should not be applicable to modify in such a way
- create cmd_..._add and cmd_..._modify that call ..._add_modify(..., true/false)

That would also immediately get rid of some of the code duplication I marked with A and B

src/fe-common/irc/fe-ircnet.c
```diff
@@ -168,12 +140,90 @@ static void cmd_network_add(const char *data)
...
+		printformat(NULL, NULL, MSGLEVEL_CLIENTNOTICE, IRCTXT_NETWORK_NOT_FOUND, name);
+	} else {
+		if (g_hash_table_lookup(optlist, "nick")) g_free_and_null(rec->nick);
+		if (g_hash_table_lookup(optlist, "user")) g_free_and_null(rec->username);
+		if (g_hash_table_lookup(optlist, "realname")) g_free_and_null(rec->realname);
+		if (g_hash_table_lookup(optlist, "host")) {
+			g_free_and_null(rec->own_host);
+                        rec->own_ip4 = rec->own_ip6 = NULL;
+		}
+
```
ailin-nemui: B

src/fe-common/irc/fe-ircnet.c
```diff
@@ -168,12 +140,90 @@ static void cmd_network_add(const char *data)
...
+	} else {
+		if (g_hash_table_lookup(optlist, "nick")) g_free_and_null(rec->nick);
+		if (g_hash_table_lookup(optlist, "user")) g_free_and_null(rec->username);
+		if (g_hash_table_lookup(optlist, "realname")) g_free_and_null(rec->realname);
+		if (g_hash_table_lookup(optlist, "host")) {
+			g_free_and_null(rec->own_host);
+                        rec->own_ip4 = rec->own_ip6 = NULL;
+		}
+	}
+
```
ailin-nemui: B

src/fe-common/core/fe-server.c
```diff
@@ -189,22 +156,107 @@ static void cmd_server_add(const char *data)
...
+		if (value != NULL && *value != '\0') rec->port = atoi(value);
+
+		if (*password != '\0') g_free_and_null(rec->password);
+		if (g_hash_table_lookup(optlist, "host")) {
+			g_free_and_null(rec->own_host);
+			rec->own_ip4 = rec->own_ip6 = NULL;
+		}
+
+		if (*password != '\0' && g_strcmp0(password, "-") != 0) rec->password = g_strdup(password);
+
```
ailin-nemui: A

src/fe-common/core/fe-server.c
```diff
@@ -189,22 +156,107 @@ static void cmd_server_add(const char *data)
...
+		if (*password != '\0') g_free_and_null(rec->password);
+		if (g_hash_table_lookup(optlist, "host")) {
+			g_free_and_null(rec->own_host);
+			rec->own_ip4 = rec->own_ip6 = NULL;
+		}
+	}
+
+	if (*password != '\0' && g_strcmp0(password, "-") != 0) rec->password = g_strdup(password);
+
```
ailin-nemui: A

src/fe-common/core/fe-channels.c
```diff
@@ -284,21 +296,60 @@ static void cmd_channel_add(const char *data)
...
+
+	rec = channel_setup_find(channel, chatnet);
+	if (rec == NULL)
+		printformat(NULL, NULL, MSGLEVEL_CLIENTNOTICE, TXT_CHANSETUP_NOT_FOUND, channel, chatnet);
+	else {
+		if (g_hash_table_lookup(optlist, "bots")) g_free_and_null(rec->botmasks);
+		if (g_hash_table_lookup(optlist, "botcmd")) g_free_and_null(rec->autosendcmd);
+		if (*password != '\0') g_free_and_null(rec->password);
+		if (*password != '\0' && g_strcmp0(password, "-") != 0) rec->password = g_strdup(password);
+
```
ailin-nemui: C

src/fe-common/core/fe-channels.c
```diff
@@ -284,21 +296,60 @@ static void cmd_channel_add(const char *data)
...
 		if (*password != '\0') g_free_and_null(rec->password);
 	}
-	if (g_hash_table_lookup(optlist, "auto")) rec->autojoin = TRUE;
-	if (g_hash_table_lookup(optlist, "noauto")) rec->autojoin = FALSE;
-	if (botarg != NULL && *botarg != '\0') rec->botmasks = g_strdup(botarg);
-	if (botcmdarg != NULL && *botcmdarg != '\0') rec->autosendcmd = g_strdup(botcmdarg);
-	if (*password != '\0' && g_strcmp0(password, "-") != 0) rec->password = g_strdup(password);
 
-	signal_emit("channel add fill", 2, rec, optlist);
+	if (*password != '\0' && g_strcmp0(password, "-") != 0) rec->password = g_strdup(password);
```
ailin-nemui: C

PR #341, opened 2015-10-23
Fix invalid reads in strsplit_len when splitting on spaces

dequis:
The symptom for this one is randomly getting lines split before the last word, even if there's no need for splitting. Also, this function is only reached if recode is on, and iconv failed (for example, due to an incorrect source charset). Thanks to vague for finding this and providing valgrind logs.

The loop that looks for spaces tried to read backwards from the end of the current line, with the end being determined by len. Assuming `strsplit_len()` with `len=400`, this meant accessing `str[399]` in the first iteration. For strings that don't need splitting, this means an invalid read always.

If that invalid read happens to hit garbage that has a space character, `(len - offset)` points after the end of string, which isn't a problem for `g_strndup()` since it stops at the first null, and no splitting happens. If the garbage doesn't have any spaces, it splits by the last word.

This commit avoids that loop entirely if `(remaining_len > len)`. It also changes the way it iterates over the string to be much less confusing.

### DIFF
--- a/src/core/misc.c
+++ b/src/core/misc.c
@@ -996,11 +996,10 @@ char **strsplit_len(const char *str, int len, gboolean onspace)
 	char **ret = g_new(char *, 1);
 	int n;
 	int split_offset = 0;
-	size_t remaining_len = strlen(str);
 
 	for (n = 0; *str != '\0'; n++) {
-		split_offset = MIN(len, remaining_len);
-		if (onspace && remaining_len > len) {
+		split_offset = MIN(len, strlen(str));
+		if (onspace && strlen(str) > len) {
 			/*
 			 * Try to find a space to split on and leave
 			 * the space on the previous line.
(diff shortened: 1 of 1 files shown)


dequis:
So, uh, anyone?

ailin-nemui:
wouldn't it be sufficient to test for if onspace && strlen(str) >= len?

dequis:
Not really. `remaining_len` isn't always the same as `strlen(str)`, it's decreased at the end of the loop in the new code.

I wasn't aiming for minimal diff size here, I also moved stuff around to make it easier to understand, because that's the main reason the bug happened. I'd have to take a few minutes again to try to understand what happened here, even with the explanation I wrote.

ailin-nemui:
I didn't particularly find it easier to understand.... ;-(

dequis:
Sooo... wat do

ailin-nemui:
Either you make it understandable to me or find someone else to review it :D
Alternatively you could review the following fix
```
-		if (onspace) {
+		if (onspace && strlen(str) >= len) {

dequis:
Okay smaller diff now. I started reverting the offset-related changes but it really seemed i was making the code worse

I mean really, using an increasing counter and then str[len-1-i] isn't the sanest way to iterate backwards from the end of a string.

PR #342, opened 2015-10-27
Add sandbox support to irssi

namnamc:
This patch adds support for sandboxing (seccomp sandboxing and capability dropping) by passing the --sandbox argument. I tested it on several systems and it worked fine for me, although the lack of a test suite for irssi makes it hard to know for sure. I've been using the seccomp patch for several months with no problems at least. ahf on OFTC suggested I try to get this patch upstreamed for the next release.

The only current problems or concerns I can think of currently are:
- disabling the rdtsc instruction use to work on my system, but now irssi seems to be using it and that triggers sigsegv, so that functionality is disabled.
- capability dropping does not work properly, so it is also disabled (but it should be easy to fix, I just don't have time to debug why cap_get_proc() mysteriously fails).
- the /exec and /upgrade commands are disabled and silently fail (the file with /upgrade does not include printtext.h)

Other than that, it seems to work fine for me and for a few people who have helped me test it out.

This is my first pull request on github so forgive me if I make any novice mistakes.

### DIFF
--- a/src/core/sandbox.c
+++ b/src/core/sandbox.c
@@ -41,8 +41,8 @@
 
 #include "sandbox.h"
 
-#define IRSSI_NEEDS_RDTSC
-#define IRSSI_NEEDS_CAPS
+#define	IRSSI_NEEDS_RDTSC
+#undef	IRSSI_NEEDS_CAPS
 
 /*
  * TODO:
@@ -53,6 +53,14 @@
  *	- add support tame() on OpenBSD, and Capsicum on FreeBSD
 */
 
+struct _cap_struct {
+	struct __user_cap_header_struct head;
+	union {
+		struct __user_cap_data_struct set;
+		uint32_t flag[3];
+	} u[2];
+};
+
 void create_namespaces(void)
 {
 	/* TODO */
@@ -62,6 +70,7 @@ void drop_privileges(void)
 {
 	int tsc_state;
 	size_t i;
+	uint32_t *raw_data;
 	cap_t cap;
 	cap_flag_value_t value;
 
@@ -91,12 +100,20 @@ void drop_privileges(void)
 
 	/* drop any dangerous capabilities the process is running with
 	 * https://forums.grsecurity.net/viewtopic.php?f=7&t=2522 */
-	/* FIXME why does cap_get_proc() fail? */
 #ifndef IRSSI_NEEDS_CAPS
 	cap = cap_get_proc();
 	if (cap == NULL) {
-		fprintf(stderr, "Could not initialize capabilities (cap_get_proc failed)\n");
-		exit(1);
+		raw_data = malloc(sizeof(uint32_t) + sizeof(*cap));
+		*raw_data = 0xCA90D0; /* CAP_T_MAGIC */
+		cap = (cap_t)(raw_data + 1);
+		memset(cap, 0, sizeof(*cap));
+		cap->header.version = 0x20080522; /* capability version 3 */
+
+		if (capget(&cap->head, &cap->u[0].set) < 0) {
+			cap_free(cap);
+			fprintf(stderr, "Could not initialize capabilities (capget failed)\n");
+			exit(1);
+		}
 	}
 
 	if (cap_clear(cap) < 0) {
@@ -283,7 +300,7 @@ void enable_seccomp_sandbox(void)
 	};
 	for (i = 0; i < 2; i++) {
 		rc = seccomp_rule_add(ctx, SCMP_ACT_ALLOW, SCMP_SYS(getsockopt), 3,
-			SCMP_A0(SCMP_CMP_EQ, getsockopt_struct[i].minsockfd),
+			SCMP_A0(SCMP_CMP_GE, getsockopt_struct[i].minsockfd),
 			SCMP_A1(SCMP_CMP_EQ, getsockopt_struct[i].level),
 			SCMP_A2(SCMP_CMP_EQ, getsockopt_struct[i].optname));
 		if (rc < 0)
@@ -413,7 +430,7 @@ void enable_seccomp_sandbox(void)
 		rc = seccomp_rule_add(ctx, SCMP_ACT_ALLOW, SCMP_SYS(mmap), 4,
 			SCMP_A2(SCMP_CMP_EQ, mmap_anon_struct[i].prot),
 			SCMP_A3(SCMP_CMP_EQ, mmap_anon_struct[i].flags),
-			SCMP_A4(SCMP_CMP_EQ, 0xffffffff),
+			SCMP_A4(SCMP_CMP_EQ, (unsigned)-1),
 			SCMP_A5(SCMP_CMP_EQ, 0));
 		if (rc < 0)
 			goto fail;
@@ -428,7 +445,7 @@ void enable_seccomp_sandbox(void)
 		rc = seccomp_rule_add(ctx, SCMP_ACT_ALLOW, SCMP_SYS(mmap), 4,
 			SCMP_A2(SCMP_CMP_EQ, mmap_noanon_struct[i].prot),
 			SCMP_A3(SCMP_CMP_EQ, mmap_noanon_struct[i].flags),
-			SCMP_A4(SCMP_CMP_NE, 0xffffffff),
+			SCMP_A4(SCMP_CMP_NE, (unsigned)-1),
 			SCMP_A5(SCMP_CMP_EQ, 0));
 		if (rc < 0)
 			goto fail;
@@ -437,7 +454,7 @@ void enable_seccomp_sandbox(void)
 	rc = seccomp_rule_add(ctx, SCMP_ACT_ALLOW, SCMP_SYS(mmap), 4,
 		SCMP_A2(SCMP_CMP_EQ, PROT_READ|PROT_WRITE),
 		SCMP_A3(SCMP_CMP_EQ, MAP_PRIVATE|MAP_FIXED|MAP_DENYWRITE),
-		SCMP_A4(SCMP_CMP_NE, 0xffffffff),
+		SCMP_A4(SCMP_CMP_NE, (unsigned)-1),
 		SCMP_A5(SCMP_CMP_NE, 0)); /* NE 0, so can't be in the struct */
 	if (rc < 0)
 		goto fail;
--- a/src/core/sandbox.c
+++ b/src/core/sandbox.c
@@ -22,6 +22,7 @@
 #include <linux/futex.h>
 #include <linux/in.h>
 #include <linux/prctl.h>
+#include <linux/sched.h>
 #include <seccomp.h>
 #include <signal.h>
 #include <stdio.h>
@@ -162,10 +163,9 @@ void enable_seccomp_sandbox(void)
 	}
 
 	/* syscalls without argument filtering */
-	const int noarg_whitelist[20] = {
+	const int noarg_whitelist[19] = {
 		SCMP_SYS(brk),
 		SCMP_SYS(chdir),
-		SCMP_SYS(clone),
 		SCMP_SYS(close),
 		SCMP_SYS(exit_group),
 		SCMP_SYS(fstat),
@@ -183,7 +183,7 @@ void enable_seccomp_sandbox(void)
 		SCMP_SYS(umask),
 		SCMP_SYS(uname)
 	};
-	for (i = 0; i < 20; i++) {
+	for (i = 0; i < 19; i++) {
 		rc = seccomp_rule_add(ctx, SCMP_ACT_ALLOW, noarg_whitelist[i], 0);
 		if (rc < 0)
 			goto fail;
@@ -215,6 +215,25 @@ void enable_seccomp_sandbox(void)
 	if (rc < 0)
 		goto fail;
 #endif
+
+	/* clone */
+	rc = seccomp_rule_add(ctx, SCMP_ACT_ALLOW, SCMP_SYS(clone), 3,
+		SCMP_A0(SCMP_CMP_EQ, CLONE_CHILD_CLEARTID|CLONE_CHILD_SETTID|SIGCHLD),
+		SCMP_A1(SCMP_CMP_EQ, 0),
+		SCMP_A2(SCMP_CMP_EQ, 0));
+	if (rc < 0)
+		goto fail;
+
+	rc = seccomp_rule_add(ctx, SCMP_ACT_ALLOW, SCMP_SYS(clone), 1,
+		SCMP_A0(SCMP_CMP_EQ, CLONE_VM|CLONE_FS|CLONE_FILES|CLONE_SIGHAND|
+				     CLONE_THREAD|CLONE_SYSVSEM|CLONE_SETTLS|
+				     CLONE_PARENT_SETTID|CLONE_CHILD_CLEARTID));
+	if (rc < 0)
+		goto fail;
+
+	rc = seccomp_syscall_priority(ctx, SCMP_SYS(clone), 50);
+	if (rc < 0)
+		goto fail;
 
 	/* mkdir */
 	rc = seccomp_rule_add(ctx, SCMP_ACT_ALLOW, SCMP_SYS(mkdir), 1,
--- a/src/core/sandbox.c
+++ b/src/core/sandbox.c
@@ -193,6 +193,29 @@ void enable_seccomp_sandbox(void)
 			goto fail;
 	}
 
+#ifdef __i386__
+	/* if we are on a 32 bit architecture, we need to use socketcall */
+	const int socketcall_calls[7] = {
+		SYS_SOCKET,
+		SYS_CONNECT,
+		SYS_GETPEERNAME,
+		SYS_GETSOCKOPT,
+		SYS_SETSOCKOPT,
+		SYS_SENDTO,
+		SYS_RECVFROM
+	};
+	for (i = 0; i < 7; i++) {
+		rc = seccomp_rule_add(ctx, SCMP_ACT_ALLOW, SCMP_SYS(socketcall), 1,
+			SCMP_A0(SCMP_CMP_EQ, socketcall_calls[i]));
+		if (rc < 0)
+			goto fail;
+	}
+
+	rc = seccomp_syscall_priority(ctx, SCMP_SYS(socketcall), 50);
+	if (rc < 0)
+		goto fail;
+#endif
+
 	/* mkdir */
 	rc = seccomp_rule_add(ctx, SCMP_ACT_ALLOW, SCMP_SYS(mkdir), 1,
 		SCMP_A1(SCMP_CMP_EQ, 0700)); /* src/core/settings.c:828 */
@@ -203,6 +226,7 @@ void enable_seccomp_sandbox(void)
 	if (rc < 0)
 		goto fail;
 
+#ifndef __i386__
 	/* setsockopt */
 	struct setsockopt_struct_t {
 		int minsockfd;
@@ -250,6 +274,7 @@ void enable_seccomp_sandbox(void)
 	rc = seccomp_syscall_priority(ctx, SCMP_SYS(getsockopt), 50);
 	if (rc < 0)
 		goto fail;
+#endif
 
 	/* ioctl */
 	const int ioctl_array[3] = {
@@ -412,6 +437,7 @@ void enable_seccomp_sandbox(void)
 	if (rc < 0)
 		goto fail;
 
+#ifndef __i386__
 	/* socket */
 	struct socket_struct_t {
 		int domain;
@@ -455,6 +481,7 @@ void enable_seccomp_sandbox(void)
 	rc = seccomp_syscall_priority(ctx, SCMP_SYS(getpeername), 50);
 	if (rc < 0)
 		goto fail;
+#endif
 
 	/* futex */
 	struct futex_struct_t {
@@ -578,6 +605,7 @@ void enable_seccomp_sandbox(void)
 	if (rc < 0)
 		goto fail;
 
+#ifndef __i386__
 	/* sendto */
 	rc = seccomp_rule_add(ctx, SCMP_ACT_ALLOW, SCMP_SYS(sendto), 4,
 		SCMP_A0(SCMP_CMP_GE, 4),
@@ -604,6 +632,7 @@ void enable_seccomp_sandbox(void)
 	rc = seccomp_syscall_priority(ctx, SCMP_SYS(recvfrom), 70);
 	if (rc < 0)
 		goto fail;
+#endif
 
 	/* access */
 	const int access_mode_array[3] = {
--- a/src/core/sandbox.c
+++ b/src/core/sandbox.c
@@ -205,7 +205,7 @@ void enable_seccomp_sandbox(void)
 
 	/* setsockopt */
 	struct setsockopt_struct_t {
-		int sockfd;
+		int minsockfd;
 		int level;
 		int optname;
 		socklen_t optlen;
@@ -216,7 +216,7 @@ void enable_seccomp_sandbox(void)
 	};
 	for (i = 0; i < 2; i++) {
 		rc = seccomp_rule_add(ctx, SCMP_ACT_ALLOW, SCMP_SYS(setsockopt), 4,
-			SCMP_A0(SCMP_CMP_EQ, setsockopt_struct[i].sockfd),
+			SCMP_A0(SCMP_CMP_GE, setsockopt_struct[i].minsockfd),
 			SCMP_A1(SCMP_CMP_EQ, setsockopt_struct[i].level),
 			SCMP_A2(SCMP_CMP_EQ, setsockopt_struct[i].optname),
 			SCMP_A4(SCMP_CMP_EQ, setsockopt_struct[i].optlen));
@@ -230,7 +230,7 @@ void enable_seccomp_sandbox(void)
 
 	/* getsockopt */
 	struct getsockopt_struct_t {
-		int sockfd;
+		int minsockfd;
 		int level;
 		int optname;
 	};
@@ -240,7 +240,7 @@ void enable_seccomp_sandbox(void)
 	};
 	for (i = 0; i < 2; i++) {
 		rc = seccomp_rule_add(ctx, SCMP_ACT_ALLOW, SCMP_SYS(getsockopt), 3,
-			SCMP_A0(SCMP_CMP_EQ, getsockopt_struct[i].sockfd),
+			SCMP_A0(SCMP_CMP_EQ, getsockopt_struct[i].minsockfd),
 			SCMP_A1(SCMP_CMP_EQ, getsockopt_struct[i].level),
 			SCMP_A2(SCMP_CMP_EQ, getsockopt_struct[i].optname));
 		if (rc < 0)
@@ -330,7 +330,14 @@ void enable_seccomp_sandbox(void)
 	if (rc < 0)
 		goto fail;
 
-	/* mprotect */
+	/*
+	 * mprotect
+	 *
+	 * Although this filter prevents making non-executable pages executable
+	 * again, there is nothing stopping an attacker from using mmap to load
+	 * an arbitrary executable file with PROT_READ|PROT_EXEC in the first
+	 * place
+	 */
 	const int mprotect_prot_array[3] = {
 		PROT_NONE,
 		PROT_READ,
@@ -396,8 +403,7 @@ void enable_seccomp_sandbox(void)
 		goto fail;
 
 	/* madvise */
-	rc = seccomp_rule_add(ctx, SCMP_ACT_ALLOW, SCMP_SYS(madvise), 2,
-		SCMP_A1(SCMP_CMP_EQ, 8368128),
+	rc = seccomp_rule_add(ctx, SCMP_ACT_ALLOW, SCMP_SYS(madvise), 1,
 		SCMP_A2(SCMP_CMP_EQ, MADV_DONTNEED));
 	if (rc < 0)
 		goto fail;
@@ -641,28 +647,15 @@ void enable_seccomp_sandbox(void)
 		goto fail;
 
 	/* read */
-	struct read_struct_t {
-		int fd;
-		size_t count;
-	};
-	const struct read_struct_t read_struct[6] = {
-		{ 0, 256  },
-		{ 3, 16   },
-		{ 3, 512  },
-		{ 3, 832  },
-		{ 3, 4000 },
-		{ 3, 4096 }
-	};
-	for (i = 0; i < 6; i++) {
-		rc = seccomp_rule_add(ctx, SCMP_ACT_ALLOW, SCMP_SYS(read), 2,
-			SCMP_A0(SCMP_CMP_EQ, read_struct[i].fd),
-			SCMP_A2(SCMP_CMP_EQ, read_struct[i].count));
-		if (rc < 0)
-			goto fail;
-	}
+	rc = seccomp_rule_add(ctx, SCMP_ACT_ALLOW, SCMP_SYS(read), 2,
+		SCMP_A0(SCMP_CMP_EQ, 0),
+		SCMP_A2(SCMP_CMP_EQ, 256));
+	if (rc < 0)
+		goto fail;
 
-	rc = seccomp_rule_add(ctx, SCMP_ACT_ALLOW, SCMP_SYS(read), 1,
-		SCMP_A1(SCMP_CMP_GE, 4));
+	rc = seccomp_rule_add(ctx, SCMP_ACT_ALLOW, SCMP_SYS(read), 2,
+		SCMP_A0(SCMP_CMP_GE, 3),
+		SCMP_A2(SCMP_CMP_LE, 16384));
 	if (rc < 0)
 		goto fail;
 
@@ -748,7 +741,7 @@ void enable_seccomp_sandbox(void)
 	if (rc < 0)
 		goto fail;
 
-	rc = seccomp_syscall_priority(ctx, SCMP_SYS(poll), 200);
+	rc = seccomp_syscall_priority(ctx, SCMP_SYS(select), 200);
 	if (rc < 0)
 		goto fail;
 
--- a/src/core/sandbox.c
+++ b/src/core/sandbox.c
@@ -1,0 +1,801 @@
+/*
+ sandbox.c : irssi
+
+    Copyright (C) 2015 Namsun Ch'o
+
+    This program is free software; you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation; either version 2 of the License, or
+    (at your option) any later version.
+
+    This program is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License along
+    with this program; if not, write to the Free Software Foundation, Inc.,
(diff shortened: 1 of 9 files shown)


dequis:
>This is my first pull request on github so forgive me if I make any novice mistakes.

You should learn to use the git commandline client instead of doing everything through the github web UI. Otherwise you end up with 56 commits with meaningless titles such as "Update somefile.c".

Here, have this https://try.github.io/

namnamc:
Ok, should I close this pull request and re-upload it without all the commits? I didn't expect it to try to merge all of them and not just the final state.

dequis:
This will make a single commit with all the changes and update this pull request. Write a descriptive message in the text editor that appears in the second step.

```
git reset --soft 7ab38f2
git commit
git push --force origin master
```

namnamc:
There, fixed it. Sorry about that.

dequis:
@ahf you asked for it, so you take care of this mess.

namnamc:
Heh, is it really that bad?

dequis:
It's not *bad*, or, well, I don't know if it is. It just has "hazardous materials" symbols written all over the place.

namnamc:
Any suggestions on how to improve it? The default seccomp policy can be easily changed to return ENOSYS or any other error upon syscall violation, rather than killing the process. That would allow any accidental violations to be caught by irssi or just silently ignored.

ahf:
Thanks for submitting this. I'll get a friend of mine to have a look at the sec comp related parts and I'll help out with the irssi integration :-)

cfcs:
Hi, first of all - nice to see you working on this long overdue feature.
Part of the reason that it has not already been written (and that most other software out there also lacks sandboxing) is probably that 1) not everybody sees the value 2) it's hard to get it right.

I'm adding some comments and questions, please do not feel discouraged by this, we will beat it into shape!

First of all, the timing is unfortunate. Linux seems to be moving away from seccomp-bpf and towards using [seccomp-ebpf / eBPF filters](https://github.com/iovisor/bcc), so this will likely have to be (or at least benefit from being) extended to eBPF in a year or so.

src/core/sandbox.c
```diff
@@ -0,0 +1,801 @@
...
+#ifndef IRSSI_NEEDS_CAPS
+	cap = cap_get_proc();
+	if (cap == NULL) {
+		fprintf(stderr, "Could not initialize capabilities (cap_get_proc failed)\n");
+		exit(1);
+	}
+
+	if (cap_clear(cap) < 0) {
+		fprintf(stderr, "Could not clear capabilities (cap_clear failed)\n");
+		cap_free(cap);
```
cfcs: Why the call to `cap_free(cap)` here? Are we in a thread? If yes, why is there no `cap_free()` on line 98?

src/core/sandbox.c
```diff
@@ -0,0 +1,801 @@
...
+		if (tsc_state != PR_TSC_SIGSEGV) {
+			fprintf(stderr, "Could not disable the TSC\n");
+			exit(1);
+		}
+	}
+#endif
+
+	/* drop any dangerous capabilities the process is running with
+	 * https://forums.grsecurity.net/viewtopic.php?f=7&t=2522 */
+	/* FIXME why does cap_get_proc() fail? */
```
cfcs: Does this still cause trouble? Consider debugging this rather than merge a FIXME comment.

- examine `errno` and consult the source of libcap:
- [cap_get_proc code](https://git.kernel.org/cgit/linux/kernel/git/morgan/libcap.git/tree/libcap/cap_proc.c)
- [cap_init code](https://git.kernel.org/cgit/linux/kernel/git/morgan/libcap.git/tree/libcap/cap_alloc.c)
- [capget syscall documentation](http://linux.die.net/man/2/capget)

src/core/sandbox.c
```diff
@@ -0,0 +1,801 @@
...
+		int level;
+		int optname;
+		socklen_t optlen;
+	};
+	const struct setsockopt_struct_t setsockopt_struct[2] = {
+		{ 4, SOL_SOCKET, SO_KEEPALIVE, 4 },
+		{ 4, SOL_SOCKET, SO_REUSEADDR, 4 }
+	};
+	for (i = 0; i < 2; i++) {
+		rc = seccomp_rule_add(ctx, SCMP_ACT_ALLOW, SCMP_SYS(setsockopt), 4,
```
cfcs: Note that support for this on 32-bit is *quite* bleeding edge; see this link: 
http://www.serverphorums.com/read.php?12,1250428

On 32bit Linux socket-related system calls are dispatched through [socketcall()](http://linux.die.net/man/2/socketcall), so this code will not work as intended (if you aim to support x86 kernels compiled more than two weeks ago)

src/core/sandbox.c
```diff
@@ -0,0 +1,801 @@
...
+		if (rc < 0)
+			goto fail;
+	}
+
+	rc = seccomp_syscall_priority(ctx, SCMP_SYS(ioctl), 50);
+	if (rc < 0)
+		goto fail;
+
+	/* kill */
+	rc = seccomp_rule_add(ctx, SCMP_ACT_ERRNO(EPERM), SCMP_SYS(kill), 1,
```
cfcs: Why don't you make an array and loop over it like you did with `ioctl`?

src/core/sandbox.c
```diff
@@ -0,0 +1,801 @@
...
+		{ PROT_NONE,            MAP_PRIVATE|MAP_ANONYMOUS|MAP_NORESERVE },
+		{ PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_ANONYMOUS               },
+		{ PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_ANONYMOUS|MAP_STACK     },
+		{ PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_FIXED|MAP_ANONYMOUS     }
+	};
+	for (i = 0; i < 4; i++) {
+		rc = seccomp_rule_add(ctx, SCMP_ACT_ALLOW, SCMP_SYS(mmap), 4,
+			SCMP_A2(SCMP_CMP_EQ, mmap_anon_struct[i].prot),
+			SCMP_A3(SCMP_CMP_EQ, mmap_anon_struct[i].flags),
+			SCMP_A4(SCMP_CMP_EQ, 0xffffffff),
```
cfcs: Was this `0xffff ffff` gleaned from strace or similar? This value is `(int) (-1)`, as should probably be documented in a comment or whatever? (see `man mmap` on `MAP_ANONYMOUS`)

src/core/sandbox.c
```diff
@@ -0,0 +1,801 @@
...
+	if (rc < 0)
+		goto fail;
+
+	rc = seccomp_syscall_priority(ctx, SCMP_SYS(mmap), 60);
+	if (rc < 0)
+		goto fail;
+
+	/* madvise */
+	rc = seccomp_rule_add(ctx, SCMP_ACT_ALLOW, SCMP_SYS(madvise), 2,
+		SCMP_A1(SCMP_CMP_EQ, 8368128),
```
cfcs: `8368128`, is this a hardcoded length? comments explaining these magical numbers would be nice :-)

src/core/sandbox.c
```diff
@@ -0,0 +1,801 @@
...
+		if (rc < 0)
+			goto fail;
+	}
+
+	rc = seccomp_syscall_priority(ctx, SCMP_SYS(socket), 50);
+	if (rc < 0)
+		goto fail;
+
+	/* connect */
+	rc = seccomp_rule_add(ctx, SCMP_ACT_ALLOW, SCMP_SYS(connect), 2,
```
cfcs: "allow connect(int sockfd = 4 , ... , socklen_t addrlen = 16)", why are we doing this?

src/core/sandbox.c
```diff
@@ -0,0 +1,801 @@
...
+	if (rc < 0)
+		goto fail;
+
+	rc = seccomp_syscall_priority(ctx, SCMP_SYS(connect), 50);
+	if (rc < 0)
+		goto fail;
+
+	/* getpeername */
+	rc = seccomp_rule_add(ctx, SCMP_ACT_ALLOW, SCMP_SYS(getpeername), 1,
+		SCMP_A0(SCMP_CMP_GE, 4));
```
cfcs: again, what is the point of allowing only fd 4? it seems like something that would not be very portable, and an attacker can likely bypass it given that `close()` and `pipe()` are already whitelisted.

src/core/sandbox.c
```diff
@@ -0,0 +1,801 @@
...
+		goto fail;
+
+	/* write */
+	rc = seccomp_rule_add(ctx, SCMP_ACT_ALLOW, SCMP_SYS(write), 2,
+		SCMP_A0(SCMP_CMP_LT, 3),
+		SCMP_A2(SCMP_CMP_LE, 4096));
+	if (rc < 0)
+		goto fail;
+
+	rc = seccomp_rule_add(ctx, SCMP_ACT_ALLOW, SCMP_SYS(write), 2,
```
cfcs: My interpretation of these three rules:
- if fd < 3, and byte count <= 4096, then ALLOW
- if fd > 3, and byte count <= 4096, then ALLOW
- if fd == 3, and byte count == 8, then ALLOW

I would argue that there would probably not be much security lost by just whitelisting `write()`, what are your thoughts about that?

src/core/sandbox.c
```diff
@@ -0,0 +1,801 @@
...
+		goto fail;
+
+	/* select */
+	rc = seccomp_rule_add(ctx, SCMP_ACT_ALLOW, SCMP_SYS(select), 2,
+		SCMP_A2(SCMP_CMP_EQ, NULL),
+		SCMP_A3(SCMP_CMP_EQ, NULL));
+	if (rc < 0)
+		goto fail;
+
+	rc = seccomp_syscall_priority(ctx, SCMP_SYS(poll), 200);
```
cfcs: I think you mean `SCMP_SYS(select)`

src/core/sandbox.c
```diff
@@ -0,0 +1,801 @@
...
+	if (rc < 0)
+		goto fail;
+
+	/* apply filters */
+	rc = seccomp_load(ctx);
+	if (rc < 0)
+		goto fail;
+
+	/* seccomp has been seccessfully loaded and is now enforced */
+	return;
```
cfcs: you might want to `seccomp_release()` the filter after loading it as well, before this return call, in case the user cares about their memory load.

dequis:
Okay I think i'm starting to understand this change.

Will this need constant maintenance? Some rules are so strict that I can imagine breaking them accidentally after applying patches for unrelated stuff.

Won't this break pretty much all perl scripts? Or is this supposed to be used with perl disabled?

src/core/sandbox.c
dequis: `fd == 3` seems to be the poll() eventfd of the glib event loop, at least right now. I agree write could just be whitelisted given these rules.

src/core/sandbox.c
```diff
@@ -0,0 +1,801 @@
...
+		int fd;
+		size_t count;
+	};
+	const struct read_struct_t read_struct[6] = {
+		{ 0, 256  },
+		{ 3, 16   },
+		{ 3, 512  },
+		{ 3, 832  },
+		{ 3, 4000 },
+		{ 3, 4096 }
```
dequis: This looks like relying too much on implementation details of poll() (which is what reads fd 3 with those sizes, i assume). Same thing as with write(), could just whitelist the whole call. This is just protecting stdout and stderr from read attempts, heh.

src/core/sandbox.c
ahf: `cap` is NULL on line 98.

ahf:
@dequis It's an opt-in feature (via `--sandbox`) and yes, it will require some maintenance, but it's not something we need to focus on for every single patch we make. It will require some testing before every release and we need to ask some sandbox users to use the Git version of irssi and update regularly - I think they are willing to do that :-)

src/core/sandbox.c
```diff
@@ -0,0 +1,801 @@
...
+			SCMP_A4(SCMP_CMP_EQ, 0xffffffff),
+			SCMP_A5(SCMP_CMP_EQ, 0));
+		if (rc < 0)
+			goto fail;
+	}
+
+	const struct mmap_struct_t mmap_noanon_struct[3] = {
+		{ PROT_READ,           MAP_PRIVATE               },
+		{ PROT_READ,           MAP_SHARED                },
+		{ PROT_READ|PROT_EXEC, MAP_PRIVATE|MAP_DENYWRITE }
```
cfcs: This (`PROT_READ | PROT_EXEC`) allows loading arbitrary data as executable pages, an attractive target for ROP exploit payloads. In line 334 you try to avoid such things by limiting the use of `mprotect()`.
If you want to keep this, I would like a comment here explicitly stating that this is possible (and a comment around line 340 stating that it is still possible to make pages executable)

src/core/sandbox.c
```diff
@@ -0,0 +1,801 @@
...
+
+	rc = seccomp_rule_add(ctx, SCMP_ACT_ALLOW, SCMP_SYS(write), 2,
+		SCMP_A0(SCMP_CMP_GT, 3),
+		SCMP_A2(SCMP_CMP_LE, 4096));
+	if (rc < 0)
+		goto fail;
+
+	rc = seccomp_rule_add(ctx, SCMP_ACT_ALLOW, SCMP_SYS(write), 2,
+		SCMP_A0(SCMP_CMP_EQ, 3),
+		/* SCMP_A1(SCMP_CMP_EQ, "\1\0\0\0\0\0\0\0"), */
```
cfcs: Stuff like this will soon be possible using eBPF, but it currently isn't.
Maybe the commented out stuff should not be merged?

src/fe-text/irssi.c
```diff
@@ -74,11 +75,20 @@ void mainwindow_activity_deinit(void);
 void mainwindows_layout_init(void);
 void mainwindows_layout_deinit(void);
 
+#ifdef HAVE_SECCOMP
```
cfcs: I am not familiar with how irssi prefers to do, but it is generally considered better to have functions like this be defined no matter what, and then the function bodies themselves can be guarded. Doing so limits the amount of `#ifdef`s you need to scatter across the code.

src/core/sandbox.c
cfcs: @ahf: gotcha. but is there any reason to clean up heap variables before `exit()`?

src/core/sandbox.c
```diff
@@ -0,0 +1,801 @@
...
+	if (ctx == NULL) {
+		fprintf(stderr, "Could not initialize the sandbox (seccomp_init failed)\n");
+		exit(1);
+	}
+
+	/* syscalls without argument filtering */
+	const int noarg_whitelist[20] = {
+		SCMP_SYS(brk),
+		SCMP_SYS(chdir),
+		SCMP_SYS(clone),
```
cfcs: Consider more restrictions on this, for example blacklisting `CLONE_UNTRACED`

src/core/sandbox.c
namnamc: It's just good practice, or so I'm told, to never assume that exit() cleans up everything.

src/core/sandbox.c
namnamc: I'll get to debugging it in a few minutes. It looks like it'll be pretty simple.

src/core/sandbox.c
namnamc: Should I change it to support socketcall() as well? In the kernel source, the SYSCALL_DEFINE for socketcall() is literally just a case statement which runs things like sys_getsockopt(), sys_connect(), etc manually, so the arguments should be the same.

src/core/sandbox.c
namnamc: The ioctls may change and we may need to adjust them in the future, so putting it in an array makes it much easier to add support to new ones. The kill array comes from hardcoded signals which are not likely to change, so putting them in an array would only serve to needlessly increase the size of the code.

src/core/sandbox.c
namnamc: -1 didn't work, and I so I just put in 0xffffffff. I can add a comment explaining that any mmap() with MAP_ANON/MAP_ANONYMOUS needs the file descriptor to be -1.

src/core/sandbox.c
namnamc: That's a remnant from my personal patch which is much stricter (but likely more specific to my system). As that number seemed the same on every system I tested the patch on, I just left in in there. I can look up what the value is, and in the meantime just whitelist the MADV flag and not the value.

src/core/sandbox.c
namnamc: Is there a problem with that? The sockfd will always be 4 unless some major changes to the irssi code is made, from what I can tell.

src/core/sandbox.c
namnamc: I was contemplating removing all hardcoded references to fds for that very reason, but as it seemed to cause no problems, I just left it in.

src/core/sandbox.c
namnamc: Correct. Not sure how that slipped by.

src/core/sandbox.c
namnamc: It also limits the size that each fd can write, which can be useful for preventing things like tty pushback attacks, or for denial of service. The protecting of stdout/stderr from read attempts is just a side effect with minimal security benefit.

src/core/sandbox.c
namnamc: The sandbox is intended to prevent a compromised irssi from breaking out, not from preventing it from being ptraced by malicious binaries on the system. That's the job of things like ptrace_scope in the Yama LSM.

Also, clone() takes arguments as pointers to structs iirc, not masks of flags. Seccomp cannot whitelist something like:
clone(0x3d0f00, 0x395dc6aeff0, 0x395dc6af9d0, 0x395dc6af9d0, 0x395dc6af700)

src/core/sandbox.c
namnamc: I think keeping that commented is a good idea so we don't have to re-discover the string. Alternatively we could audit the eventfd stuff to see if it's clearly impossible for any other string to be given to poll() or not.

src/core/sandbox.c
namnamc: There are many ways to make pages executable, and the only realistic way to protect from them in irssi (as oppose to with W^X/PaX MPROTECT or gcc flags) is to limit the files which open() can access. I don't believe this can be done easily with Seccomp, but it may be possible with unprivileged namespaces. As long as open() cannot take a file and mmap() it with executable pages, then allowing executable pages is not a problem. Things like ROP will always be a risk (at least until RAP becomes popular), but even with ROP, you cannot bypass seccomp protections.

However I agree with you, and I'd like to find some way to prevent adding arbitrary data as executable pages, but I can't think of any simple ways to do it with seccomp, other than waiting until all executable libraries are loaded before restricting mmap() which might be impractical. If there is no way to do this, then I guess I'll just add a comment mentioning it.

src/fe-text/irssi.c
namnamc: So the only ifdef should be this one?

if (sandbox)
#ifdef HAVE_SECCOMP
        ... do sandbox stuff
#else 
        ... tell the user sandboxing is not supported
#endif 
}

src/core/sandbox.c
cfcs: On my system it looks like this (`man 2 clone`):
```c
       long clone(unsigned long flags, void *child_stack,
                 void *ptid, void *ctid,
                 struct pt_regs *regs);
```

seccomp-bpf is perfectly capable of making comparisons on arguments `0x003d0f00`, whether it makes sense or not depends on the context. In this case you would do a logical NOT on `CLONE_UNTRACED`.

The purpose of **blacklisting** (disallowing) `CLONE_UNTRACED` would be the opposite of protecting against malicious processes ptracing the cloned thread, namely to ensure that you can follow new threads with ptrace when debugging irssi. Documentation for the flag can be found by typing `man 2 clone` and typing `/CLONE_UNTRACED` (the `/` is used to search in `less` / `more` / your pager):
```
       CLONE_UNTRACED (since Linux 2.5.46)
              If  CLONE_UNTRACED  is  specified, then a tracing process cannot
              force CLONE_PTRACE on this child process.
```

src/core/sandbox.c
cfcs: If we are to keep the comment, the comment should have some text attached that explains why it's there, and what the purpose would be in a future change.

src/fe-text/irssi.c
cfcs: I would get rid of the ifdef here, and instead have it inside the function bodies in `sandbox.c`, like this:
```void drop_privileges(void)
{
#ifdef HAVE_SECCOMP
...
```

Does that makes sense?

src/core/sandbox.c
cfcs: Well, we have a good opportunity to implement a sort of "poor man's PaX MPROTECT" here since we're doing syscall whitelisting. You already took care of `mprotect()` itself, which I think is great.

`mmap()` (see `man 2 mmap`) operates on file descriptors, not filename strings, so we do not have to deal with dereferencing userspace pointers (which seccomp-bpf is not capable of).
Namespaces will not solve the problem as we are whitelisting `pipe()`, so an attacker can simply feed the desired code to mmap using an unnamed pipe.

We can make it harder to exploit flaws by forcing attackers to have long ROP chains (with a good probability that they will be capped due to IRC's line limit of 512 bytes). This is speculation, of course, but I could imagine a variety of situations where this line limit could turn out to be an advantage.

I believe that waiting for library initialization is the right way to go (or having several seccomp stages where the policy is gradually tightened). What are your thoughts about that?

src/core/sandbox.c
cfcs: This seems a bit too system-specific for my taste; it depends on the buffering algorithm used to read/write, amongst other things.

Since `close()` is whitelisted (we could limit it to file descriptors `>4` perhaps?), we cannot rely on these integers to carry any significance.

If there is no solid reasoning for these numbers, would you consider changing it to some sort of "sane upper bound" instead? (ie ALLOW if it's `<=16384`)

src/core/sandbox.c
cfcs: Yes, but we can't rely on `fd 3` being the eventfd.
See my comment below: https://github.com/irssi/irssi/pull/342#discussion_r43468637

src/core/sandbox.c
namnamc: Unfortunately, libseccomp does not allow for gradual tightening of the policy. As soon as the policy is enabled with seccomp_load(), it calls PR_SET_NO_NEW_PRIVS, although you can override that if you have CAP_SYS_ADMIN. This means that once a policy is enforced, it cannot be modified at all, including modified to be made more tight.

Waiting for library initialization would not be bad, although it does mean that seccomp would not work with plugins unless they are loaded right away. Perhaps waiting to for irssi to connect to the network before enabling seccomp? I can't think of all that many threat models where it would be important to protect irssi before it connects to the network.

src/core/sandbox.c
cfcs: I think that would seem reasonable (removing the hardcoded fd checks).
This is highly dependent on `libc` version, and future library updates may bump default read/write lengths, some platforms may have different default fds, etc.
If there is no obvious security benefit of having these hardcoded fds, I really think they should be removed, if you can't test all combinations of supported architectures, libc, etc.

src/core/sandbox.c
namnamc: Seccomp-bpf may be capable of it, but is libseccomp? I do not feel like I am able to write BPF filters directly. It would be very sloppy and would require high maintenance.

src/core/sandbox.c
cfcs: For example, irssi may be started with an open inherited file descriptor for use with some plugin or whatever, causing the sockfd to be 5.

src/fe-text/irssi.c
namnamc: That would require far more ifdef statements, and they would be scattered all over sandbox.c, rather than put in one place in irssi.c. IMO, I think that putting the ifdef in irssi.c makes much more sense and you can instantly tell what its purpose is.

src/core/sandbox.c
cfcs: That'd be great :)

src/core/sandbox.c
namnamc: Alright. A stupid question, but what is the proper way to update the code with git to add it? Would I save it for later and make a new pull request when everything's ready, like a v2 patch on a mailing list?

src/core/sandbox.c
namnamc: I actually used torsocks for most of the stracing for that very reason, and I couldn't find any case where torsocks changes what the file descriptors are. All torsocks does is wrap the connect() and socket() calls to go through a socks5 proxy. But I guess there are probably others which do... Would it be better to have a rule that allows it if it is >= 4, or should it be whitelisted entirely?

src/core/sandbox.c
namnamc: The descriptor limit would have to be >= 4 rather than > 4, because it is used for several temporary file descriptors iirc, unless we enable seccomp later after everything else is initialized.

Yeah, I can set the size limit to a sane upper bound.

src/core/sandbox.c
cfcs: See `/usr/include/seccomp.h`:
```
typedef uint64_t scmp_datum_t;
// ...
struct scmp_arg_cmp {
        unsigned int arg;       /**< argument number, starting at 0 */
        enum scmp_compare op;   /**< the comparison op, e.g. SCMP_CMP_* */
        scmp_datum_t datum_a;
        scmp_datum_t datum_b;
};
// ...
#define SCMP_CMP(...)           ((struct scmp_arg_cmp){__VA_ARGS__})
// ...
#define SCMP_A4(...)            SCMP_CMP(4, __VA_ARGS__)
```
The problem is that seccomp casts operands to unsigned 64bit longs while `mmap()` operates on `int fd`. Things like these are exactly what makes a hardcoded `0xffffffff` value problematic; an unsigned 64bit int `-1` is silently cast to `0xffffffffffffffff`.

Example:
```
$ cat lol.c
#include <stdio.h>
#include <stdint.h>
void main(){
  printf("%Lx\n", (uint64_t) (-1), 0);
  printf("%Lx\n", (uint64_t) ((unsigned int)-1), 0);
}
$ gcc lol.c -o a && ./a
ffffffffffffffff
ffffffff
```

src/core/sandbox.c
cfcs: Ok, no objections to that.

src/core/sandbox.c
dequis: Just edit the code, `git add file`, `git commit`, `git push` and it will be updated here.

src/core/sandbox.c
cfcs: The problem is that `socketcall()` takes a struct pointer, and we can't dereference that from seccomp-bpf (`man 2 socketcall`):
```
       int socketcall(int call, unsigned long *args);
```

The thread above aims to fix that, but since that doesn't exist yet, it means we can't filter these syscalls on 32bit yet. Since these system calls do not exist on 32bit x86, it means you're going to have to make these things depend on the target architecture at compile-time, and that you can't filter arguments to these functions on x86.

src/core/sandbox.c
namnamc: This is a simple implementation of cap_get_proc() which I can confirm works. I think it's simple enough that it'd be very low maintenance. It basically uses capget() directly.

This is what cap_get_proc() can be replaced with. It works fine for me, and if for some reason something goes wrong with it, it will be caught down the line by the for loop.
```c
uint32_t *raw_data;
cap_t cap;
struct _cap_struct {
        struct __user_cap_header_struct head;
        union {
                struct __user_cap_data_struct set;
                uint32_t flat[3];
        } u[2];
};

raw_data = malloc(sizeof(uint32_t) + sizeof(*cap));
*raw_data = 0xCA90D0; /* CAP_T_MAGIC */
cap = (cap_t)(raw_data + 1);
memset(cap, 0, sizeof(*cap));
cap->head.version = 0x20080522; /* capability version 3 */

if (capget(&cap->head, &cap->u[0].set) < 0) {
        cap_free(cap);
        cap = NULL;
}
```

Though I still haven't found out what the exact problem is (for some reason trying printf debugging with a patched libcap wasn't working), I could replicate it on Debian stable, and I was calling a function which should not fail under normal circumstances (i.e. I didn't use cap_get_proc() incorrectly). It might be that it simply won't work out of the box in irssi.

src/core/sandbox.c
namnamc: Libseccomp supports filtering syscalls for only specific architectures, so we can have it whitelist all of `socketcall()` on x86, and be tighter on x86_64. This seems like yet another one of those times when x86 just has to suffer with less security. Honestly the only thing I would worry about is `getsockopt()` and `setsockopt()` because of their history of insecurity.

src/core/sandbox.c
namnamc: Then I agree. For these at least, hardcoded fds seem like a pain. Perhaps if we have seccomp wait until irssi connects to the network, we could whitelist the fds that are actually in use though, something like `enable_seccomp_sandbox(network_fd, some_other_fd, yet_another_fd)`.

src/fe-text/irssi.c
cfcs: Yes, but it would be contained to the relevant file (`sandbox.c`).

Anyway, this is a question of policy related to irssi, and since I'm not an irssi developer, you should ignore me for now and wait for input from @dequis or @ahf to hear their opinion. :-)

src/core/sandbox.c
cfcs: Oh god no, *nobody* is capable of writing readable and maintainable filters using BPF opcode macros. eBPF will have compilation of filters from LLVM languages, so this stuff will all be much easier in the future, do not feel alone!

Sorry, I should have included an example, this is not very straightforward with libseccomp (which is also quite hard to use and quite poorly documented imho).
The documentation for the functionality we need in this case is stowed away in `man seccomp_rule_add`:
```
       SCMP_CMP_MASKED_EQ
              Matches  when  the  masked argument value is equal to the masked
              datum value, example:
              SCMP_CMP( arg , SCMP_CMP_MASKED_EQ , mask , datum )
```

So in this case we want to emulate "MASKED_NOT_EQ":
```c
#include <linux/sched.h>
SCMP_A0(SCMP_CMP_MASKED_EQ, CLONE_UNTRACED, 0)
// equal to: (flags & CLONE_UNTRACED) == 0
```

src/core/sandbox.c
cfcs: Hm, is it PR_SET_NO_NEW_PRIVS that is the problem?
Anyway, we are whitelisting, so we can't really restrict things once we have whitelisted them.

I think that seems like a decent trade-off, but it could cause problems if an attacker can write something to disk that will be executed before the seccomp filter is loaded on next start of irssi, bypassing the sandbox. Hmm. I assume some of the irssi devs will have a stronger opinion on this than I have, so let's hear them out.

src/fe-text/irssi.c
namnamc: @ahf was actually the one to tell me to do it that way. Previously I had done it something like like so:
```c
#ifdef HAVE_SECCOMP
if (/*stuff that checks if --sandbox is passed on the command line*/)
    int sandbox = 1;
#else
const int sandbox = 0;
#endif
```

But was told instead to change it to be in the `if (sandbox)` statement in `irssi.c`, so the `--sandbox` argument would always be available, but would display an error if used on a build without sandbox support.

src/core/sandbox.c
namnamc: Ah, and then set `SCMP_ACT_KILL` if the `SCMP_CMP_MASKED_EQ` is true?

src/core/sandbox.c
cfcs: Alright, I'm still not convinced that hardcoding fds is an entirely good idea (because of torsocks etc), but that's definitely better. We would have to be careful that 0-3 do not have `FD_CLOEXEC` or similar set so they would be closed on `clone()` for this to make sense.

src/core/sandbox.c
cfcs: Yep, still a lot better than nothing, so if that's the best we can do, that's what we'll have to do.

src/core/sandbox.c
namnamc: The same thing came up in a thread on the qemu-devel mailing list for a different seccomp patch I was writing. They ended up deciding to enable the sandbox later, even though that means a maliciously crafted qcow2 file could exploit it. But in exchange, they were able to drop `chroot()` and `setuid()` syscalls from the filter.

Personally, I think a policy which is enabled later, but is more strict is better. I mean, anyone could just put a malicious script and set it to autoload if they have write access to `~/.irssi`. They would have no need to maliciously craft some config file to execute arbitrary code, when irssi has all the hooks they need. As a result, I don't think it is important (or even possible) to modify irssi to protect itself from code execution against an actor who has write access to `~/.irssi`.

src/core/sandbox.c
namnamc: I made a commit which adds `socketcall()` support if `__i386__` is defined. Unfortunately the network calls are all over the place. Should I re-order it into chunks, e.g. for networking, file operations, and misc? That would allow the entire networking chunk to be in the same `#ifdef` statement.

src/core/sandbox.c
cfcs: No, this is for allow (this code above verifies that the `CLONE_UNTRACED` flag is NOT set).

To check if a flag IS set, you would do `SCMP_A0(SCMP_CMP_MASKED_EQ, MY_FLAG, MY_FLAG)`

src/core/sandbox.c
cfcs: Nice work!

Ultimately that would be nice, but you can wait with that change till you're satisfied with the rest of the patchset so as not to disturb the line numbers / comments too much.

src/core/sandbox.c
namnamc: You're right, I misread. I'll tighten up the filter for `clone()` then.

namnamc: For now I just added an argument whitelist to `clone()`.

namnamc:
I just made a commit fixing capability support (more or less), changing the `0xffffffff` to `(unsigned)-1`, and fixing a typo where I still had a `SCMP_CMP_EQ, fd` where I should have had `SCMP_CMP_GE, fd`. I'm not sure what else there is to be done.

Btw, why does the Travis build fail? It compiles fine on my system. Does Travis use -Werror or something?

dequis:
You can follow the "details" link, then click in one of the failed builds to get the output log.

```
sandbox.c:26:10: fatal error: 'seccomp.h' file not found
#include <seccomp.h>
         ^
1 error generated.
```

namnamc:
Clicking the failed builds didn't display anything. Might be my browser's fault. Good to know it's not a problem with the code at least.

ailin-nemui:
please ensure that compilation without sandbox remains working

namnamc:
Compilation works fine without the sandbox.

src/core/sandbox.c
namnamc: So turns out I was wrong. It is possible to have a multi-layered seccomp sandbox, but it's generally not recommended to leave it in the first stage for very long. As long as `prctl()` (or `seccomp()`) is whitelisted in the first stage, then the second stage can come in to further tighten it. PR_SET_NO_NEW_PRIVS is only a problem for revoking previous filters, not for adding onto them.

namnamc:
I decided to re-write a lot of the sandbox code from scratch and maybe just make a new pull request when it's all just about ready, instead of doing it gradually here. I think that'll allow me to get a better (and neater) policy with less clutter.

cfcs:
Please link your new branch here, and bump when ever you want reviews.

ailin-nemui:
as per comment I will close this, feel free to resubmit

PR #346, opened 2015-11-02
More conventional cut buffer handling.

toddpratt:
Make delete_previous_word and delete_to_previous_space prepend to the cut buffer on contiguous keypresses that trigger them.

### DIFF
--- a/src/fe-text/gui-entry.c
+++ b/src/fe-text/gui-entry.c
@@ -35,21 +35,21 @@ static unichar i_toupper(unichar c)
 {
 	if (term_type == TERM_TYPE_UTF8)
 		return g_unichar_toupper(c);
-	return c <= 255 ? toupper(c) : c;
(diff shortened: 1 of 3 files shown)


dequis:
Error from the clang builds:

```
gui-entry.c:38:12: error: comparison of unsigned expression >= 0 is always true
      [-Werror,-Wtautological-compare]
        return (c >= 0 && c <= 255) ? toupper(c) : c;
                ~ ^  ~

gui-entry.c:45:12: error: comparison of unsigned expression >= 0 is always true
      [-Werror,-Wtautological-compare]
        return (c >= 0 && c <= 255) ? tolower(c) : c;
                ~ ^  ~

gui-entry.c:52:12: error: comparison of unsigned expression >= 0 is always true
      [-Werror,-Wtautological-compare]
        return (c >= 0 && c <= 255) ? isalnum(c) : 0;
                ~ ^  ~
```

ailin-nemui:
:+1:

PR #347, opened 2015-11-03
Avoiding to compile portions of statements and expressions with preprocessor conditionals.

flaviommedeiros:
Removing conditional directives that split up parts of statements, which might influence code understanding, maintainability and error-proneness negatively, as suggested by code style guidelines from the Linux Kernel, practitioners and researchers.

- https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/tree/Documentation/CodingStyle#n892
- https://www.cqse.eu/en/blog/living-in-the-ifdef-hell/

More than 80% of developers prefer to avoid this kind of preprocessor conditionals according to an online survey with 202 developers from 24 C open-source projects.

- http://www.cs.cmu.edu/~ckaestne/pdf/ecoop15.pdf

### DIFF
--- a/src/core/misc.c
+++ b/src/core/misc.c
@@ -431,11 +431,11 @@ int mkpath(const char *path, int mode)
 		dir = g_strndup(path, (int) (p-path));
 		if (stat(dir, &statbuf) != 0) {
 #ifndef WIN32
-			if (mkdir(dir, mode) == -1)
+            gboolean failed = (mkdir(dir, mode) == -1);
 #else
-			if (_mkdir(dir) == -1)
+            gboolean failed = (_mkdir(dir) == -1);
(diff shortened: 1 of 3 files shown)


ailin-nemui:
indentation is messy

dequis:
I'd rather just remove the win32 ifdefs. I don't think we support that directly at all.

LemonBoy:
There's `g_mkdir_with_parents` which was introduced in GLib 2.8, since we've finally raised the minimum version number I guess it's a good time to catch two birds with a stone.

ailin-nemui:
@flaviommedeiros please address the raised issues

dequis:
I'm just going to close this in favor of #376

PR #350, opened 2015-11-09
Strip trailing whitespace from commands

LemonBoy:
We probably need to do that for some other commands too...

### DIFF
--- a/src/core/commands.c
+++ b/src/core/commands.c
@@ -748,6 +748,11 @@ int cmd_get_params(const char *data, gpointer *free_me, int count, ...)
 			if (cnt == 0 && count & PARAM_FLAG_GETREST) {
 				/* get rest */
 				arg = datad;
+
+				/* strip the trailing whitespace */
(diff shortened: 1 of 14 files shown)


src/core/commands.c
```diff
@@ -748,6 +748,11 @@ int cmd_get_params(const char *data, gpointer *free_me, int count, ...)
 			if (cnt == 0 && count & PARAM_FLAG_GETREST) {
 				/* get rest */
 				arg = datad;
+
+				/* strip the trailing whitespace */
+				if (count & PARAM_FLAG_STRIP_TRAILING_WS) {
+					arg = g_strchomp (arg);
```
dequis: You should strip the trailing whitespace after the function name

LemonBoy:
Updated to cover some more cases, review dis kthxbai

ailin-nemui:
seems useful to me, any other opinions?

dequis:
Yeah it seems good. The approach is very conservative so the only potential problem i can imagine is having to add this flag to other commands not considered here, and that is a good thing.

PR #352, opened 2015-11-11
silent make perl

### LINKED ISSUE #321: Make Perl compile nicely with silent rules enabled
After a66bb95d0e3d76dfb6abda88d6f128540279b461 landed we are now compiling with silent rules enabled, which will hopefully mean that people will introduce less warnings in the code.

Our Perl build-system for `src/perl/` is still doing a lot of stuff itself outside of Automake. It would be nice if someone would be up for looking into making the output less verbose when not build with `make V=1`.

See the bottom of https://www.gnu.org/software/automake/manual/html_node/Automake-Silent-Rules.html for more information.

ailin-nemui:
Implement "silent" Makefiles by post-processing the EU::MM generated
Makefile. Fixes #321

### DIFF
--- a/src/perl/Makefile.am
+++ b/src/perl/Makefile.am
@@ -117,16 +117,26 @@ textui_sources = \
 EXTRA_DIST = \
 	get-signals.pl \
 	irssi-core.pl \
+	Makefile_silent.pm \
 	$(common_sources) \
 	$(irc_sources) \
 	$(ui_sources) \
 	$(textui_sources)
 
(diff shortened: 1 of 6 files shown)


ailin-nemui:
I could not find any reference to autotools/kernel style "silent" rules for Perl, so here's one attempt to create this. However in its current form it lacks integration with the automake-generated Makefiles in the rest of the code.

I'd prefer if someone else can continue this work as I suspect the rest will be mostly automake/Makefile centric ;-))

ailin-nemui:
@ahf maybe you want to test this?

Issue #99, opened 2014-07-16
"/join channel" works when you have not joined #channel yet. If you have, it stops working.


RichiH:
As requested by my all-time favourite ahf:

12:31:29 < ahf> we have moved to github!
12:31:34 < ahf> just file an issue with a link to bugs.i.o!
12:31:44 < ahf> haven't you read _the_ announcement RichiH
12:31:54 < ahf> I came out as a pokemon furry

http://bugs.irssi.org/index.php?do=details&task_id=581
http://bugs.irssi.org/index.php?do=details&task_id=758

ailin-nemui:
white-space should be trimmed for the call to `channel_find` in `cmd_join`. Patch welcome

flaw600:
I'm confused about this issue. The command goes haywire if there's whitespace?

ailin-nemui:
the special case that makes `/join #channel` do the same as `/window goto #channel` does not work with trailing white space

RichiH:
Yes. And auto completion introduces whitespace when tabbing.

Richard

Sent by mobile; excuse my brevity.

KindOne-:
I use [trigger.pl](https://github.com/irssi/scripts.irssi.org/blob/gh-pages/scripts/trigger.pl) with ```-send_command -regexp '\s+$' -replace ''``` to strip all the trailing whitespaces from commands. #274

RichiH:
Neat, I will steal that.

Does not fix the non-hash join, though.

Richard

Sent by mobile; excuse my brevity.

ahf:
I've sold my soul to @RichiH for getting him to fix things for me elsewhere.

Let this be a note to everyone that even in the open source community we must be able to do under the table negotiations! :-D

RichiH:
Less talking, more fixing.

RichiH:
Less idling, more fixing.

dequis:
I don't know what's going on here but I like this guy.

flaw600:
As do I.

RichiH:
I am myself.

@ahf is the lazy.

RichiH:
That's all? That's what @ahf has been wailing about for _almost ten years_ that it's hard to fix, that he needs proper bug reports in various trackers and body cavities, plus warm milk and a blankie?

Son, I am disappoint.

That being said, the commit does not look as if it fixes the first part of the bug report, merely the latter.

dequis:
[...]

PR #357, opened 2015-11-19
Correctly alias 'channel' to '#channel'

LemonBoy:
Solves the other half of #99 (#49.5 ?)

### DIFF
--- a/src/irc/core/irc-channels.c
+++ b/src/irc/core/irc-channels.c
@@ -174,6 +174,13 @@ static CHANNEL_REC *irc_channel_find_server(SERVER_REC *server,
 					    const char *channel)
 {
 	GSList *tmp;
+	char *fmt_channel;
+
+	/* if 'channel' has no leading # this lookup is going to fail, add a
+	 * octothorpe in front of it to handle this case. */
+	fmt_channel = server_ischannel(SERVER(server), channel) ?
+	    g_strdup(channel) :
+	    g_strdup_printf("#%s", channel);
 
 	for (tmp = server->channels; tmp != NULL; tmp = tmp->next) {
 		CHANNEL_REC *rec = tmp->data;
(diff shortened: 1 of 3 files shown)


dequis:
1. God dammit stop fixing what ahf should fix
2. This will change **all** calls to channel_find (including `Irssi::channel_find()` and `$server->channel_find()` from perl) to succeed where they could have failed before.

After inspecting the results of grepping `channel_find` i can't find any obvious consequences of this. Would need something like a command that takes both nicks and channels in the same parameter, and uses the result of `channel_find()` to determine if it's a channel instead of calling `server_ischannel()` before.

At least from the C side, what i've seen so far seems to behave sanely:

```c
	if (server_ischannel(SERVER(server), target))
		item = irc_channel_find(server, target);
	else
		item = irc_query_find(server, target);
```

But this piece of code doesn't even handle user input, it handles *server* input. So probably not even relevant.

The perl side is messier. Some scripts don't even care about checking the null. Many are handlers of channel events. Very few call `$server->ischannel`. Most seem to fail or return if it's null, i haven't seen a single `} else {` that falls back to handling the channel parameter as a nick.

I guess this is probably safe.

RichiH:
@dequis to be fair to @ahf, he is epically useless.

ailin-nemui:
I think we should keep the addition of # IRC-specific

LemonBoy:
@ailin-nemui This is currently being done in `irc_channel_find_server` so it's pretty much limited to IRC only as-is.

WRT script breakage, there was no guarantee that `channel_find` could be used to successfully determine if a given name is a channel or not, I'd just shrug it off and blame the script developers for trying to break the rules.

ailin-nemui:
the code seems fine, I'm still not convinced if this is the best way to do it, would sort of prefer to make it just apply to /join (but that is a generic command so it wouldn't be as straight forward as changing channel_find). Not my call ;-)

dequis:
I think it's okay.

LemonBoy:
Hmm, we're aliasing <name> as #<name> thus potentially shadowing a valid <name>. Can't check atm if `channel_find` is used in such situations but I'd go the safe route and add the check in `cmd_join` (and remove it from the actual join-a-channel code-path).

dequis:
```
19:11 < TheLemonMan> Nei, ahf gave his blessing to #357, let's merge ?
19:15 < dx> 'blessing'
19:15 < dx> he gave a not very enthusiastic 'okay'
19:17 < TheLemonMan> using a meme
19:17 < dx> a stale meme, even
19:17 < dx> shit tier blessing imo
19:19 < TheLemonMan> in jpeg format, full of compression artifacts
19:20 < Nei> sure
19:21 < Xena> Needs more jpg
19:22 < TheLemonMan> print it, scan it, send it as a huge ass uncompressed tiff
```

RichiH:
`shit tier`? I wouldn't expect anything less, or more, from @ahf 

Thanks to everyone else :)

PR #360, opened 2015-11-22
Save a patch from the Flyspray oblivion, plus a nice unrelated commit to make everyone (un)happy

LemonBoy:
http://bugs.irssi.org/index.php?do=details&task_id=904

### DIFF
--- a/src/fe-text/gui-readline.c
+++ b/src/fe-text/gui-readline.c
@@ -177,7 +177,7 @@ static void paste_buffer_join_lines(GArray *buf)
 	if (buf->len == 0)
 		return;
 
-	arr = (unichar *) paste_buffer->data;
+	arr = (unichar *)buf->data;
 
 	/* first line */
 	if (IS_WHITE(arr[0]))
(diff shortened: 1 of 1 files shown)


dequis:
I don't think that's what backport means.

Also nice unrelated commit.

dequis:
:heart:

That patch is really useless in practice, but, whatever. It's certainly more correct.

LGTM

dequis:
Actually, can you use `i_isblank()` instead?

From `common.h`:

```c
/* ctype.h isn't safe with chars, use our own instead */
...
#define i_isblank(x) isblank((int) (unsigned char) (x))
```

ailin-nemui:
maybe you shouldn't mix different commits like this

LemonBoy:
Usingb`i_isblank` on a `unichar` would bring an unwanted downcast, introducing a bug we'll probably hear people complain about in ~10 years.

The commits are related, both target the same function and don't introduce any functional change.

dequis:
Lemon is right, so i'm merging this.

PR #361, opened 2015-11-22
Use the expanded filename when picking the awaylog

LemonBoy:
This fixes a long-standing bug where 'fname' was being feed to cat
instead of 'real_fname', causing it to quit with a 'No such file or directory'
error.

[FS#377](http://bugs.irssi.org/index.php?do=details&task_id=377)


dequis:
Yeah, feeding real_fname to your cat is important for its health.

LGTM

ailin-nemui:
noone uses away log with %Y...

PR #363, opened 2015-11-23
module check irssi version (date/time based way)

### LINKED ISSUE #324: ABI versioning and compatibility check from modules
0.8.18 breaks abi and users might get segfaults if they don't rebuild / reinstall the perl stuff or external modules.

Relevant PR #289

>I'd rather introduce some version checks in the ABI to catch these fuckups early. We should be allowed to change our ABI at will and will need to do so in many cases where it cannot be avoided

ailin-nemui:
Add explicit checks into every module to match the Irssi version date.

One way to fix #324


dequis:
Not a huge fan of the `g_strdup_printf("%d.%d", ...)`, and having to deal with heap-allocated strings for this kind of stuff.

How about returning a guint64 with `IRSSI_VERSION_DATE * 10000 + IRSSI_VERSION_TIME`?

Or just `IRSSI_VERSION_DATE`.

Or something like `#define ABI_VERSION 1` in common.h that is manually increased whenever we feel the need to force rebuilds of third party plugins - we don't break abi every day.

ailin-nemui:
we don't have any existing guint64's. Manual ABI versioning also came to my mind, the basic code would still be the same

ailin-nemui:
can we agree on how to do it?

dequis:
I prefer the manual method.

*pokes @ahf*

PR #368, opened 2015-11-25
Random fixes

incertia:
Made the suggested build script in `INSTALL` less confusing (`./configure` does not exist in fresh clones) and ignore `MYMETA.*` in the home directory to be consistent with ignoring `MYMETA.*` in subdirectories.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,6 +24,7 @@ ltversion.m4
 lt~obsolete.m4
 pm_to_blib
 stamp-h1
+MYMETA.*
 
 docs/help/Makefile.am
 docs/help/[a-z]*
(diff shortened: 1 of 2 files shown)


vague666:
I would mention that autogen.sh has to be run only when sources are downloaded from git, the source tarball does have the configure script.

incertia:
Thanks for pointing out that the source tarball has the configure script. A note was added advising that users should run `./autogen.sh` if and only if `./configure` does not exist.

GeertHauwaerts:
:heart:  your avatar ;)

incertia:
<3

ailin-nemui:
The INSTALL document used to be targeted at releases. Release tarball users should not run ./autogen.sh. You can find more information on http://irssi-import.github.io/download/#sources under *Development sources*

PR #372, opened 2015-12-06
Handle ERR_CANTJOINOPERSONLY numeric.

LemonBoy:
FS#759

This needs as much testing as I need sleep.


ailin-nemui:
hesitant about adding support for non-standard numerics. looking forward to hear whether it works though

ailin-nemui:
we can't integrate this as it conflicts with G/Kline MASK TOO WIDE error on ircu networks

dequis:
Tested with unreal, the format is this:

```
<< JOIN #opers
>> :irc2.unrealircd.org 520 dx|ebooks :Cannot join channel #opers (IRCops only)
```

Note the lack of 

It displays like this:

```
-!- Cannot join channel #opers (IRCops only)
```

This PR would probably show it like this 

```
Cannot join to channel Cannot join channel #opers (IRCops only) (You must be an operator)
```

It also won't fix the failure to join the channel, that's handled by a completely different code path that expects the numeric parameters to be `<channel> :<reason>` and this one is just `:<reason>`

I vote for not caring.

PR #377, opened 2015-12-09
module check irssi version (manual way)

ailin-nemui:
Add explicit checks into every module to match the ABI version defined
in common.h

Alternative to #363


dequis:
I like this better.

```
expandos.c:421:32: warning: format specifies type 'unsigned long' but the
      argument has type 'int' [-Wformat]
        return g_strdup_printf("%lu", IRSSI_ABI_VERSION);
                                ~~~   ^~~~~~~~~~~~~~~~~
                                %d
```

Don't really see the point of using unsigned long, int is fine.

dequis:
@ahf

ailin-nemui:
@jillest ahf would like to know what you think

jillest:
How do developers know they need to bump the ABI version?

ailin-nemui:
@jillest any other suggestions?
manual version check: devs need to manually keep track and remember to bump version
version by build date: may require needless module recompilation and possibly obstruct binary module providers

dequis:
I'd say the main case would be changing structs in any places other than the end. We catch that in the review process.

IMO bumping the ABI version during development versions doesn't need to be 100% accurate, it doesn't hurt to do it more often than needed and doesn't hurt too much if we accidentally miss one (after all, git users can expect some breakage and a lot of rebuilding). But it's important to at least bump it once between releases when something ABI-related changed. I think that by setting the bar that low it's hard to mess up.

ailin-nemui:
btw your description is not enough, for example if the module allocates memory then changes at the end need to be accounted for so that is also a change. and if you fill in some new functions you need to NULL them and make them conditional. similarly if we change the signalling layer (that is actually an invisible API change)

ahf:
We cannot let the ABI be a numerical representation of our version? Just require people to (re)build their modules against the given release they want to run?

ailin-nemui:
isn't that what I did in #363?

dequis:
If it's just the release version, it won't help with abi breakages until the release happens.

If it's just the date/time like the other PR, it will require too many rebuilds of other modules, which might not be a bad thing.

(If we go with the other PR I'd rather have a guint64 instead of passing heap allocated strings around. Or just IRSSI_VERSION_DATE to have day-level granularity)

Either way, anything is better than nothing.

PR #382, opened 2015-12-13
ensure staticlib perl works on 5.22

ailin-nemui:
may nevertheless be good to get some better understanding of mark, SP (and thus ax) in the boot process. What we figured out so far is that `items` becomes -1 in the second boot process if we do not pass the required parameters (which sounds like we may be messing with the Perl stack by making some assumptions) (Could also be some part of Perl itself doing it since we boot without the required arguments)


dequis:
This is ready, right?

ailin-nemui:
yes since we came to the conclusion that this is a correct fix

PR #334, opened 2015-10-08
Command history reset

vague666:

PR #385, opened 2015-12-15
option to clear the history

ailin-nemui:
finishes #334


ailin-nemui:
you can give each window which you want to have a separate history not shared with the global history, a custom history name (commonly you would equal the history name with the window or main item name)

If you use the same history name in multiple windows, all those windows would be sharing one history. It's the UI to setting the window->history pointer.

Issue #78, opened 2014-07-02
ignore overwrites


swiniopas:
I don't know if I'm missing something here (new user), but after ignoring, say,  messages that contain 'lol' only like this: /ignore -regexp -pattern "^[lL][oO][lL]$" * PUBLIC, when trying to /ignore * JOINS PARTS QUITS, the two entries get merged without the regexp pattern and I end up ignoring all public messages. If I add the ignore rules in opposite order or edit config manually it seems to work fine.

ailin-nemui:
@swiniopas would you be interested in testing that patch?

ailin-nemui:
please reopen if this is still an issue in 0.8.18-beta2 or later

PR #390, opened 2016-01-03
Clean up the ignore_find API to make it more powerful.

LemonBoy:
I'm quite unsatisfied with the enum naming (and the alignment of some code, but might just be my need to please the eye).

Got #78 covered.


dequis:
Can we actually safely break this api? Maybe it would be better to keep the original definitions and add another one.

ailin-nemui:
can we remove the ignore_find_noact ?

LemonBoy:
Well the whole point of the latest commit was to keep the api intact, so I think we can't (at least for now, we might be able to get rid of the duplicate functions in a far far future)

PR #392, opened 2016-01-06
irssiproxy: avoid using pointer after freeing it

mauke:


ailin-nemui:
thanks

dequis:
How/where did this show up as a bug? `g_slist_remove` doesn't access the contents of the passed pointers as far as i know.

mauke:
It doesn't have to dereference it. `free(ptr)` makes the value of `ptr` indeterminate; using it (e.g. by passing it to a function) has undefined behavior.

dequis:
I know, but did anything happen in practice?

mauke:
I never ran the code. I just looked through it.

PR #396, opened 2016-01-08
make /knockout use /unban to remove bans

mauke:
/knockout uses /ban to set bans but calls ban_remove() directly to
remove them. This commit makes it use /unban instead. This allows
scripts that hook ban/unban to work automatically with /knockout.


ailin-nemui:
lgtm

PR #398, opened 2016-01-14
Use GLib's regexp interface (backed by PCRE)

LemonBoy:
```
A little less conversation, a little more action please
All this aggravation ain't satisfactionin' me
A little more bite and a little less bark
A little less fight and a little more spark
Shut your mouth and open up your heart and baby satisfy me
Satisfy me baby
```
Up for discussion and test results (my hilight/ignore game is pretty weak) :balloon:

### DIFF
--- a/src/core/ignore.c
+++ b/src/core/ignore.c
@@ -67,12 +67,8 @@ static int ignore_match_pattern(IGNORE_REC *rec, const char *text)
 		return FALSE;
 
 	if (rec->regexp) {
-#ifdef HAVE_REGEX_H
 		return rec->regexp_compiled &&
-			regexec(&rec->preg, text, 0, NULL, 0) == 0;
-#else
-                return FALSE;
-#endif
+		    g_regex_match(rec->preg, text, 0, NULL);
 	}
 
 	return rec->fullword ?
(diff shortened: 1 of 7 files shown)


src/core/ignore.c
```diff
@@ -326,26 +322,23 @@ static void ignore_remove_config(IGNORE_REC *rec)
...
-				REG_EXTENDED|REG_ICASE|REG_NOSUB);
-		if (errcode != 0) {
-			errbuf_len = regerror(errcode, &rec->preg, 0, 0);
-			errbuf = g_malloc(errbuf_len);
-			regerror(errcode, &rec->preg, errbuf, errbuf_len);
-			g_warning("Failed to compile regexp '%s': %s", rec->pattern, errbuf);
-			g_free(errbuf);
+		GError *re_error;
+
+		rec->preg = g_regex_new(rec->pattern, G_REGEX_CASELESS | G_REGEX_EXTENDED, 0, &re_error);
```
dequis: This doesn't mean the same thing as `REG_EXTENDED`

>If a pattern is compiled with the G_REGEX_EXTENDED option, whitespace in the pattern (other than in a character class) and characters between a # outside a character class and the next newline are ignored. An escaping backslash can be used to include a whitespace or # character as part of the pattern. 

For posix regex `REG_EXTENDED` means using the extended flavor instead of the basic one (so irssi defaults to not have to escape `(|)` and others)

LemonBoy: My bad, here's an excerpt from `pcreposix(3)` for more informations.

    I have implemented only those POSIX option bits that can be
    reasonably mapped to PCRE native options. In addition, the option
    REG_EXTENDED is defined with the value zero. This has no effect, but
    since programs that are written to the POSIX interface often use it,
    this makes it easier to slot in PCRE as a replacement library. Other
    POSIX options are not even defined.

dequis:
Similar weechat PR mentioning some issues that are relevant for us too: https://github.com/weechat/weechat/pull/627

wilhelmy:
I would like to express my approval to generally replacing libc regex, since using it makes configs from different systems incompatible with each other because of syntax differences like e.g. \< vs [[:<:]] on GNU vs BSD.

Issue #175, opened 2014-11-11
-botcmd only works on small channels


madduck:
It's my understanding that `autosendcmd` (`-botcmd`) only works on channels with less than 1,000 users. The reason seems to be that it runs when receiving the output of `/who` and that call seems to be skipped when the channel has a large number of users for obvious reasons.

Could this be made configurable?

Alternatively, maybe there'd be an easy way to enable running of command after `/join` that are not dependent on `/who` results?

ailin-nemui:
@madduck are you interested in evaluating the fix proposed by LemonBoy

madduck:
also sprach ailin-nemui <notifications@github.com> [2016-01-20 04:01 +1300]:
> @madduck are you interested in evaluating the fix proposed by LemonBoy

Sorry, I have not seen the fix… until now. I just left a comment and
will gladly try the patch next week when back at my desk.

Best regards,

-- 
The sender e-mail address used (reply.github.com@pobox.madduck.net)
is valid and specific to our correspondence. It should not indicate any
affiliation with your organisation.
 
Die verwendete E-mail-Adresse (reply.github.com@pobox.madduck.net)
ist gültig und spezifisch für unsere Korrespondenz. Sie soll in keiner Weise
auf eine Verbindung mit Ihrer Organisation hindeuten.
 
Spamtrap: reply.github.com.bogus@pobox.madduck.net

madduck:
I tried the fix and it works. Thanks!

ailin-nemui:
closing via #399

PR #399, opened 2016-01-14
Change when the autocmds are sent.

LemonBoy:
As per #175 if a -botcmd is specified for a given channel without a
-bots parameter then the command is sent right after joining the
channel.

### DIFF
--- a/src/core/channels.c
+++ b/src/core/channels.c
@@ -245,9 +245,18 @@ void channel_send_autocommands(CHANNEL_REC *channel)
 	if (rec == NULL || rec->autosendcmd == NULL || !*rec->autosendcmd)
 		return;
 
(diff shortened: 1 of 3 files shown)


ailin-nemui:
mhm, maybe 2 separate signal handlers would be better and more clear

ailin-nemui:
@madduck could you also re-test this alternate approach?

madduck:
That also seems to work.

I am not sure I like it better though, as it requires more conditionals in each of the hooks. I feel like the first approach with a little prose in a comment to help people understand might be better.

But using the `botmask` variable as a "global state" to decide whether to fire `autosendcmd` in the first or the second hook does does seem like another way to do it, and it's not even unelegant. ;)

ailin-nemui:
thanks

Issue #195, opened 2015-01-01
re-resolve DNS record if the connection broke


wilhelmy:
```
[23:48] [Freenode] !morgan.freenode.net Server Terminating. Received SIGTERM
[23:48] *** Irssi: warning SSL read error: server closed connection unexpectedly
[23:48] [Freenode] *** Irssi: Connection lost to 2610:150:4b0f::
[23:48] *** Irssi: Removed reconnection to server 2610:150:4b0f:: port 7070
[23:48] [Freenode] *** Irssi: Looking up 2610:150:4b0f::
[23:48] [Freenode] *** Irssi: Reconnecting to 2610:150:4b0f:: [2610:150:4b0f::] port 7070 
          - use /RMRECONNS to abort
[23:48] *** Irssi: warning SSL handshake failed: Broken pipe
[23:48] [Freenode] *** Irssi: Connection lost to 2610:150:4b0f::
[23:53] *** Irssi: Removed reconnection to server 2610:150:4b0f:: port 7070
[23:53] [Freenode] *** Irssi: Looking up 2610:150:4b0f::
[23:53] [Freenode] *** Irssi: Reconnecting to 2610:150:4b0f:: [2610:150:4b0f::] port 7070 
          - use /RMRECONNS to abort
[23:53] *** Irssi: warning SSL handshake failed: Broken pipe
[23:53] [Freenode] *** Irssi: Connection lost to 2610:150:4b0f::
[23:58] *** Irssi: Removed reconnection to server 2610:150:4b0f:: port 7070
[23:58] [Freenode] *** Irssi: Looking up 2610:150:4b0f::
[23:58] [Freenode] *** Irssi: Reconnecting to 2610:150:4b0f:: [2610:150:4b0f::] port 7070 
          - use /RMRECONNS to abort
[23:58] *** Irssi: warning SSL handshake failed: Broken pipe
[23:58] [Freenode] *** Irssi: Connection lost to 2610:150:4b0f::
```
It keeps going on like this until I manually issue `/rmreconns` and `/connect Freenode`
Ideally, it would re-resolve the rrDNS on chat.freenode.net and connect to a different IP in the round-robin in case the connection fails.

khepler:
Also a problem when a server moves to a different IP.  Had a client constantly trying to connect to an old IP after DNS pointed to the new one.  Only picked up the new IP after /rmreconns and /connect cycle.

ailin-nemui:
:+1:

[...]

PR #409, opened 2016-01-27
Reverse lookup flags

LemonBoy:
Stacked on top of #408.
Might solve #195, can't say for sure since I'm not able to reproduce the error, see the comments made on the issue for more details.


ailin-nemui:
yolo

PR #410, opened 2016-01-27
Getaddrinfo v6 flag

LemonBoy:
Stacked on top of #408.

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -29,10 +29,6 @@ configure options
 
   Build the irssi proxy (see startup-HOWTO).
 
-  --disable-ipv6
-
-  Disable IPv6 support.
-
   --disable-ssl
 
   Disable SSL support.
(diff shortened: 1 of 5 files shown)


ailin-nemui:
I am not sure how this would affect reconnects after interface configuration changes (IPv6 added etc)

LemonBoy:
The currently behaviour is the worst case scenario with this. 10/10 would merge.

ailin-nemui:
currently  you can use -4 or -6 to choose which server you want to connect to, but if you do not even retrieve those IPs then that won't be possible anymore

LemonBoy:
If there are no AAAA records that means there's no interface capable of connecting to an ipv6 address, so the `-6` is going to fail anyway.
With this patch we just avoid returning an unsupported address.

ailin-nemui:
let's try this then

PR #412, opened 2016-01-29
Use GLib's regexp interface (backed by PCRE)

LemonBoy:
Pros:
- Uniform syntax between systems (as brought up by @wilhelmy)
- More people are familiar with the syntax used by PCRE
- Faster (?)
- More robust utf8 support

Cons:
- People may need to update their regexps
- We need to make sure to feed utf8 encoded strings

This proposal is up for discussion :balloon: 

---

@ailin-nemui
I would really like to see some test cases wrt. UTF8 and with and without the raw flags and some thoughts about how to go best about this (compile 2 regexen and match them depending on utf state?) there is also some secret unicode command flag in pcre iirc, does this apply?

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -144,6 +144,15 @@ AC_ARG_ENABLE(true-color,
 	fi,
 	want_truecolor=no)
 
+AC_ARG_ENABLE(gregex,
+[  --disable-gregex     Build without GRegex (fall back to regex.h)],
+	if test x$enableval = xno ; then
+		want_gregex=no
+	else
+		want_gregex=yes
+	fi,
+	want_gregex=yes)
+
 dnl **
 dnl ** just some generic stuff...
 dnl **
@@ -534,6 +543,12 @@ else
 	want_truecolor=no
 fi
 
+if test "x$want_gregex" = "xyes"; then
+	AC_DEFINE([USE_GREGEX], [], [use GRegex for regular expressions])
+else
+	want_gregex=no
+fi
+
 AH_TEMPLATE(HAVE_GMODULE)
 AH_TEMPLATE(HAVE_SOCKS_H, [misc..])
 AH_TEMPLATE(HAVE_STATIC_PERL)
@@ -648,6 +663,7 @@ echo
 
 echo "Building with 64bit DCC support .. : $offt_64bit"
 echo "Building with true color support.. : $want_truecolor"
+echo "Building with GRegex ............. : $want_gregex"
 
 echo
 echo "If there are any problems, read the INSTALL file."
--- a/configure.ac
+++ b/configure.ac
@@ -21,7 +21,7 @@ AC_PATH_PROG(perlpath, perl)
 AC_CHECK_HEADERS(unistd.h dirent.h sys/ioctl.h sys/resource.h)
 
 # check posix headers..
-AC_CHECK_HEADERS(sys/socket.h sys/time.h sys/utsname.h regex.h)
+AC_CHECK_HEADERS(sys/socket.h sys/time.h sys/utsname.h)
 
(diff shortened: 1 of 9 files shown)


ailin-nemui:
I think this is a good idea. what we need is
* examples of breakage in regexes when using the new engine
* documentation for users how to manually adjust their regexes

ailin-nemui:
we (probably) need to conditionally use the raw flag. Two cases:
- irssi is being used with 8bit term_type (-> everything is bytes)
- incoming messages are not valid utf8

ailin-nemui:
we should consider if it shouldn't be possible to disable the RAW mode? otherwise we cannot benefit from the unicode advantage in gregex

foice:
Does this support negative lookahead? I'd like to try it, I get irssi from brew on OS X if that matters.

vague666:
@foice , see man pcresyntax, I assume that's the place to read about pcre and what it supports (it mentions negative look ahead)

vague666:
@LemonBoy @ailin-nemui how about using both regex systems in parallel for a while to let people adjust to the change and use a setting to switch between the two? It would make the codebase bigger for the time being but might not give users as much headache, or is there a library that you know of(I didn't find anything with google) that would allow conversion of POSIX regex to PCRE?

dequis:
>is there a library that you know of(I didn't find anything with google) that would allow conversion of POSIX regex to PCRE?

Converting POSIX regex to PCRE would mean trying to emulate it while keeping bug-compatibility with it. Nope. Maybe it's a matter of adding backslashes to some characters and removing them from others, but we might end up learning about missed edge cases after releasing a new version breaks stuff. I don't want to go there.

Better allow switching regex engines with a global setting.

ailin-nemui:
I would really like to see some test cases wrt. UTF8 and with and without the raw flags and some thoughts about how to go best about this (compile 2 regexen and match them depending on utf state?) there is also some secret unicode command flag in pcre iirc, does this apply?

ailin-nemui:
I will merge this, issues caused by that can then be found and solved later

dequis:
I really don't like the idea of **replacing** the regexp engine.

ailin-nemui:
Why?

dequis:
It's a breaking change, and a subtle one at that. People will upgrade irssi and find their stuff stopped working for no particular reason. Let's not do that. The improvements brought by this change aren't worth it. I'd rather have it under a setting.

ailin-nemui:
how about making it a compile time switch

ailin-nemui:
@LemonBoy

lotheac:
dequis wrote:
> It's a breaking change, and a subtle one at that. People will upgrade
> irssi and find their stuff stopped working for no particular reason.
> Let's not do that. The improvements brought by this change aren't
> worth it. I'd rather have it under a setting.

Exactly this happened to me -- I realize that this is a compile-time
setting, but if subtle breakage was to be avoided, then it should not
have been the default.

ailin-nemui:
sorry for your inconvenience, I pushed for this change but dequis was against it. I added a further hint about it to the on-site release notes (cannot change the released notes anymore)

would you like to expand on what regex broke and how you had to modify it?

lotheac:
On Mon, Jan 09 2017 01:32:33 -0800, ailin-nemui wrote:
> sorry for your inconvenience, I pushed for this change but dequis was against it. I added a further hint about it to the on-site release notes (cannot change the released notes anymore)

Thanks.

> would you like to expand on what regex broke and how you had to modify it?

Sure. I was using \< to match a word boundary in some of my highlights,
and I had to replace that with the PCRE \b (\> is likely similarly
affected).

-- 
Lauri Tirkkonen | lotheac @ IRCnet

PR #414, opened 2016-01-30
add an append operation to cut buffer handling

toddpratt:
also, checks that g_new succeeded and NOOPs if it failed.


dequis:
I don't think we care about handling memory allocation failures...

Also, use tabs for indentation instead of spaces.

toddpratt:
OK.  On a previous PR, someone suggested I check the success of g_new(), but either way is fine with me.

src/fe-text/gui-entry.c
```diff
@@ -582,25 +582,36 @@ void gui_entry_erase(GUI_ENTRY_REC *entry, int size, CUTBUFFER_UPDATE_OP update_
 	if (size == 0 || entry->pos < size)
 		return;
 
+        if (entry->cutbuffer_len == 0) {
+          update_cutbuffer = CUTBUFFER_UPDATE_REPLACE;
```
ailin-nemui: -> fix indentation

PR #416, opened 2016-01-31
Use the locale's preferred date format for daychange notifications.

lanurmi:
Locale data was already used for formatting dates, but the result is more or less awkward, depending on language.
This commit will change the formatting of the "Day changed to ..." line even for English.

On the other hand, if it is undesirable to change the date format for English, perhaps the date shouldn't obey locale in the first place.
Note that the additional comment to `irssi.c` is already true whether or not this PR is merged or not.


ailin-nemui:
mhm. I think it's more clear to spell out the month by default, so I'm not so sure about this change. Note that you can easily change this locally in your theme

PR #422, opened 2016-02-13
Clean up misc.c

LemonBoy:
The weekend's approaching, feel free to review this and/or pick some of the suggestions below:

- [x] `find_substr` could be replaced by something less hairy and simple
- [ ] `gslist_find_{,icase}_string` could be implemented in terms of `g_slist_find_custom`
- [ ] Same as above for `gslist_foreach_find`
- [ ] `stristr` can be replaced with the nonstandard but portable `strcasestr`
- [ ] `strstr_full_case` could be implemented in terms of the above or via a regexp
- [ ] `convert_home` should support the more generic `~USERNAME/` format
- [x] `g_istr_hash` could use the `djb` hash function (the same used by the case-sensitive version)
- [ ] `match_wildcards` sounds like something that a regexp could do just fine
- [ ] `str_to_uofft` doesn't care about overflowing nor about errors

### DIFF
--- a/src/common.h
+++ b/src/common.h
@@ -6,7 +6,7 @@
 #define IRSSI_GLOBAL_CONFIG "irssi.conf" /* config file name in /etc/ */
 #define IRSSI_HOME_CONFIG "config" /* config file name in ~/.irssi/ */
 
-#define IRSSI_ABI_VERSION 5
+#define IRSSI_ABI_VERSION 6
 
 #define DEFAULT_SERVER_ADD_PORT 6667
 
--- a/src/core/ignore.c
+++ b/src/core/ignore.c
@@ -241,7 +241,7 @@ IGNORE_REC *ignore_find_full(const char *servertag, const char *mask, const char
 		if (channels == NULL || rec->channels == NULL)
 			continue; /* other doesn't have channels */
 
-		if (strarray_length(channels) != strarray_length(rec->channels))
+		if (g_strv_length(channels) != g_strv_length(rec->channels))
 			continue; /* different amount of channels */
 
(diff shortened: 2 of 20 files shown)


dequis:
Remember third party modules are a thing. Maybe it's safer to replace those misc.c functions with straight wrappers of the equivalent glib functions. Wouldn't help with the regexp_match removal though.

I think we should make a list of public third party modules that we should review for usages. Like... https://irssi.org/modules/

The changes themselves seem good, at least as far as i reviewed.

ailin-nemui:
so we can merge this once we have verified the modules

LemonBoy:
Just a reminder, the checkboxes up there are for items up for discussion :balloon:

ailin-nemui:
All the following commits seem fine to me and can be merged IMO

-  Replace mkpath with g_mkdir_with_parents  
-   Check the return value of open() in rawlog.c 
-  Replace strocpy with g_strlcpy 
-  Replace strarray_length with g_strv_length 

-  Replace hashtable_get_keys with g_hash_table_get_keys 
^ In that one I see some code duplication ....
-  Update the g_istr_hash function to use the djb hash  
-  str_to_uofft is a tiny wrapper over strtoul{,l} 


This commit is probably wrong:

-  Replace ascii_str{up,down} with g_ascii_str{up,down} 
: Returns a **newly-allocated string**, with all the upper case characters in str converted to lower case

-  Implement find_substr with strarray_find 
^ The original is probably more efficient


INSTALL should mention glib 2.16 (minimum version in configure.ac)

LemonBoy:
Updated

PR #425, opened 2016-02-18
irssi proxy: allow multiplexing multiple networks over a single port

mauke:


dequis:
Would have been better to avoid unrelated formatting changes, it makes the diff harder to review.

The implementation looks good. Haven't tested it yet.

Not a huge fan of the `?` on the UI side but I see how it makes sense to keep the code simpler.

ailin-nemui:
tip for @dequis use https://github.com/irssi/irssi/pull/425/files?w=1 for review . sadly does not allow line comments in that view

mauke:
I've been running this for a few days now. No problems so far.

PR #426, opened 2016-02-23
Make pasting warning appear when long pastes are going to be split into many lines

Manishearth:
This can cause accidental channel floods when the message being pasted is longer
than you thought it was, but doesn't contain newlines (and will not trigger the
usual Ctrl-K/Ctrl-C pasting warning)

### DIFF
--- a/src/fe-text/gui-readline.c
+++ b/src/fe-text/gui-readline.c
@@ -40,6 +40,9 @@
 #include <string.h>
 #include <signal.h>
 
+/* After LINE_SPLIT_LIMIT characters, the message will be split into multiple lines */
+#define LINE_SPLIT_LIMIT 400
+
 typedef void (*ENTRY_REDIRECT_KEY_FUNC) (int key, void *data, SERVER_REC *server, WI_ITEM_REC *item);
 typedef void (*ENTRY_REDIRECT_ENTRY_FUNC) (const char *line, void *data, SERVER_REC *server, WI_ITEM_REC *item);
 
@@ -227,7 +230,7 @@ static void paste_buffer_join_lines(GArray *buf)
 	}
 
 	/* all looks fine - now remove the whitespace, but don't let lines
-	   get longer than 400 chars */
+	   get longer than LINE_SPLIT_LIMIT chars */
 	dest = arr; last_lf = TRUE; last_lf_pos = NULL; line_len = 0;
 	for (i = 0; i < buf->len; i++) {
 		if (last_lf && isblank(arr[i])) {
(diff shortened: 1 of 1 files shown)


src/fe-text/gui-readline.c
```diff
@@ -227,7 +230,7 @@ static void paste_buffer_join_lines(GArray *buf)
 	}
 
 	/* all looks fine - now remove the whitespace, but don't let lines
-	   get longer than 400 chars */
+	   get longer than LINE_SPLIT_LIMIT(400) chars */
```
LemonBoy: Do not mention the 400 in the comment here

src/fe-text/gui-readline.c
```diff
@@ -357,11 +360,23 @@ static void insert_paste_prompt(void)
...
 	char *str;
 
+    /* The actual number of lines that will show up post-line-split */
+    int actual_line_count = paste_line_count;
+
+    /* in case this prompt is happening due to line-splitting, calculate the
+       number of lines obtained from this. The number isn't entirely accurate;
+       we just choose the greater of the two since the exact value isn't
+       important */
+    if (paste_buffer->len / LINE_SPLIT_LIMIT > paste_verify_line_count &&
```
LemonBoy: Have a temporary variable for `paste_buffer->len / LINE_SPLIT_LIMIT`

src/fe-text/gui-readline.c
```diff
@@ -638,7 +653,8 @@ static gboolean paste_timeout(gpointer data)
 {
 	paste_was_bracketed_mode = paste_bracketed_mode;
 
-	if (paste_line_count == 0) {
+    /* Take into account the fact that a line may be split every LINE_SPLIT_LIMIT characters */
+	if (paste_line_count == 0 && paste_buffer->len / LINE_SPLIT_LIMIT <= paste_verify_line_count) {
```
LemonBoy: I think this should be consolidated with another flag `paste_have_lines` (?)

Manishearth: What would that flag do? I think sharing the verify length count is fine here, they become the same in the end anyway?

Manishearth:
Anything else we could do? I think that using the exact value for the line limit from irc-servers.c is more accurate, but unnecessary since this warning is mostly about figuring out if you're pasting something too big or too small.

ailin-nemui:
not all protocols require line splitting so the warning would be misleading in that case...

src/fe-text/gui-readline.c
```diff
@@ -638,7 +654,11 @@ static gboolean paste_timeout(gpointer data)
 {
 	paste_was_bracketed_mode = paste_bracketed_mode;
 
-	if (paste_line_count == 0) {
+	/* number of lines after splitting extra-long messages */
+	int split_lines = paste_buffer->len / LINE_SPLIT_LIMIT;
+
+	/* Take into account the fact that a line may be split every LINE_SPLIT_LIMIT characters */
+	if (paste_line_count == 0 && split_lines <= paste_verify_line_count) {
```
ailin-nemui: I'm not this is correct. try pasting a very long line without linebreaks with and without your patch

Manishearth: I'm not sure what you mean here; I've tested that (that's the entire point of this PR), it works.

ailin-nemui: ok I see now sorry for the confusion

PR #432, opened 2016-03-08
fix SASL EXTERNAL

grawity:
First commit makes the authentication actually _work_ rather than hanging on "You have not registered".

<del>Second commit adds authzid support, similar to how it is implemented in PLAIN</del>.

[As tradition goes,](https://github.com/irssi/irssi/commit/b8d3301d34f383f039071214872570385de1bb59) the latter is untested as of now.


grawity:
Upon further consideration, sending the current username as authzid doesn't make _enough_ sense to be worth the extra code, so that part was dropped.

(Although in that case, PLAIN should be updated to send an empty authzid as well.)

PR #434, opened 2016-03-09
Check for tag/ before tag/#channel for network activity_hide_targets

vague666:


vague666:
LemonBoy asked for commentary: this lets you create a network-wide mask, instead of having to specify Freenode/#chan1 Freenode/#chan2 etc for /set activity_hide_targets Freenode/ is enough to hide activity in all Freenode targets

LemonBoy:
Please add a small comment in the code too

PR #436, opened 2016-03-11
fix IRC ischannel function

### LINKED ISSUE #435: joining +channels is broken
the ischannel function removes the "statusmsg" characters and assumes those are @ and + thus breaking `@channels` and `+channels`. 

the ischannel function must not break `/join`. maybe we need to separate ischannel for joins and ischannel for privmsgs.

furthermore, statusmsgs should not be assumed to be `@+`, if it is not present and no WALLCHOPS and no WALLVOICES are in isupport then it should be treated as empty (imo).

CHANTYPES should always be checked prior to stripping statusmsgs
Regression due to #248 

cf https://github.com/irssi/irssi/blob/master/src/irc/core/irc-servers.c#L84

ailin-nemui:
fixes #435

### DIFF
--- a/src/irc/core/irc-servers.c
+++ b/src/irc/core/irc-servers.c
@@ -78,17 +78,47 @@ static int ischannel_func(SERVER_REC *server, const char *data)
 	IRC_SERVER_REC *irc_server = (IRC_SERVER_REC *) server;
 	char *chantypes, *statusmsg;
 
+	/* empty string is no channel */
+	if (*data == '\0')
+		return FALSE;
+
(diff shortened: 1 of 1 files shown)


LemonBoy:
I think that the problem lies in our defaults, I think the only valid sequence is `[statusmsg]+[chantypes]*` which is the only ordering that makes sense (at least to me) and requires the `statusmsg` and `chantypes` set to be disjoint sets.
The suggestion here is to only keep the new code that infers the `statusmsg` set from the `wallch{ops,voices}` and maybe  shorten the `chantypes` to a saner default of just `#`.

PR #437, opened 2016-03-11
Use glob matching for activity_hide_targets

vague666:


src/core/misc.c
```diff
@@ -181,6 +181,24 @@ int strarray_find(char **array, const char *item)
 	return -1;
 }
 
+
+int strarray_find_glob(char **array, const char *item)
+{
+	char **tmp;
+	int index;
+
+	g_return_val_if_fail(array != NULL, 0);
```
LemonBoy: Hrm, to be honest it might be better to return `-1` instead of 0 to follow the `return index of `item' in `array' or -1 if not found` principle.
(This also applies to `strarray_find`)

LemonBoy:
This deserves a mention in the docs.

LemonBoy:
From `RFC 2812`

```
Channels names are strings (beginning with a '&', '#', '+' or '!'
   character) of length up to fifty (50) characters.  Apart from the
   requirement that the first character is either '&', '#', '+' or '!',
   the only restriction on a channel name is that it SHALL NOT contain
   any spaces (' '), a control G (^G or ASCII 7), a comma (',').  Space
   is used as parameter separator and command is used as a list item
   separator by the protocol).  A colon (':') can also be used as a
   delimiter for the channel mask.  Channel names are case insensitive.
```

and Glib's documentation

```
Note that in contrast to glob(), the '/' character can be matched by the wildcards, there are no '[...]' character ranges and '*' and '?' can not be escaped to include them literally in a pattern. 
```

With the attached patch you have no way to reference a channel containing a `*` or a `?`.

ailin-nemui:
it may have unfortunate consequences if you want to ignore activity in the channel #*

vague666:
Ah, thanks, yes, we don't want that, maybe we can just skip the matching if
ignore targets is a * or tag/*

/Jari

ailin-nemui:
I think this is fine now, @LemonBoy ?

PR #438, opened 2016-03-11
Remove sasl timeout source when the server disconnects

dequis:


LemonBoy:
LGTM

src/irc/core/sasl.c
```diff
@@ -153,6 +153,20 @@ static void sasl_step(IRC_SERVER_REC *server, const char *data, const char *from
 	server->sasl_timeout = g_timeout_add(SASL_TIMEOUT, (GSourceFunc) sasl_timeout, server);
 }
 
+static void sasl_disconnected(IRC_SERVER_REC *server)
+{
+	g_return_if_fail(server != NULL);
+
+	if (!IS_IRC_SERVER(server)) {
```
ailin-nemui: any particular reason you're not using g_return_if_fail here too? not that'd matters

dequis: ...dunno. I copied this part from equivalent signal handlers.

PR #442, opened 2016-03-17
Do not assume any default value for statusmsg.

### LINKED ISSUE #435: joining +channels is broken

LemonBoy:
If the server didn't send it then just skip the check, the old value it
defaulted to was possibly overlapping with the CHANTYPES leading to an
incorrect behaviour. Fixes #435.

### DIFF
--- a/src/irc/core/irc-servers.c
+++ b/src/irc/core/irc-servers.c
@@ -78,17 +78,23 @@ static int ischannel_func(SERVER_REC *server, const char *data)
 	IRC_SERVER_REC *irc_server = (IRC_SERVER_REC *) server;
 	char *chantypes, *statusmsg;
 
+	g_return_val_if_fail(data != NULL, FALSE);
+
+	/* empty string is no channel */
+	if (*data == '\0')
+		return FALSE;
+
 	chantypes = g_hash_table_lookup(irc_server->isupport, "chantypes");
 	if (chantypes == NULL)
 		chantypes = "#&!+"; /* normal, local, secure, modeless */
 
 	statusmsg = g_hash_table_lookup(irc_server->isupport, "statusmsg");
-	if (statusmsg != NULL) {
-		while (strchr(statusmsg, *data) != NULL)
-			data++;
-	}
+	if (statusmsg != NULL)
+		data += strspn(data, statusmsg);
 
-	return strchr(chantypes, *data) != NULL;
+	/* strchr(3) considers the trailing NUL as part of the string, make sure
+	 * we didn't advance too much. */
+	return *data != '\0' && strchr(chantypes, *data) != NULL;
 }
 
 static char **split_line(const SERVER_REC *server, const char *line,
--- a/src/irc/core/irc-servers.c
+++ b/src/irc/core/irc-servers.c
@@ -81,12 +81,12 @@ static int ischannel_func(SERVER_REC *server, const char *data)
 	chantypes = g_hash_table_lookup(irc_server->isupport, "chantypes");
 	if (chantypes == NULL)
 		chantypes = "#&!+"; /* normal, local, secure, modeless */
-	statusmsg = g_hash_table_lookup(irc_server->isupport, "statusmsg");
-	if (statusmsg == NULL)
-		statusmsg = "@+";
 
-	while (strchr(statusmsg, *data) != NULL)
-		data++;
+	statusmsg = g_hash_table_lookup(irc_server->isupport, "statusmsg");
+	if (statusmsg != NULL) {
+		while (strchr(statusmsg, *data) != NULL)
+			data++;
+	}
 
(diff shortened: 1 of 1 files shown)


ailin-nemui:
This fix still is buggy, I cannot part channels on IRCnet anymore

LemonBoy:
More info please ?

dequis:
Did /leave with no parameters

```c
(gdb) bt
#0  ischannel_func (server=<optimized out>, data=0x55d2186d7db0 "") at irc-servers.c:82
#1  0x000055d217313cd4 in get_optional_channel (require_name=0, data=0x7fff76f88320, active_item=0x55d2186d79c0) at commands.c:681
#2  cmd_get_params (data=data@entry=0x55d2173478f4 "", free_me=free_me@entry=0x7fff76f883d0, count=count@entry=73730) at commands.c:740
#3  0x000055d2172f1c8f in cmd_part (data=0x55d2173478f4 "", server=0x55d2186d07a0, item=0x55d2186d79c0) at irc-commands.c:161
```

Note the empty data parameter.

```c
681             } else if (!server_ischannel(active_item->server, channel)) {
682                     /* we don't have channel parameter - use active channel */
683                     ret = window_item_get_target(active_item);
684             } else {
685                     /* Find the channel first and use it's name if found.
```

That server_ischannel function is supposed to return false, but it returns true

```c
(gdb) p chantypes
$9 = 0x55d2186c85f0 "#&!+"
(gdb) p data
$10 = 0x55d2186d7db0 ""
(gdb) p strchr(chantypes, *data)
$11 = 409765364
(gdb) p strchr(chantypes, *data) - (int) chantypes
$12 = 4
```

Nei's PR wasn't affected by this because

```c
+	/* empty string is no channel */
+	if (*data == '\0')
+		return FALSE;
```

So just do that, imo.

src/irc/core/irc-servers.c
```diff
@@ -78,6 +78,9 @@ static int ischannel_func(SERVER_REC *server, const char *data)
 	IRC_SERVER_REC *irc_server = (IRC_SERVER_REC *) server;
 	char *chantypes, *statusmsg;
 
+	g_return_val_if_fail(data != NULL, FALSE);
+	g_return_val_if_fail(*data != '\0', FALSE);
```
ailin-nemui: I don't think that macro is being used correctly here, see https://developer.gnome.org/glib/stable/glib-Warnings-and-Assertions.html#g-return-val-if-fail

fail means precondition check so this is a programmer error, but the irssi code is legally calling the function like this

ailin-nemui:
your function will accept "@" as being a valid channel if the memory allocated after the string happens to be any sequence of \0, @ or # (undefined behaviour due to out-of-bounds read)

LemonBoy:
Should be fine now, I admit I didn't know that `strchr(3)` considers the trailing NUL as part of the string, TIL

ailin-nemui:
sorry to break you the news but you're still in UB with the strchr loop, since you may be reaching end of string any time so you need to check this condition on every loop. Or reconsider strspn?

LemonBoy:
I should pay more attention, no ?

ailin-nemui:
@ahf and @dequis might want to, they didn't feel anything wrong with your initial PR

LemonBoy:
Shots fired :fireworks:

dequis:
:( I had no idea strchr behaved like that

src/irc/core/irc-servers.c
```diff
@@ -78,17 +78,23 @@ static int ischannel_func(SERVER_REC *server, const char *data)
 	IRC_SERVER_REC *irc_server = (IRC_SERVER_REC *) server;
 	char *chantypes, *statusmsg;
 
+	g_return_val_if_fail(data != NULL, FALSE);
+
+	/* empty string is no channel */
```
ailin-nemui: This was just a common-case optimisation in my suggestion, I don't think you need it

LemonBoy: Doesn't hurt to have it

ailin-nemui:
Honestly? This whole odyssey makes me prefer my own suggestion even more. I'm biassed though

dequis:
"odyssey"? Am I missing something here? I don't see it.

PR #443, opened 2016-03-18
Throw an error when a chatnet has no available url

LemonBoy:
I spent too much time thinking `IRCNet` servers were shit but it turns out I had no server block for it and irssi kept trying to connect to my ISP page-not-found redirect.

Hilarious :rocket:


src/core/chat-commands.c
```diff
@@ -73,6 +73,12 @@ static SERVER_CONNECT_REC *get_server_connect(const char *data, int *plus_addr,
 
 	conn = server_create_conn(proto != NULL ? proto->id : -1, addr,
 				  atoi(portstr), chatnet, password, nick);
+	if (conn == NULL) {
+		g_warning("Cuntnugget");
```
dequis: uhm...

LemonBoy: Well, if I get a ACK on the behaviour I'll add a new CMDERR_ later

ailin-nemui:
We need to raise abi version due to formats addition

ailin-nemui:
lgtm

src/core/servers-setup.c
```diff
@@ -323,6 +323,8 @@ server_create_conn(int chat_type, const char *dest, int port,
 		rec = create_chatnet_conn(chatrec->name, port, password, nick);
```
ailin-nemui: just return rec; ?

LemonBoy:
PR updated and ready to :rocket:

PR #444, opened 2016-03-20
Use 0 as a sentinel value for sasl_timeout

LemonBoy:
If sasl_timeout is never initialized with a valid timeout id then
calling /disconnect on the server calls g_source_remove() with 0 as tag,
causing an harmless error message to be printed.
Beside that, the sasl_timeout field is defined as a unsigned int.
We can use 0 as sentiel since g_timeout_add returns tags that are always
greater than zero.


ailin-nemui:
when is the timeout not initialised?

LemonBoy:
When the server doesn't support SASL at all (eg Efnet)

PR #447, opened 2016-03-20
correct quoting in /help dcc

ailin-nemui:


ailin-nemui:
I will merge this for 0.8.18a, the help should be re-updated in #453

PR #448, opened 2016-03-21
terminfo-core: Don't send smkx/rmkx

### LINKED ISSUE #430: Binding meta-O causes arrow keys to stop working
I am using `/bind meta-O change_window 49`. In irssi 0.8.17 and earlier this works fine, but in 0.8.18 my arrow keys stop functioning normally — they insert ABCD characters instead.

A fresh config plus this binding suffices to reproduce the issue.

dequis:
Partially reverts b5b73cb.

Partially, because the commit is mostly unrelated whitespace changes, but also because I left the TI_smkx and TI_rmkx caps in there, to reconsider later.

-----

Fixes #430.

As the branch name says, this is the lazy solution to the problem. We're a bit tight on time and the not-lazy one might break more than it fixes.

(Also, I personally use st and don't even have that kind of backspace/del behavior. I must have "accidentally" fixed it by editing the config file)

If someone cares we should make this a setting, but that's after the bugfix release.

### DIFF
--- a/src/fe-text/terminfo-core.c
+++ b/src/fe-text/terminfo-core.c
@@ -532,9 +532,6 @@ void terminfo_cont(TERM_REC *term)
 	if (term->TI_smcup)
 		tput(tparm(term->TI_smcup));
 
-	if (term->TI_smkx)
-		tput(tparm(term->TI_smkx));
-
         terminfo_input_init(term);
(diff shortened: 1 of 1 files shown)


dequis:
Closing in favor of #459

PR #451, opened 2016-03-21
Reset the status of the bracketed paste mode on exit

LemonBoy:
Closes (?) #449


ailin-nemui:
we probably need to do this also on suspend

LemonBoy:
We don't have any way to get the user preference (afair) at suspension-time

ailin-nemui:
we need to keep track of it in the terminal file (static variable). That'll also save us from sending the escape sequence more often than required

ailin-nemui:
(the same approach will be used for toggling the kx bit in the future)

PR #452, opened 2016-03-21
Don't call terminfo_cont() twice on resume

### LINKED ISSUE #450: potential terminal state not reset after 2x suspend
testing with zsh, after suspending irssi 2x with Ctrl+Z, I don't get working Ctrl+C anymore (for example if I then launch cat). Curiously enough it doesn't seem to happen in bash.

LemonBoy:
It was only called to push the current state onto the stack since the
terminal would be reset in the SIGCONT handler.
Fixes some weirdness when using ^Z with zsh.

Fixes #450, at least for me.

### DIFF
--- a/src/fe-text/term-terminfo.c
+++ b/src/fe-text/term-terminfo.c
@@ -611,8 +611,6 @@ void term_stop(void)
 {
 	terminfo_stop(current_term);
 	kill(getpid(), SIGTSTP);
-	terminfo_cont(current_term);
-	irssi_redraw();
 }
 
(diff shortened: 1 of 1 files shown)


ailin-nemui:
thanks for looking into it. I'm not convinced this is enough, I'll comment more later

ailin-nemui:
are there systems which do not properly send sigcont, thus warranting or being the reason for the original code?

LemonBoy:
Not that I am aware of

ailin-nemui:
title/message still needs update

LemonBoy:
Updated

PR #453, opened 2016-03-21
Support quoted filenames in some /DCC commands

LemonBoy:


ailin-nemui:
did you test this yourself?

LemonBoy:
Tested right now, I confirm it's working as expected

PR #455, opened 2016-03-21
improved cutbuffer handling

ailin-nemui:
This adds cutbuffer handling building on the work of toddpratt. It adds two new keys which you can bind in /bind: yank_next_cutbuffer and append_next_kill. Consecutive kills are appended to the current cutbuffer


ailin-nemui:
@toddpratt

PR #456, opened 2016-03-22
cutbuffer: do not unconditionally use replace when noop was requested

### LINKED ISSUE #454: the first backspaced character is incorrectly stored in the kill buffer
@toddpratt

ailin-nemui:
Fixes #454


LemonBoy:
lgtm

PR #457, opened 2016-03-22
fix race condition in terminal init

### LINKED ISSUE #450: potential terminal state not reset after 2x suspend

ailin-nemui:
remove the tcgetattr call to a single time on irssi load instead of
querying it each time. Fixes #450

### DIFF
--- a/src/fe-text/terminfo-core.c
+++ b/src/fe-text/terminfo-core.c
@@ -496,7 +496,7 @@ void terminfo_setup_colors(TERM_REC *term, int force)
 	}
 }
 
-static void terminfo_input_init(TERM_REC *term)
+static void terminfo_input_init0(TERM_REC *term)
 {
 	tcgetattr(fileno(term->in), &term->old_tio);
 	memcpy(&term->tio, &term->old_tio, sizeof(term->tio));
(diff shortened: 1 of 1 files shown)


ailin-nemui:
(I hope)

LemonBoy:
I think there's no need to call `terminfo_term_init` again after the kill SIGSTP returns, a strace shows that the terminal settings are applied twice because of that

ailin-nemui:
they may be applied twice because of term_init being called twice, once from the do_redraw sigcont handler https://github.com/irssi/irssi/blob/master/src/fe-text/term-terminfo.c#L91 and once from the kill-stop function https://github.com/irssi/irssi/blob/master/src/fe-text/term-terminfo.c#L610

LemonBoy:
Exactly, and #452 fixes the problem by sending the smcup only (if available) when we resume from the TSTP signal, we save the current state and then the old settings are re-applied in the SIGCNT handler

ailin-nemui:
in that case I wonder if even the smcup of the code in your suggestion from #452 is necessary? if so, why?

LemonBoy:
The smcup saves the terminal window to be re-applied with a rmcup on the next resume/exit

ailin-nemui:
but terminfo_cont (which is run by the sigcont hander) also runs smcup so why do we need to send *that* twice?

LemonBoy:
You're right, upon further inspection I've updated the PR

ailin-nemui:
after discussion on IRC I conclude that we want to have both this and the #452 eventually

PR #458, opened 2016-03-22
Properly toggle bracketed paste mode on stop/cont

### LINKED ISSUE #449: bracketed paste mode not deinitialised on leaving irssi

ailin-nemui:
Fixes #449

### DIFF
--- a/src/fe-text/terminfo-core.c
+++ b/src/fe-text/terminfo-core.c
@@ -395,6 +395,14 @@ static void _ignore_parm(TERM_REC *term, int param)
 {
 }
 
+static void term_dec_set_bracketed_paste_mode(int enable)
+{
+	if (enable)
+		tputs("\e[?2004h", 0, term_putchar);
+	else
+		tputs("\e[?2004l", 0, term_putchar);
+}
+
 static void term_fill_capabilities(TERM_REC *term)
 {
 	int i, ival;
@@ -535,6 +543,9 @@ void terminfo_cont(TERM_REC *term)
 	if (term->TI_smkx)
 		tput(tparm(term->TI_smkx));
 
+	if (term->bracketed_paste_enabled)
+		term_dec_set_bracketed_paste_mode(TRUE);
+
         terminfo_input_init(term);
 }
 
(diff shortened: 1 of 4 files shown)


src/fe-text/terminfo-core.c
```diff
@@ -48,6 +48,8 @@ typedef struct {
 
 TERM_REC *current_term;
 
+static int bracketed_paste_enabled;
```
LemonBoy: Move this into TERM_REC

src/fe-text/terminfo-core.c
```diff
@@ -677,6 +689,17 @@ static int term_setup(TERM_REC *term)
         return 1;
 }
 
+void term_set_bracketed_paste_mode(int enable)
+{
+	if (enable != bracketed_paste_enabled) {
```
LemonBoy: Too much noise for nothing

```
if (enable == bp_mode)
    return;
```

src/fe-text/terminfo-core.c
```diff
@@ -535,6 +537,11 @@ void terminfo_cont(TERM_REC *term)
 	if (term->TI_smkx)
 		tput(tparm(term->TI_smkx));
 
+	if (bracketed_paste_enabled) {
+		bracketed_paste_enabled = FALSE;
+		term_set_bracketed_paste_mode(TRUE);
```
LemonBoy: If `term_set_bracketed_paste_mode` didn't touch `term_set_bracketed_paste_mode` this could be just `term_set_bracketed_paste_mode(bracketed_paste_enabled)`, the code's too confusing

ailin-nemui: we need to keep track of state. suggestions?

LemonBoy: `term_set_bracketed_paste_mode(bracketed_paste_enabled)` on CONT,
`term_set_bracketed_paste_mode(FALSE)` on stop, have `term_set_bracketed_paste_mode` only send the escape sequences, move `bracketed_paste_enabled` into `TERM_REC` and profit ?

ailin-nemui: I guess that's possible. I kinda dislike sending the same sequence over and over though... currently any random setting change will re-set that mode in the terminal

LemonBoy: We shouldn't be messing with the tty settings so frequently but that's another problem, anyway that's not a problem since it's just a bunch of bytes and I'd vouch for simple and clear code instead

ailin-nemui:
better?

LemonBoy:
:sparkling_heart:

PR #459, opened 2016-03-22
Make use of terminal application keys configurable

### LINKED ISSUE #430: Binding meta-O causes arrow keys to stop working

ailin-nemui:
adds a new setting term_appkey_mode which can enable or disable the use
of keyboard transmit (application keys) mode. Fixes #430


LemonBoy:
lgtm

PR #461, opened 2016-03-24
Revert "Removed the obsolete SQUERY and SERVLIST commands"

ailin-nemui:
This reverts commit 4beebe3238dda96efd4c3a0a6f6002ad5afc96e7.

### DIFF
--- a/docs/help/in/list.in
+++ b/docs/help/in/list.in
@@ -25,10 +25,10 @@
 %9Remarks:%9
 
     Not all networks support server-side filtering and may provide a network
-    service instead; on IRCnet, you may use the ALIS service:
+    service or service bot instead; on IRCnet, you may use the List service:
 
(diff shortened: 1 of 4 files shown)


LemonBoy:
This belongs to a script

ailin-nemui:
These are rfc commands,  idk why they should be removed

LemonBoy:
The help files in the commit you mentioned state they only work for IRCNet, is this still true?

ailin-nemui:
They work on any network running ircd (like ircnet)  or any network which chooses to implement them. Rfc documents don't seem any worse than "ircv3" to me...

LemonBoy:
I was trying to undestand if those commands were effectively obsolete and/or exclusively for ircnet ( as the documentation leads to believe). Lgtm as long as the ‎misunderstanding is cleared up in the doc

Issue #420, opened 2016-02-10
Find a less-global method of deciding when to print netsplits/mode/joins


GinjaNinja32:
When irssi is looking for more netsplit QUITs, more MODEs, etc, _any_ printed text to _any_ window will cause the accumulated QUITs/MODEs/etc to be printed; I have a window set up as a 'raw' window, printing exactly what irssi receives from connected servers, with no processing. My script to do so currently requires stopping the "print starting" signal while I'm trying to print;

``` perl
Irssi::signal_add_first("print starting", "ignore_signal");
$window->print($text, MSGLEVEL_NEVER) if ($window);
Irssi::signal_remove("print starting", "ignore_signal");
```

If I don't have this in place, the result is one netsplit line for every user, printed immediately (because my 'raw' script has printed to my raw-input window, causing netsplit detection to conclude the end of the splitting users):

```
-!- Netsplit *.net <-> *.split quits: SomeUser
-!- Netsplit *.net <-> *.split quits: AnotherUser
-!- Netsplit *.net <-> *.split quits: SomeOtherUser
```

Could this be done in a way that it doesn't rely on watching for printed text? Perhaps if the server sends something that is _not_ another QUIT/MODE/etc?

LemonBoy:
I agree it *shouldn't* absolutely hook the `print starting` signal, but I can't think of another way to solve this than to print the `parts` minus `joins` after the netsplit is over (after the timer runs out) but that way you don't get any feedback in case the person you're talking with has been split (in half ?)

ailin-nemui:
it could be done more granular for example on a server/channel flag and/or excluding certain message levels. could be much effort...

dequis:
Can we consider this fixed by #465?

dequis:
So this goes back to unfixed.

PR #465 tried to fix this

Issue #567 was a regression from that fix

PR #577 fixed that regression

Issue #809 appeared much later and its cause is still not fully understood.

PR #812 reverted most of the original fix (made the filtering a no-op)

So to get this done we have to figure out a fix to #809 that isn't a revert.

ailin-nemui:
one thing I noticed

https://github.com/irssi/irssi/blob/7de1378dab8081932d9096e19ae3d0921e560230/src/fe-common/irc/fe-netsplit.c#L111
here the whole split is marked as printed. note how this gets triggered even when you try to filter out a single channel for display; it would need to be done per channel

ailin-nemui:
[...]

PR #465, opened 2016-04-02
Some small adjustments to the netsplit code.

LemonBoy:
Wrt issue #420

This is a RFC, I repeat, this is a RFC :warning:

### DIFF
--- a/src/fe-common/irc/fe-netjoin.c
+++ b/src/fe-common/irc/fe-netjoin.c
@@ -164,7 +164,7 @@ static void print_channel_netjoins(char *channel, TEMP_PRINT_REC *rec,
 	g_free(channel);
 }
 
-static void print_netjoins(NETJOIN_SERVER_REC *server)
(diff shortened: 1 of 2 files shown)


ailin-nemui:
I haven't reviewed it but it does seem like a good improvement over the previous approach

src/fe-common/irc/fe-netjoin.c
```diff
@@ -235,20 +241,31 @@ static void print_netjoins(NETJOIN_SERVER_REC *server)
...
+	 * c) There's a NETJOIN record for this server
+	 */
+	if (!IS_IRC_SERVER(dest->server))
+		return;
 
-		next = tmp->next;
-		if (server->netjoins != NULL)
-			print_netjoins(server);
-	}
+	if (dest->level != MSGLEVEL_MSGS)
```
ailin-nemui: MSGS should be private messages, so we are never printing the netjoins?

PR #472, opened 2016-04-26
Remove dane support and drop the dependency on dnsval

LemonBoy:


ahf:
Needs the alternative implementation to be in place though.

PR #474, opened 2016-04-27
Window creation tweaks

seveas:
I group my windows per topic, and want private messages well away from them.
These patches add options to let me do that. The first patch is best viewed
with git log -p -w as git's diff machinery gets a bit confused by it.

### DIFF
--- a/src/fe-common/core/fe-windows.c
+++ b/src/fe-common/core/fe-windows.c
@@ -47,7 +47,7 @@ static int window_get_new_refnum(void)
 	GSList *tmp;
 	int refnum;
 
-	refnum = 1;
+	refnum = settings_get_int("window_create_min_number");
 	tmp = windows;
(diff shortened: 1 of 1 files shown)


ailin-nemui:
Thanks for your submission. I feel like this may be better off implemented as a script, what do you think?

```
use List::Util qw(max);
Irssi::signal_add_first(
    'window created' => sub {
        my ($win) = @_;
        if ($win->{refnum} < 10) {
            $win->set_refnum(max 10, 1+ max map { $_->{refnum} } Irssi::windows); } })
```

seveas:
Script works too, didn't think it would be possible with a script, so thanks for writing the script for me :smile:

ailin-nemui:
feel free to submit an improved and configurable script to scripts.irssi.org

PR #475, opened 2016-04-27
Don't print 'day changed' messages until something happens

seveas:
I usually have hundreds of windows open, most are private messages. This
patch prevents them from all showing up as having unread things in the
window list, and also prevents long series of just 'day changed to'
messages in the more idle queries.

I didn't see how to do this from a script, hence another patch to irssi.

### DIFF
--- a/src/fe-common/core/fe-windows.c
+++ b/src/fe-common/core/fe-windows.c
@@ -575,7 +575,7 @@ static void sig_server_disconnected(SERVER_REC *server)
 	}
 }
 
-static void window_print_daychange(WINDOW_REC *window, struct tm *tm)
+void window_print_daychange(WINDOW_REC *window, struct tm *tm)
 {
         THEME_REC *theme;
         TEXT_DEST_REC dest;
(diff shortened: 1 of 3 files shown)


vague666:
You can use /format -delete daychange instead

seveas:
But then I wouldn't get the message at all, right?

vague666:
That is correct, but why do you need it? Only case you'd need it might be if you don't use timestamps. You can also check which level the daychanged format is written and add it to /set activity_hide_levels

ailin-nemui:
hi, I like the idea but it needs to be configurable and I don't know if I like the direction your code is going.. I'm thinking about some constructive comments ;-)

ailin-nemui:
after looking through it I conclude it's probably fine if you just make it configurable!

dequis:
I was thinking the ugly cast in `((WINDOW_REC *)(tmp->data))->daychanged = 1;` could go away if it's turned into a function call like `window_daychange(tmp->data)` which sets that variable or prints it right away, depending on the settings.

ailin-nemui:
@seveas

ailin-nemui:
@seveas thanks for your contribution. feel free to reopen once you addressed those comments. thanks for your understanding

Issue #421, opened 2016-02-12
Add an option to make /hilight case sensitive


LemonBoy:
What it says on the tin

ailin-nemui:
closing via #476 please test

PR #476, opened 2016-04-28
Add an option to make /hilight case sensitive

isundil:
See #421

### DIFF
--- a/src/fe-common/core/hilight-text.c
+++ b/src/fe-common/core/hilight-text.c
@@ -122,8 +122,11 @@ static void hilight_init_rec(HILIGHT_REC *rec)
 {
 #ifdef HAVE_REGEX_H
 	if (rec->regexp_compiled) regfree(&rec->preg);
-	rec->regexp_compiled = !rec->regexp ? FALSE :
-		regcomp(&rec->preg, rec->text, REG_EXTENDED|REG_ICASE) == 0;
+	if (!rec->regexp)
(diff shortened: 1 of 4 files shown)


isundil:
I'm not sure I understand Travis, the outputed error seems to be non-blocking on the others pull-requests

LemonBoy:
Looks fine code-wise with one minor remark about the choice of names, right now you use `-case` and `matchcase` in the command line and the config respectively and it'd be nice to have both named the same way.

ailin-nemui:
Thanks for your contribution, if I'm not mistaken the code is lacking the regex case

isundil:
Yes, I thought they were a regexp flag in preg, but obviously it doesn't

src/fe-common/core/hilight-text.c
```diff
@@ -200,19 +204,26 @@ static int hilight_match_text(HILIGHT_REC *rec, const char *text,
...
 #endif
 	} else {
-		match = rec->fullword ?
-			stristr_full(text, rec->text) :
-			stristr(text, rec->text);
+		if (rec->case_sensitive) {
+			match = rec->fullword ?
+				strstr_full(text, rec->text) :
+				strstr(text, rec->text);
+		}
```
ailin-nemui: please keep `} else {` style

ailin-nemui:
thanks. please increase ABI version in src/common.h (change in struct/header file)

PR #478, opened 2016-05-02
Fix some ANSI-C issues

pisculichi:


dequis:
Thanks but all of those are perfectly valid c99. Do you have a specific example of compiler/platform that doesn't support this? Because we don't support microsoft visual studio.

pisculichi:
it's for Haiku-os gcc2 (https://github.com/haikuports/haikuports/pull/553)

ailin-nemui:
Are you sure those casts are required?

pisculichi:
@ailin-nemui honestly no, I'll check if are required.

src/fe-text/gui-entry.c
```diff
@@ -582,12 +583,14 @@ char *gui_entry_get_cutbuffer(GUI_ENTRY_REC *entry)
 
 char *gui_entry_get_next_cutbuffer(GUI_ENTRY_REC *entry)
 {
+	GUI_ENTRY_CUTBUFFER_REC *tmp = NULL;
```
ailin-nemui: = NULL not necessary here

src/fe-text/gui-entry.c
```diff
@@ -652,7 +655,8 @@ void gui_entry_erase(GUI_ENTRY_REC *entry, int size, CUTBUFFER_UPDATE_OP update_
 	if (update_cutbuffer != CUTBUFFER_UPDATE_NOOP) {
 		int cutbuffer_new_size;
 		unichar *tmpcutbuffer;
-		GUI_ENTRY_CUTBUFFER_REC *tmp = get_cutbuffer_rec(entry, update_cutbuffer);
+		GUI_ENTRY_CUTBUFFER_REC *tmp;
```
ailin-nemui: this change is not necessary as far as I can tell

src/fe-text/gui-readline.c
```diff
@@ -696,10 +696,11 @@ static void key_append_next_kill(void)
 
 static gboolean paste_timeout(gpointer data)
 {
+	int split_lines = 0;
```
ailin-nemui: this initialisation is also not required

pisculichi:
Fixed!

PR #479, opened 2016-05-09
improve perl @INC detection

### LINKED ISSUE #132: sometimes wrong/useless path set for perl_use_lib
And, after much trial and error....its because in 1.70 it switched from List::Util to List::AllUtils and wanting the 'any' method.  Which exists in List::Util, providing its new enough....either from using Perl 5.20 or having Scalar-List-Utils installed.

The problem is on systems where the latter is being done.

Scalar-List-Utils installs a List::Util in SITEARCH to override the base version in ARCH.  And site_perl paths normally come before the base paths in @INC.

However, with irssi, the configure figures out what the INSTALLARCHLIB path is and promotes that to being searched first.  Which causes the older List::Util to get loaded and eventually break things like 'twirssi'. (unable to determine local time zone)

The explanation I gleaned is that its trying to determine where Irssi.pm is going to get installed and making that come first.  Problem is that Irssi.pm goes under INSTALLSITEARCH, which by default is first...

While the work around is to use '/set perl_use_lib ...', it might be nice this wasn't needed or required so much hair pulling....

ailin-nemui:
might fix #132


ailin-nemui:
personally I would go so far as to suggest packagers that they ``./configure --with-perl-lib=/usr/lib64/irssi/perl5/$(perl -e 'print substr $^V, 1') `` instead of the current `--with-perl-lib=site` default. Maybe  we can change our default in the future

dequis:
I don't know enough about perl or autotools to "review" this, so just commenting to say hi.

ailin-nemui:
hi @dequis !

gedge:
this does work on my system, with a minor tweak (that may need to be put into the FreeBSD port patches) on line 483 (to add the make flag `--no-print-directory`)

ailin-nemui:
hi, does it work with `-s` instead of --no-print-directory (I added such a commit above)? latter is not portable

PR #480, opened 2016-05-13
Handle utf8 nicks with mk_wcwidth()

xavierog:
Following pull request #471, I created a whole new branch in my fork (hence that separate pull request) which, I believe, reflects ailin-nemui's advices. Feel free to review, comment and/or integrate these changes.


src/core/utf8.c
```diff
@@ -0,0 +1,135 @@
...
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+
+#include "utf8.h"
+#include "module.h"
+#include "wcwidth.c"
+/* Provide is_utf8(): */
+#include "recode.h"
+
+int string_advance(char const **str, int policy)
```
ailin-nemui: can you convert the policy to enum?

ailin-nemui:
thank you very much for your continued effort on this, I'm looking forward to integrating this into irssi.

mhm, it looks like we don't need the characters count and probably never will. (my suggestion was in error) how would you like renaming string_chars_for_width and make it return the bytes instead?

I'm a bit ambivalent about the policy-autodetection with -1 (tending to dislike). anyone else got an opinion on that? @dequis ?

src/core/utf8.c
```diff
@@ -0,0 +1,135 @@
...
+		*str += 1;
+
+		return 1;
+	}
+}
+
+int string_policy(const char *str)
+{
+	if (is_utf8()) {
+		if (!str || g_utf8_validate(str, -1, NULL)) {
```
ailin-nemui: in irssi code, please always use explicit `str == NULL`

src/core/utf8.c
```diff
@@ -0,0 +1,135 @@
...
+
+int string_chars_for_width(const char *str, int policy, unsigned int n, unsigned int *bytes)
+{
+	const char *c, *previous_c;
+	int str_width, char_width, char_count;
+
+	g_return_val_if_fail(str != NULL, -1);
+
+	/* Handle the dummy case where n is 0: */
+	if (!n) {
```
ailin-nemui: write `n == 0`

src/core/utf8.c
```diff
@@ -0,0 +1,135 @@
...
+		if (bytes) {
```
ailin-nemui: `bytes != NULL`

src/core/utf8.c
xavierog: Sure, no problem :)

src/core/utf8.c
xavierog: Sure, no problem :)

src/core/utf8.c
xavierog: Sure, no problem :)

src/core/utf8.c
xavierog: Sure, no problem :)

xavierog:
> mhm, it looks like we don't need the characters count and probably never will. (my suggestion was in error) how would you like renaming string_chars_for_width and make it return the bytes instead?

Actually, the function returns the amount of characters and optionally provides the number of bytes. It turns out that my former pull request needed characters whereas my new pull request needs bytes; therefore, it does not seem unlikely to me that the number of characters could be needed in the future (never say never...). Plus, "he who can do more can do less": it does not hurt to have extra features like this in that kind of toolbox (because this is what we are doing here: glib and libc functions are not sufficient for us, so we are building our own little toolbox on top of them); for instance: have you noticed that string_length() is absolutely unused as of now? :)

> I'm a bit ambivalent about the policy-autodetection with -1 (tending to dislike).

I implemented that approach because, when writing the code, it seemed obvious to me that, without this mechanism, we would just end up writing a lot of "foobar(str, string_policy(str))" everywhere in the code and possibly regret it at some point as it goes against one of the most fundamental principle of programming: DRY -- Don't Repeat Yourself. How come you disapprove of it?

PR #485, opened 2016-06-06
Do not crash on OPTCHAN when item has no server

ailin-nemui:
May fix [bugs.debian.org#826525](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=826525)


dequis:
Tested, it does fix it.

That's an unusually good bug report.

josephbisch:
Hi, I'm the reporter of the bug in Debian. My laptop unfortunately stopped working, so I am unable to confirm that the change does fix the bug. Probably won't be able to confirm at all in the near future, unless I were to do so over SSH from my phone, which I don't plan on doing.

PR #488, opened 2016-06-06
Be smart about case-matching the nicks.

LemonBoy:
Not very sure about this, we might want to gate this behaviour under a configurable option...
Long story short, it makes the completion match the case of the user-typed prefix if it contains at least an uppercase letter, I found it quite disappointing that `Ne<tab>` didn't give `Nei` but another nick starting with a lowercase `n`.

### DIFF
--- a/src/fe-common/core/chat-completion.c
+++ b/src/fe-common/core/chat-completion.c
@@ -38,11 +38,18 @@
 #include "chat-completion.h"
 #include "window-items.h"
 
+enum {
+	COMPLETE_MCASE_NEVER = 0,
+	COMPLETE_MCASE_ALWAYS,
+	COMPLETE_MCASE_AUTO,
+};
+
 static int keep_privates_count, keep_publics_count;
 static int completion_lowercase;
 static char *completion_char, *cmdchars;
 static GSList *global_lastmsgs;
 static int completion_auto, completion_strict;
+static int completion_match_case;
 
 #define SERVER_LAST_MSG_ADD(server, nick) \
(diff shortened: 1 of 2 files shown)


ailin-nemui:
maybe a tristate /set completion_nicks_matchcase uppercase/always/never ?

LemonBoy:
Yeah, I'd use `YES/NO/AUTO` as options though.

ailin-nemui:
can you check this vs. rejoin_channels_on_reconnect and make sure we use the same code paths in all cases

src/fe-common/core/chat-completion.c
```diff
@@ -52,6 +58,18 @@ static int completion_auto, completion_strict;
 	last_msg_add(&((MODULE_CHANNEL_REC *) MODULE_DATA(channel))->lastmsgs, \
 		     nick, own, keep_publics_count)
 
+static int contains_uppercase(const char *s1)
```
ailin-nemui: gboolean

src/fe-common/core/chat-completion.c
```diff
@@ -43,6 +43,12 @@ static int completion_lowercase;
 static char *completion_char, *cmdchars;
 static GSList *global_lastmsgs;
 static int completion_auto, completion_strict;
+/*
+ * 0 = Don't match the case
+ * 1 = Match the case
+ * 2 = Don't match the case unless the user typed at least a uppercase letter
+ */
+static int completion_match_case;
```
ailin-nemui: let's use enum constants?

Issue #491, opened 2016-06-11
Implement SETTING_TYPE_ENUM


ailin-nemui:
will allow us to handle more gracefully /set actlist_sort, rejoin_channels_on_reconnect and completion_match_case, by giving proper list of allowed settings on /set (and maybe tab-completion of values?)

-  completion_match_case in #488 
-  actlist_sort: https://github.com/irssi/irssi/blob/master/src/fe-text/statusbar-items.c#L417
-  rejoin_channels_on_reconnect: https://github.com/irssi/irssi/blob/master/src/irc/core/irc-servers.c#L630

Add setting type here: https://github.com/irssi/irssi/blob/master/src/fe-common/core/fe-settings.c#L137
and https://github.com/irssi/irssi/blob/master/src/core/settings.h

valid setting choices can be specified in the call to setting_add_enum

ailin-nemui:
Implemented in #492

PR #492, opened 2016-06-12
Enum backwards is munE

LemonBoy:
yo, #491 is here.
After some quick testing it does work pretty well (see the XXX note), I've only ported `rejoin_channels_mode` so far to see how ergonomic the end result is.
I've initially made `settings_add` to use a `char **` as a list of choices but decided to use a comma separated string instead so that's easy to use from the perl side too.
Completion is included :tada:


ailin-nemui:
I like it, besides what we discussed on irc the error message has to move to fe-settings and a TXT . the code in settings should probably just fail silently by returning false?

src/fe-common/core/fe-settings.c
```diff
@@ -99,6 +100,24 @@ static void set_int(const char *key, const char *value)
...
+static void set_choice(const char *key, const char *value)
+{
+	char *stripped_value;
+
+	stripped_value = g_strdup(value);
+	g_strstrip(stripped_value);
+
+	if (settings_set_choice(key, stripped_value) == FALSE) {
+		SETTINGS_REC *rec = settings_get_record(key);
+		char *msg = g_strjoinv(",", rec->choices);
```
ailin-nemui: ", " would look nicer

PR #503, opened 2016-06-30
Better interaction between reverse-DNS resolution and reconnection.

### LINKED ISSUE #494: Reverse dns-resolution doesn't play well with /reconnect
When the reverse dns resolution is enabled the resolved hostname replaces the original one in the connrec structure, but this doesn't play well with the reconnection commands.
`cmd_reconnect` blindly tries to reconnect to the same server and `sig_reconnect` fails in a similar way (I guess) since `server_setup_find` fails to match the address when comparing it with the list of `setupservers`.

LemonBoy:
Closes #494

### DIFF
--- a/src/core/server-rec.h
+++ b/src/core/server-rec.h
@@ -36,6 +36,7 @@ unsigned int server_operator:1;
 unsigned int usermode_away:1;
 unsigned int banned:1; /* not allowed to connect to this server */
 unsigned int dns_error:1; /* DNS said the host doesn't exist */
+unsigned int dns_reverse:1;
 
 GTimeVal lag_sent; /* 0 or time when last lag query was sent to server */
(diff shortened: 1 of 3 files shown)


ailin-nemui:
what is the desired behaviour? what is the current behaviour and how does it change after this patch? maybe the connrec rather shouldn't be updated by reverse resolve at all?

LemonBoy:
The desired behaviour is to use the non-resolved address on subsequent reconnections.
I didn't change how the connrec is updated since I wanted to make the patch to be as least invasive as possible.

ailin-nemui:
I think we should better analyse this and fix the underlying design issue (like updating connrec) if that turns out to be the root issue

ailin-nemui:
see #494

PR #505, opened 2016-07-01
Add "GBytes *" as a possible perl signal arg

kruton:
This allows the possibility of sending an immutable byte vector as an
argument to signals. This is helpful for upcoming changes to support
SASL scripts.

### DIFF
--- a/src/perl/perl-signals.c
+++ b/src/perl/perl-signals.c
@@ -80,6 +80,7 @@ void perl_signal_args_to_c(
                 unsigned long v_ulong;
                 GSList *v_gslist;
                 GList *v_glist;
+                GBytes *v_gbytes;
         } saved_args[SIGNAL_MAX_ARGUMENTS];
         void *p[SIGNAL_MAX_ARGUMENTS];
         PERL_SIGNAL_ARGS_REC *rec;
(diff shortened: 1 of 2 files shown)


kruton:
@LemonBoy I'm splitting up the SASL script change into smaller pieces that are easier to logically review.

src/perl/perl-signals.c
```diff
@@ -291,6 +300,14 @@ static void perl_call_signal(PERL_SCRIPT_REC *script, SV *func,
...
 			   int type; as it's first variable (dcc) */
 			perlarg = simple_iobject_bless((SERVER_REC *) arg);
+		} else if (g_strcmp0(rec->args[n], "gbytes") == 0) {
+			GBytes *tmp;
+			const char *data;
+			STRLEN data_len;
+
+			tmp = arg;
+			data = g_bytes_get_data(tmp, &data_len);
+			perlarg = newSVpvn(data, data_len);
```
kruton: I'm not sure of the lifecycle of this perl argument. Should I be allocating from the heap and memcpy this for Perl?

ailin-nemui:
You can add multiple logical commits to a single pull request

dequis:
Since #506 moved to gstring instead of gbytes (because of the glib 2.32 version requirement), this one should be updated.

(Or maybe drop this PR and put it together with the other one. I don't care either way. Github no longer sucks to review individual commits within a PR)

ailin-nemui:
thanks for your contribution, according to comments I will close this for now. that is not meant to discourage you, please reopen when required

PR #506, opened 2016-07-04
SASL: handle fragmentation

kruton:
The IRCv3 SASL extension says that AUTHENTICATION payloads of exactly
400 bytes in length indicate that the message is fragmented and will
continue in a subsequent message. Handle the reassembly and splitting of
these messages so that we are compliant with the specification.

### DIFF
--- a/src/irc/core/sasl.c
+++ b/src/irc/core/sasl.c
@@ -26,6 +26,19 @@
 #include "irc-servers.h"
 #include "sasl.h"
 
+/*
+ * Based on IRCv3 SASL Extension Specification:
+ * http://ircv3.net/specs/extensions/sasl-3.1.html
+ */
+#define AUTHENTICATE_CHUNK_SIZE 400 // bytes
+
+/*
+ * Maximum size to allow the buffer to grow to before the next fragment comes in. Note that
+ * due to the way fragmentation works, the maximum message size will actually be:
+ * floor(AUTHENTICATE_MAX_SIZE / AUTHENTICATE_CHUNK_SIZE) + AUTHENTICATE_CHUNK_SIZE - 1
+ */
+#define AUTHENTICATE_MAX_SIZE 8192 // bytes
+
 #define SASL_TIMEOUT (20 * 1000) // ms
 
 static gboolean sasl_timeout(IRC_SERVER_REC *server)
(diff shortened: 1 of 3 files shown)


dequis:
Travis failed because you used `g_string_free_to_bytes` which was added in glib 2.34, a bit too recent.

src/irc/core/sasl.c
```diff
@@ -105,49 +113,149 @@ static void sasl_success(IRC_SERVER_REC *server, const char *data, const char *f
...
-	IRC_SERVER_CONNECT_REC *conn;
-	GString *req;
-	char *enc_req;
+	char *buffer;
+	GBytes *decoded;
+
+	/* Check if there is an existing fragment to prepend. */
+	buffer = g_hash_table_lookup(sasl_buffers, server->tag);
+	if (buffer != NULL) {
+		if (!g_strcmp0("+", enc_req)) {
```
LemonBoy: Use an explicit comparison (`== 0`), same as below.

kruton:
Okay, think my latest commit in this PR addresses the comment.

LemonBoy:
Why not use a `GBytes` in `sasl_buffers` instead of a plain string ? And what do you think about adding a upper limit on the data that we collect ?

src/irc/core/sasl.c
```diff
@@ -105,49 +113,149 @@ static void sasl_success(IRC_SERVER_REC *server, const char *data, const char *f
...
+	char *buffer;
+	GBytes *decoded;
+
+	/* Check if there is an existing fragment to prepend. */
+	buffer = g_hash_table_lookup(sasl_buffers, server->tag);
+	if (buffer != NULL) {
+		if (g_strcmp0("+", enc_req) == 0) {
+			enc_req = buffer;
+		} else {
+			enc_req = g_strconcat(buffer, enc_req, NULL);
```
dequis: This replaces the received end_req with a longer version, so the check for `strlen(enc_req) == AUTHENTICATE_CHUNK_SIZE` below will never pass after the first chunk.

It also sets a string that should be freed (and is leaking) in a variable that shouldn't.

src/irc/core/sasl.c
```diff
@@ -105,49 +113,149 @@ static void sasl_success(IRC_SERVER_REC *server, const char *data, const char *f
...
+			g_string_append(resp, conn->sasl_username);
+			g_string_append_c(resp, '\0');
+			g_string_append(resp, conn->sasl_username);
+			g_string_append_c(resp, '\0');
+			g_string_append(resp, conn->sasl_password);
 
-			enc_req = g_base64_encode((const guchar *)req->str, req->len);
+			/* This could be g_string_free_to_bytes */
+			resp_len = resp->len;
+			resp_bytes = g_bytes_new_take(g_string_free(resp, FALSE), resp_len);
```
dequis: Added in glib 2.32 (current requirement is 2.16, shouldn't be bumped past 2.28 because rhel 6)

src/irc/core/sasl.c
```diff
@@ -105,49 +113,149 @@ static void sasl_success(IRC_SERVER_REC *server, const char *data, const char *f
...
+	const guchar *p;
+	gsize len;
+	size_t offset, enc_len, chunk_len;
+
+	if (response == NULL) {
+		irc_send_cmdv(server, "AUTHENTICATE +");
+		return;
+	}
+
+	p = g_bytes_get_data(response, &len);
```
dequis: Added in glib 2.32, ditto.

dequis:
Uh, you know, just avoid GBytes entirely. I thought it was some functions but all of it was added in glib 2.32

kruton:
Doh! I rewrote this using `GString` instead. I also wrote unit tests to confirm everything works as expected.

kruton:
I removed a now-superfluous comment; should be ready for another look.

dequis:
Instead of using a hash table from tag to gstring, you could just add a GString * field to IRC_SERVER_REC, which would also make it easier to free the GString buffer if the authentication is interrupted for whatever reason. Also, I think that in those cases with the current version the partial sasl payload would persist across disconnections. That's not desirable.

Also +1 for an upper limit to stop servers from making this buffer grow forever. Atheme seems to use 8192. https://github.com/atheme/atheme/blob/59ca306d8faf1b7a69a82e3d37bb3c751e680f13/modules/saslserv/main.c#L318

kruton:
Added the SASL buffer to IRC_SERVER_REC and set an upper limit to 8192.

PR #509, opened 2016-07-06
Make autoclose_window work when /disconnect'ing

### LINKED ISSUE #508: autoclose_windows ON but windows are still open on /disconnect
When /disconnecting from a server/network, all orphaned windows remain open, even if `autoclose_windows` is set to `ON`

This happens as of irssi 0.8.19 (20160323 0008)
It also happened with 0.8.17 ...

LemonBoy:
I guess it makes sense.
Closes #508.

### DIFF
--- a/src/core/servers.c
+++ b/src/core/servers.c
@@ -449,6 +449,8 @@ static int server_remove_channels(SERVER_REC *server)
 		CHANNEL_REC *channel = tmp->data;
 
 		next = tmp->next;
+		/* mark all the channels as parted */
+		channel->left = TRUE;
 		channel_destroy(channel);
 		found = TRUE;
 	}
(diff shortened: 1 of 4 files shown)


ailin-nemui:
not sure about this one I feel like there maye a reason

dequis:
Yeah this is subtle stuff. I wouldn't touch it.

ailin-nemui:
ehm, I meant we should think twice, not cannot touch ;)

LemonBoy:
For reference, [here](https://github.com/LemonBoy/irssi/blob/cf3361e5fdc2b90924519ccc8479b7fcf4e77670/src/fe-common/core/fe-channels.c#L91) is where new bindings for the windows are introduced during the disconnection phase.

ailin-nemui:
one potential use case, leaving windows open so that once you connect again the channels are rejoined there. whereas autoclose_windows' target was the case where /part should close a window without requiring /wc if no items are left

LemonBoy:
Hmm, maybe we could just add a `-kill-win` argument to `/disconnect` ?

strk:
`-kill-win` argument to `/disconnect` sounds like a good start. It would be even better if it could be configured to be the default for `/disconnect` (thus allowing passing a -keep-win).

ailin-nemui:
I will close this and welcome any PR which adds a setting autoclose_windows_on_disconnect (maybe that should also close queries?)

PR #510, opened 2016-07-08
Bump the minimum glib version to 2.28

LemonBoy:
`( dx) feel free to bump it up to 2.28 (rhel 6)`

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -4,7 +4,7 @@
 
 To compile irssi you need:
 
-- glib-2.6 or greater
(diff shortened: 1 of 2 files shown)


dequis:
Okay, but I expected this to be done together with changes that actually need it.

This drops support for debian 6.0 (oldstable, squeeze) and ubuntu 10.04 lts, both EOL'd already.

ailin-nemui:
please integrate this in a patch that actually requires it

PR #511, opened 2016-07-09
Follow the XDG specification.

LemonBoy:
We're still 100% backward-compatible.

Supersedes #51, we now default to `~/.config/irssi` if and only if it exists, otherwise `~/.irssi` is used as before.

### DIFF
--- a/src/common.h
+++ b/src/common.h
@@ -1,8 +1,6 @@
 #ifndef __COMMON_H
 #define __COMMON_H
 
-#define IRSSI_DIR_FULL "%s/.irssi" /* %s == g_get_home_dir() */
-
 #define IRSSI_GLOBAL_CONFIG "irssi.conf" /* config file name in /etc/ */
 #define IRSSI_HOME_CONFIG "config" /* config file name in ~/.irssi/ */
 
--- a/src/core/core.c
+++ b/src/core/core.c
@@ -182,16 +182,21 @@ void core_register_options(void)
 
 void core_preinit(const char *path)
 {
-	const char *home;
 	char *str;
 	int len;
 
 	if (irssi_dir == NULL) {
-		home = g_get_home_dir();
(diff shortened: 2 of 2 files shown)


dequis:
:-1:. Quoting previous rejected PR:

>Also, given users often need to know where the Irssi directory is (e.g. to add scripts) I think adding an alternate location is just going to make it confusing

And

>In general I don't think the cost of this is worth it. (Remember Irssi has users who are not hugely familiar with Unix and use this on a shell account; the cost of making a change and then obsoleting random tutorials on the web is very large).

ailin-nemui:
scripts have to be in $XDG_DATA_DIR and so on...

LemonBoy:
I took a simpler approach and jumped over all the bikeshedding, let's consider scripts as part of the configuration itself and have everything in `~/.config`, let's not be so strict.

dequis:
Well, no one said it explicitly, but the commit that happened after my comment changed the meaning of this PR, and it's less awful regarding backwards compatibility.

```diff
+		/* Try the XDG-compliant path first, if the config directory
+		 * exists use it instead of the one at ~/.irssi */
````

I'm still not very enthusiastic about this whole thing. But at least it doesn't "break" new installs.

>let's consider scripts as part of the configuration itself and have everything in ~/.config, let's not be so strict

I do like this mindset

ailin-nemui:
in my opinion we should either have a fully compliant solution with scripts in .local/share/irssi/scripts and maybe a new autorun concept or no XDG

Earnestly:
> 53. So many good ideas are never heard from again once they embark in a voyage on the semantic gulf.

dequis:
@Earnestly Some ideas are also bad.

ailin-nemui:
I'll closse this for now, either a complete implementation or none. also voted by @dequis

PR #513, opened 2016-07-12
Allow CWD-relative paths with /script load

LemonBoy:
Replace the usage of printfs with glib functions to build the paths.

### DIFF
--- a/src/perl/perl-core.c
+++ b/src/perl/perl-core.c
@@ -293,6 +293,9 @@ PERL_SCRIPT_REC *perl_script_load_file(const char *path)
 
         g_return_val_if_fail(path != NULL, NULL);
 
+	if (g_file_test(path, G_FILE_TEST_IS_REGULAR) == FALSE)
+		return NULL;
+
         name = script_file_get_name(path);
         return script_load(name, path, NULL);
 }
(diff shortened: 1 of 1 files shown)


ailin-nemui:
doesn't that break ./ pointing to ~/.irssi/scripts?

src/perl/perl-core.c
```diff
@@ -354,10 +357,11 @@ PERL_SCRIPT_REC *perl_script_find_package(const char *package)
 /* Returns full path for the script */
 char *perl_script_get_path(const char *name)
 {
-	struct stat statbuf;
 	char *file, *path;
 
-	if (g_path_is_absolute(name) || (name[0] == '~' && name[1] == '/')) {
+	if (g_path_is_absolute(name) ||
+	    (name[0] == '~' && name[1] == '/') ||
+	    (name[0] == '.' && name[1] == G_DIR_SEPARATOR)) {
```
ailin-nemui: inconsistent usage of such things as G_DIR_SEPARATOR. do we want to introduce this windows madness or not

ailin-nemui:
question if we should map ./ to cwd. it doesn't *seem* very useful to me

LemonBoy:
Different strokes for different folks, I find that mapping `./` to entirely different folders depending on the context to be quite counter-intuitive, after all `./` introduces a cwd-relative path.

src/perl/perl-core.c
dequis: Apparently all the versions of windows derived from NT support / as a path separator. So, everything newer than 9x/ME. I don't think we support non-cygwin windows anyway.

[x] burninate

PR #514, opened 2016-07-12
Add an option to stop the connection when SASL fails.

LemonBoy:


ailin-nemui:
I guess it's fine, I'm a bit sceptic of the internal flag and related code duplication but whatever. @dequis ?

ailin-nemui:
(should update abi constant)

src/irc/core/irc-servers.h
```diff
@@ -79,6 +79,7 @@ struct _IRC_SERVER_REC {
 	int cap_complete:1;    /* We've done the initial CAP negotiation */
 
 	guint sasl_timeout; /* Holds the source id of the running timeout */
+	int sasl_success:1; /* Did we authenticate successfully ? */
```
dequis: Put this after `cap_complete:1`, otherwise it won't take advantage of the bitfield.

dequis:
That seems ok, but i'm a bit worried about that `server_disconnect` call. Should check that the code that runs after the signal is emitted doesn't depend on the connection still existing.

ahf:
`sasl_disconnect_on_fail` sounds wrong in my head. I would at least change it to something like `sasl_disconnect_on_failure` or `_on_error`.

Should the policy also be enforced on a global "per-client" level?

dequis:
Renaming to _on_failure seems ok (but i don't feel too strongly about it)

>Should the policy also be enforced on a global "per-client" level?

What do you mean? It looks like a global setting to me. I think that's fine.

ahf:
Ack. I think I wrote the wrong question: should it be possible to control this on a per connection level.

dequis:
I don't think we need to have this as a per-connection setting. It's more of a personal preference, similar to "do i care about appearing decloaked or not"

dequis:
Should this also set `server->no_reconnect = TRUE` too?

ailin-nemui:
Maybe the sasl failure is temporary

PR #516, opened 2016-07-12
Fix the tab completion for paths starting with ./

LemonBoy:


ailin-nemui:
which issue exactly does this fix?

I found the following issue:
- /load <tab> - no completion
- /load <tab> with hidden files present, and this patch, - no completion
- /load a<tab> - completes files with a*
- /load .<tab> and this patch, - no completion
- /load .<tab> and this patch, completes ./.*

at least with this patch the no-completion behaviour seems to be more consistent. but I wonder if it shouldn't always go into completion (of all files) in such case?

LemonBoy:
I honestly don't remember anymore.

ailin-nemui:
github ate my `<tab>`s. maybe we can fix it to complete all files if required in that spot?

ailin-nemui:
if you don't have a good explanation for a patch it cannot be merged imo

dequis:
The title is misleading, it makes more sense in the comment in the code:

```
/* Skip the dotfiles unless the user explicitly asked us
 * to do so. Basename might be './', beware of that */
```

And I think there's something mistyped in the comment above, maybe should be:

- `/load <tab>` with this patch, no completion
- `/load <tab>` without this patch, completes to .* (hidden files in the current directory)
- `/load a<tab>`  with or without this patch, completes files with a*
- `/load .<tab>` with this patch, no completion
- `/load .<tab>` without this patch, completes ./.*

Adding to that

- `/load ./<tab>` with this patch, no completion
- `/load ./<tab>` without this patch, completes ./.*

The last one is certainly unexpected - you're referring to the local directory and it completes dotfiles. Same thing with the case of empty input, that's weird

I think this makes sense now.

ailin-nemui:
no idea what this does, I think completion still is less than ideal

dequis:
yolomerge

PR #521, opened 2016-08-04
remove curses terminal and ncurses macro

ailin-nemui:
the curses code had been broken for a while (doesn't compile)


ailin-nemui:
may be good if some of the obscure platform users test this

Issue #500, opened 2016-06-24
Random crash


LemonBoy:
Reported by `KindOne` on irc.

[gdb backtrace](https://github.com/irssi/irssi/files/332574/gRiZ70ba.txt) & [script list](https://github.com/irssi/irssi/files/332577/DDGSVVnQ.txt) are available.
The culprit is [this line](https://github.com/irssi/irssi/blob/master/src/fe-common/core/fe-messages.c#L622).

*(edited to include logs in the ticket description)*

----

Internet connection went a little crazy and I got this crash when znc was constantly reconnecting to networks.

```
irssi <---> localhost <---> znc <---> irc networks 
```

```
username@localhost ~ $ irssi -v
irssi 0.8.20-g1a6ec1b (20160518 1657)

username@localhost ~ $ znc -v
ZNC 1.7.x-git-611-b7309cd - http://znc.in
IPv6: yes, SSL: yes, DNS: threads, charset: yes, build: autoconf, i18n: no
```

```
Program received signal SIGSEGV, Segmentation fault.
__strchr_sse2_bsf () at ../sysdeps/i386/i686/multiarch/strchr-sse2-bsf.S:97
97	../sysdeps/i386/i686/multiarch/strchr-sse2-bsf.S: No such file or directory.
No locals.
#1  0x08087d51 in sig_nicklist_new (channel=0x8aa8160, nick=0x8aac738) at fe-messages.c:622
        firstnick = 0x8aace20
        newnick = <optimized out>
        nickhost = <optimized out>
        p = <optimized out>
        n = <optimized out>
#2  0x080d1713 in signal_emit_real (rec=rec@entry=0x81593f8, params=params@entry=3, va=<optimized out>, 
    va@entry=0xbffff3a8 "`\201\252\b Ϊ\bQV\252\b`\201\252\b", first_hook=0x81a4138) at signals.c:242
        arglist = {0x8aa8160, 0x8aace20, 0x8aa5651, 0x0, 0x0, 0x0}
        prev_emitted_signal = 0x818b610
        hook = 0x81a4138
        prev_emitted_hook = 0x818b630
        i = <optimized out>
        stopped = 0
        stop_emit_count = 0
        continue_emit_count = 0
        __func__ = "signal_emit_real"
#3  0x080d1bcb in signal_emit (signal=0x80e6d6e "nicklist changed", params=3) at signals.c:286
        rec = 0x81593f8
[...]

PR #529, opened 2016-08-22
fix nick->host == NULL crash

ailin-nemui:
may resolve issue #500


dequis:
Welp, I give up, can't figure out how to reproduce this path. I got a lot of assertion warnings of missing hosts, but no idea how to get the first host to exist and the printnick one to be null.

Either way this does look like something that would fix the issue of the other ticket, so LGTM

ailin-nemui:
thanks for trying

Issue #538, opened 2016-09-09
/script reset unloads all scripts, but doesn't load autorun scripts


hintss:
Not sure if this is unintended behavior or not.

vague666:
Well the help says it unloads all scripts, it doesn't reload scripts, but you are welcome to write a PR and the developers can take it under consideration

ailin-nemui:
we believe this was fixed by #540

PR #540, opened 2016-09-11
/script reset can now also run the autorun scripts

LemonBoy:


dequis:
Ref #538 

The code is ok, not sure if the feature belongs there but that's probably just me being reluctant to add features, i'll just let nei decide if it's ok.

ailin-nemui:
@hintss can you test this

PR #542, opened 2016-09-13
Expose the CAP fields to the perl scripts.

LemonBoy:


ailin-nemui:
@dequis @grawity ?

dequis:
@ailin-nemui I would have pinged you, I'm not very familiar with these perl incantations.

But I guess you're just asking if it's okay to expose `cap_complete`, `cap_supported` and `cap_active`. Seems like a step in the right direction, sure.

Are those supposed to read-only?

LemonBoy:
Yes, you're supposed to use `cap_toggle` to add/remove some capabilities. But that's not exposed to the Perl side yet.

grawity:
Aside: When should scripts actually use `cap_toggle`? Will there be a signal after irssi receives CAP LS, and updates `cap_supported`, or after it receives CAPACK and updates `cap_active`?

LemonBoy:
If you call `cap_queue` before the connection then the caps are just queued and batch-requested when a `CAP LS` arrivare, otherwise they're requested on the spot.
Same goes for removals, it should be pretty much DWIM.

dequis:
Looks like you're missing an #include for the last commit

PR #556, opened 2016-10-04
update INSTALL requirement: terminfo or ncurses for text frontend

koollman:


ailin-nemui:
I would suggest changing it to `terminfo` only since we do not use curses

koollman:
correct, but the dependancy name is still ncurses for debian-like and redhat-like distributions (and it is the package recommended by ./configure)

ailin-nemui:
I think the dependency is libtinfo-dev for debian. still investigating this matter...

PR #557, opened 2016-10-16
TLS enhancements

ahf:
For some reason Github decided not to update the original PR at https://github.com/irssi/irssi/pull/311 when I force pushed to the `ahf/tls` branch.

This is a cleaned up version of the initial patch-set from https://github.com/irssi/irssi/pull/311 with a few things changed:
- We now support both pinning the x509 certificate and/or the subject public key information section of the certificate. This is especially useful for IRC networks that are adopting Lets Encrypt certificate where the public key remains the same for a long period, but the certificate is updated often.
- I've fixed the changes that @dequis requested.

Let's try to get this in for some proper testing.


dequis:
Fancy. The stuff in the commit log of 6ef36775fa6622a23bc2dd6b4fb10d0158f1189a should probably be included somewhere in the documentation.

(Haven't looked at everything in depth yet.)

ahf:
I didn't include it in NEWS (or elsewhere) since the output format of the `openssl dgst` command is wrong and cannot be put directly into the configuration files.

ahf:
Should we do like OpenSSH and use base64 instead of base16 when displaying the fingerprints?

src/core/tls.h
```diff
@@ -0,0 +1,98 @@
...
+	char *public_key_fingerprint_algorithm;
+	size_t public_key_size;
+
+	char *certificate_fingerprint;
+	char *certificate_fingerprint_algorithm;
+
+	char *not_after;
+	char *not_before;
+
+#ifdef SSL_get_server_tmp_key
```
dequis: Maybe it would be better to keep these fields in the struct even if SSL_get_server_tmp_key isn't defined

dequis:
> Should we do like OpenSSH and use base64 instead of base16 when displaying the fingerprints?

Keep it hex. ZNC does it the same way, and it's how openssl shows it.

dequis:
>I didn't include it in NEWS (or elsewhere) since the output format of the `openssl dgst` command is wrong and cannot be put directly into the configuration files.

How about these:

```
$ openssl s_client -connect chat.freenode.net:6697 < /dev/null 2>/dev/null | \
  openssl x509 > freenode.cert

$ openssl x509 -in freenode.cert -fingerprint -sha256 -noout 
SHA256 Fingerprint=CB:76:07:BB:D5:A4:9B:A5:9E:1F:BB:FD:BD:2B:85:33:20:C0:B5:AB:39:A1:20:CE:00:2D:8C:5E:AF:6D:30:DF

$ openssl x509 -in freenode.cert -pubkey -noout | \
  openssl pkey -pubin -outform der | \
  openssl dgst -sha256 -c | \
  tr a-z A-Z
(STDIN)= C1:11:88:85:5C:A8:9A:61:31:5B:27:C9:1D:1D:F6:C1:42:D5:39:05:61:36:99:EF:85:12:B2:D9:FD:08:5D:AF
```

NEWS
```diff
@@ -46,6 +46,7 @@ v0.8.21-head 2016-xx-xx  The Irssi team <staff@irssi.org>
 	        openssl dgst -sha256 -c | \
 	        tr a-z A-Z
 
+	+ Remove support for DANE validation of TLS certificates.
```
dequis: Maybe add more stuff from the commit log to this line? (or maybe we should start linking to commits)

dequis:
Did... github just flag the comment i just made as outdated?

Anyway the DANE entry in NEWS should probably have a couple of sentences from the commit log:

>There wasn't enough support in the IRC community to push for this on the majority of bigger IRC networks. If you believe this should be reintroduced into irssi, then please come up with an implementation that does not rely on the libval library. It is causing a lot of troubles for our downstream maintainers.

Also, the commands in the log of the cert/pubkey pinning commit should be updated to match the NEWS stuff.

src/core/network-openssl.c
```diff
@@ -565,6 +496,192 @@ static GIOChannel *irssi_ssl_get_iochannel(GIOChannel *handle, int port, SERVER_
 	return gchan;
 }
 
+static void set_cipher_info(TLS_REC *tls, SSL *ssl)
+{
+	g_assert(tls != NULL);
+	g_assert(ssl != NULL);
```
dequis: `s/g_assert/g_return_if_fail/`

src/core/network-openssl.c
```diff
@@ -565,6 +496,192 @@ static GIOChannel *irssi_ssl_get_iochannel(GIOChannel *handle, int port, SERVER_
...
+	tls_rec_set_protocol_version(tls, SSL_get_version(ssl));
+
+	tls_rec_set_cipher(tls, SSL_CIPHER_get_name(SSL_get_current_cipher(ssl)));
+	tls_rec_set_cipher_size(tls, SSL_get_cipher_bits(ssl, NULL));
+}
+
+static void set_pubkey_info(TLS_REC *tls, X509 *cert, unsigned char *cert_fingerprint, size_t cert_fingerprint_size, unsigned char *public_key_fingerprint, size_t public_key_fingerprint_size)
+{
+	g_assert(tls != NULL);
+	g_assert(cert != NULL);
```
dequis: `s/g_assert/g_return_if_fail/`

src/core/network-openssl.c
```diff
@@ -565,6 +496,192 @@ static GIOChannel *irssi_ssl_get_iochannel(GIOChannel *handle, int port, SERVER_
...
+
+	g_free(cert_fingerprint_hex);
+	g_free(public_key_fingerprint_hex);
+	EVP_PKEY_free(pubkey);
+}
+
+static void set_peer_cert_chain_info(TLS_REC *tls, SSL *ssl)
+{
+	g_assert(tls != NULL);
+	g_assert(ssl != NULL);
```
dequis: `s/g_assert/g_return_if_fail/`

src/core/network-openssl.c
```diff
@@ -565,6 +496,192 @@ static GIOChannel *irssi_ssl_get_iochannel(GIOChannel *handle, int port, SERVER_
...
+		}
+
+		tls_rec_append_cert(tls, cert_rec);
+	}
+}
+
+static void set_server_temporary_key_info(TLS_REC *tls, SSL *ssl)
+{
+	g_assert(tls != NULL);
+	g_assert(ssl != NULL);
```
dequis: `s/g_assert/g_return_if_fail/`

ahf:
Cool. Everything should be OK here.

@ailin-nemui you get the last seal of approval and get to merge :-)

src/core/chat-commands.c
```diff
@@ -99,27 +99,27 @@ static SERVER_CONNECT_REC *get_server_connect(const char *data, int *plus_addr,
...
-		conn->use_ssl = TRUE;
+	if (g_hash_table_lookup(optlist, "tls") != NULL || g_hash_table_lookup(optlist, "ssl") != NULL)
+		conn->use_tls = TRUE;
+	if ((tmp = g_hash_table_lookup(optlist, "tls_cert")) != NULL || (tmp = g_hash_table_lookup(optlist, "ssl_cert")) != NULL)
+		conn->tls_cert = g_strdup(tmp);
+	if ((tmp = g_hash_table_lookup(optlist, "tls_pkey")) != NULL || (tmp = g_hash_table_lookup(optlist, "ssl_pkey")) != NULL)
+		conn->tls_pkey = g_strdup(tmp);
+	if ((tmp = g_hash_table_lookup(optlist, "tls_pass")) != NULL || (tmp = g_hash_table_lookup(optlist, "ssl_pass")) != NULL)
+		conn->tls_pass = g_strdup(tmp);
+	if (g_hash_table_lookup(optlist, "tls_verify") != NULL || g_hash_table_lookup(optlist, "ssl_verify") != NULL)
```
klopsi: So, if .irssi/config specifies ssl_verify (or tls_verify) = "no", it is != NULL and the connection requires verification...   This may cause confusion.  (caused me problems)

dequis: Please don't use the code review features of pull requests for support questions, use the irc channel or github issues instead.

Issue #568, opened 2016-10-27
Cosmetic bug: repeated nick when mode change same user several times


ldev:
Hello,

A minor cosmetic "bug" in 0.8.20 (20160914 1355) (Arch Linux).

IRC server sends the following mode +o message to Irssi:
`zomgbbq@11.22.33.44 MODE #channel +oooo bobkaare bobkaare bobkaare bobkaare`

IRSSI shows the nick repeated several times like this:
`-!- mode/#channel [+oooo bobkaare bobkaare bobkaare bobkaare] by zomgbbq, zomgbbq, zomgbbq`

// Jonas L

PR #569, opened 2016-10-27
Collapse the list of nicks when possible.

LemonBoy:
When a single user changes the mode to a bunch of different users we
don't really have to print out the nick every time.

Partially solves #568, gives nicer messages overall

### DIFF
--- a/src/core/misc.c
+++ b/src/core/misc.c
@@ -261,6 +261,23 @@ char *gslist_to_string(GSList *list, const char *delimiter)
 	return ret;
 }
 
+gboolean gslist_is_unique(GSList *list, GCompareFunc func)
+{
+	const char *prev;
+
+	if (list == NULL)
+		return TRUE;
+
+	prev = (const char *)list->data;
+
+	for (list = list->next; list != NULL; list = list->next) {
+		if (func(list->data, prev) != 0)
+			return FALSE;
+	}
+
+	return TRUE;
+}
+
 void hash_save_key(char *key, void *value, GSList **list)
 {
         *list = g_slist_append(*list, key);
--- a/src/fe-common/irc/fe-modes.c
+++ b/src/fe-common/irc/fe-modes.c
@@ -83,7 +83,10 @@ static void print_mode(MODE_REC *rec)
 
 	tmp = modes; modes = NULL;
 
-	nicks = gslist_to_string(rec->nicks, ", ");
+	if (gslist_is_unique(rec->nicks, (GCompareFunc)g_strcmp0))
+		nicks = g_strdup(rec->nicks->data);
+	else
+		nicks = gslist_to_string(rec->nicks, ", ");
 	printformat(rec->channel->server, rec->channel->visible_name,
 		    MSGLEVEL_MODES, IRCTXT_CHANMODE_CHANGE,
 		    rec->channel->visible_name, rec->mode, nicks, "");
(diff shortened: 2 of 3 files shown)


src/core/misc.c
```diff
@@ -261,6 +261,26 @@ char *gslist_to_string(GSList *list, const char *delimiter)
...
+	if (list == NULL || list->next == NULL)
+		return TRUE;
+
+	prev = (const char *)list->data;
+	list = list->next;
+
+	while (list != NULL) {
+		if (strcmp(list->data, prev) != 0)
+			return FALSE;
+		prev = (const char *)list->data;
```
ahf: Since this line can only be called if `list->data` is equal to `prev`, then this operation will simply set `prev` to the same to a pointer to a new string, which contains the same data.

src/core/misc.c
```diff
@@ -261,6 +261,26 @@ char *gslist_to_string(GSList *list, const char *delimiter)
 	return ret;
 }
 
+gboolean gslist_is_unique(GSList *list)
+{
+	const char *prev;
```
ahf: Why is this `const`? You should be able to remove the constness and avoid doing the type-casts below.

src/core/misc.c
```diff
@@ -261,6 +261,26 @@ char *gslist_to_string(GSList *list, const char *delimiter)
...
+	list = list->next;
+
+	while (list != NULL) {
+		if (strcmp(list->data, prev) != 0)
+			return FALSE;
+		prev = (const char *)list->data;
+		list = list->next;
+	}
+
+	return FALSE;
```
ahf: This function only returns `TRUE` iff our list is the singleton list or an empty list?

I think this function should be reconsidered, both the name and the implementation of it. It looks like the function was named initially to do one things, but was turned into something else.

If you want to make a function, that checks that all values in the input list is the same, then you should probably take the comparison function as an argument to the function.

I think this function will also (with the current implementation in mind) require that the input list is always sorted.

src/fe-common/irc/fe-modes.c
```diff
@@ -83,7 +83,10 @@ static void print_mode(MODE_REC *rec)
 
 	tmp = modes; modes = NULL;
 
-	nicks = gslist_to_string(rec->nicks, ", ");
+	if (gslist_is_unique(rec->nicks) != TRUE)
```
ahf: Just use `if (gslist_is_unique(rec->nicks))` and turn the branches around :-)

LemonBoy: AFAIK we prefer explicit checks.

src/core/misc.c
LemonBoy: This is only meant to collapse the nicks in the trivial case when the same user does more than a single `MODE` change, could probably change the name of the function if you suggest a better one.

src/core/misc.c
LemonBoy: Whoops, this happens when you whip a quick PR :)

src/core/misc.c
LemonBoy: Well, `prev` isn't being modified so it makes sense to have it `const`.

src/core/misc.c
ahf: `int unique_nicknames(const GSList *nicknames)` maybe?

src/core/misc.c
```diff
@@ -261,6 +261,23 @@ char *gslist_to_string(GSList *list, const char *delimiter)
...
+
+	prev = (const char *)list->data;
+
+	for (list = list->next; list != NULL; list = list->next) {
+		if (func(list->data, prev) != 0)
+			return FALSE;
+	}
+
+	return FALSE;
+}
```
ahf: This entire function can currently be simplified into `return list == NULL || list->next == NULL;`. I highly doubt this is correct.

LemonBoy: The trailing return should be a `return TRUE`, I should really stop committing stuff before bedtime

PR #574, opened 2016-11-08
add completion_empty_line setting

lotheac:
I've been bitten a few times by trying to paste text beginning with a literal tab into a channel, only to have that text go to a query instead. I realized I don't actually want completion to do anything on an empty line, so here's a patch for an option to disable empty line completions.

### DIFF
--- a/src/fe-common/core/chat-completion.c
+++ b/src/fe-common/core/chat-completion.c
@@ -48,7 +48,7 @@ static int keep_privates_count, keep_publics_count;
 static int completion_lowercase;
 static char *completion_char, *cmdchars;
 static GSList *global_lastmsgs;
-static int completion_auto, completion_strict;
+static int completion_auto, completion_strict, completion_empty_line;
 static int completion_match_case;
 
 #define SERVER_LAST_MSG_ADD(server, nick) \
@@ -669,6 +669,8 @@ static void sig_complete_word(GList **list, WINDOW_REC *window,
 		return;
 
 	if (*linestart == '\0' && *word == '\0') {
+		if (!completion_empty_line)
+			return;
 		/* pressed TAB at the start of line - add /MSG */
                 prefix = g_strdup_printf("%cmsg", *cmdchars);
 		*list = completion_msg(server, NULL, "", prefix);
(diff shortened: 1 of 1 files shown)


LemonBoy:
I like this!
But I wonder if we could approach the problem from another angle and make the completion configurable (thus allowing the user to disable it) like this:
```C
if (completion_linestart_cmd[0] == '0')
    return;
prefix = g_strdup_printf("%c%s", *cmdchars, completion_linestart_cmd);
```
I wonder if that might end up being useful for someone though.

dequis:
Shouldn't there be a `signal_stop();` before the return?

lotheac:
On Wed, Nov 09 2016 00:17:42 -0800, dx wrote:
> Shouldn't there be a `signal_stop();` before the return?

I'm not familiar enough with the irssi code to know that, but there
isn't one near line 669 either.

-- 
Lauri Tirkkonen | lotheac @ IRCnet

lotheac:
On Wed, Nov 09 2016 00:10:06 -0800, LemonBoy wrote:
> I like this!
> But I wonder if we could approach the problem from another angle and make the completion configurable (thus allowing the user to disable it) like this:
> ```C
> if (completion_linestart_cmd[0] == '0')
>     return;
> prefix = g_strdup_printf("%c%s", *cmdchars, completion_linestart_cmd);
> ```

I'm afraid I don't understand what you mean -- my patch does make it
configurable whether or not completion does anything on an empty line
(the default being to keep the current behavior). What exactly is
"completion_linestart_cmd" here?

-- 
Lauri Tirkkonen | lotheac @ IRCnet

lotheac:
On Wed, Nov 09 2016 10:19:41 +0200, Lauri Tirkkonen wrote:
> On Wed, Nov 09 2016 00:17:42 -0800, dx wrote:
> > Shouldn't there be a `signal_stop();` before the return?
> 
> I'm not familiar enough with the irssi code to know that, but there
> isn't one near line 669 either.

Actually based on the other returns in this function it looks like
signal_stop() is only called if *list is non-NULL, ie. if something was
found.

-- 
Lauri Tirkkonen | lotheac @ IRCnet

ahf:
I think this is a good idea.

I think that turning the completion command into something that is configurable for commands other than `/MSG` can be handled in an additional follow-up patch, if someone is up for working on it, but do not check if the first element of the `completion_linestart_cmd ` char array is `'0'` as proposed. Instead check for whether it is `'\0'`.

ailin-nemui:
anything wrong with this patch?

dequis:
Other than being completely untested, nothing.

PR #576, opened 2016-11-17
Minimal message-tags handling - just skip the @ part

dequis:
This adds support for the ircv3.2 message-tags wire format without
actually parsing any of the tags or adding any interface to process them
(other than the 'server incoming' signal).

The API for this stuff can be decided later.

-----

"""Formal proof of correctness""":

AFL test case: http://dump.dequis.org/JxPOI.txt

Dirty bitlbee patch to send tags in every line: http://dump.dequis.org/A5dME.txt

### DIFF
--- a/src/irc/core/irc.c
+++ b/src/irc/core/irc.c
@@ -318,7 +318,16 @@ static char *irc_parse_prefix(char *line, char **nick, char **address)
 
 	*nick = *address = NULL;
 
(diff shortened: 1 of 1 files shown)


ailin-nemui:
I would like to see actual use case implementation before proceeding here...

dequis:
I'll work on a full implementation later.

Issue #159, opened 2014-10-22
Freezing when receiving dcc


tiagoapimenta:
I'm running irssi `0.8.15 (20100403 1617):armhf` on a Raspberry Pi Model B with Raspbian 7 (wheezy) kernel `3.12.28+ armv6l` distribution.

Ever I request a DCC streaming after accept it the screen and keyboard freezes, I am unable to type any command, even <kbd>Meta</kbd> + <kbd>Arrow</kbd> doesn't respond, but I'm using it over `screen`, and <kbd>Ctrl</kbd> + <kbd>A</kbd> <kbd>#</kbd> does work properly, that way I can see the streaming file growing.

When the streaming finishes (it usually breaks because some kind of timeout) it unfreeze and all the previous typed commands runs at once. When the timeout isn't reached it's fine, but when it is, the connection is restarted either.

I realised it only happen on lightweight hardware. I wonder it could be fixed by multithreading or sort of asynchronous file I/O solution.

Questions: [U&L](http://unix.stackexchange.com/questions/163121/is-it-normal-to-irssi-freezes-when-receiving-dcc) [RPi](http://raspberrypi.stackexchange.com/questions/24012/is-it-normal-to-irssi-freezes-when-receiving-dcc)

**Edit**: Even though RPi uses a SD-card I'm writing the file on a USB HDD, it should support the streaming speed but it seems it doesn't.

uwe-schwarz:
I have the same problem with a VIA nano processor irssi 0.8.15 (Ubuntu 0.8.15-5ubuntu3). Happens mostly on DCC SENDs with more than 15MB/s.

giulange:
I had the same problem on my raspberry pi:
Linux rpi-giuliano 4.1.7+ #817 PREEMPT Sat Sep 19 15:25:36 BST 2015 armv6l GNU/Linux

irssi freezes both via screen window or not (anyway via ssh) as soon as the first xdcc request is submitted.
I use an external ntfs HDD where dcc downloads are stored.
But I remember the first time I used the irssi application in which it was slow but was working.
I also highlight the large cpu usage (30-40%) by mount.ntfs.

I wonder if there could be something to solve this issue/bug.

Giuliano

dequis:
Can you try with 0.8.18-beta1?

https://github.com/irssi/irssi/releases/tag/0.8.18-beta1

giulange:
[...]

PR #578, opened 2016-11-23
add a static buffer for dcc received data

ailin-nemui:
increased buffersize might make irssi freeze less / #159

### DIFF
--- a/src/irc/dcc/dcc-get.c
+++ b/src/irc/dcc/dcc-get.c
@@ -30,6 +30,8 @@
 #include "dcc-get.h"
 #include "dcc-send.h"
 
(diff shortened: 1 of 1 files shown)


src/irc/dcc/dcc-get.c
```diff
@@ -580,6 +588,7 @@ static void cmd_dcc_get(const char *data)
 
 void dcc_get_init(void)
 {
+	dcc_get_recv_buffer = NULL;
```
dequis: Could move this = NULL to the initialization (although i think being .bss stuff this isn't needed at all)

ailin-nemui: removed

PR #581, opened 2016-11-29
Enforce the is_node_list contract in lib-config setters.

LemonBoy:
An assertion failure is better than a segfault.
This complements #570

### DIFF
--- a/src/lib-config/set.c
+++ b/src/lib-config/set.c
@@ -82,6 +82,7 @@ void config_node_clear(CONFIG_REC *rec, CONFIG_NODE *node)
 void config_nodes_remove_all(CONFIG_REC *rec)
 {
 	g_return_if_fail(rec != NULL);
+	g_return_if_fail(is_node_list(rec->mainnode));
 
 	while (rec->mainnode->value != NULL)
 		config_node_remove(rec, rec->mainnode, ((GSList *) rec->mainnode->value)->data);
@@ -94,6 +95,7 @@ void config_node_set_str(CONFIG_REC *rec, CONFIG_NODE *parent, const char *key, 
 
 	g_return_if_fail(rec != NULL);
 	g_return_if_fail(parent != NULL);
+	g_return_if_fail(is_node_list(parent));
 
 	no_key = key == NULL;
 	node = no_key ? NULL : config_node_find(parent, key);


ailin-nemui:
imo the config file should be fixed up instead and g_critical be thrown, similar to https://github.com/irssi/irssi/blob/master/src/lib-config/set.c#L108
not exactly sure what needs to be taken care of since in this case we need to fix parent and mainnode

dequis:
>imo the config file should be fixed up

How can we fix something that wasn't intended to make sense? This thing is just the result of fuzzing, not actual user mistakes.

ailin-nemui:
That doesn't really matter, we have had users accidentally corrupt their config files quite similarly to the fuzz. By fix in this case I mean removing and overwriting the incorrect node with a fresh one of correct type. Otherwise settings would be permanently broken 

Am 29. November 2016 23:19:53 MEZ, schrieb dx <notifications@github.com>:
>>imo the config file should be fixed up
>
>How can we fix something that wasn't intended to make sense? This thing
>is just the result of fuzzing, not actual user mistakes.
>
>-- 
>You are receiving this because you commented.
>Reply to this email directly or view it on GitHub:
>https://github.com/irssi/irssi/pull/581#issuecomment-263718532

-- 
Diese Nachricht wurde von meinem Android-Mobiltelefon mit K-9 Mail gesendet.

dequis:
Fair enough.

So i tested with this config

```
chatnets{0 0
```

And applying both this PR and the other one results in these visible errors:

```
Error: Ignored errors in configuration file:
Error: /home/dx/test/irssi/irssi/con:1: Warning: missing '='
Error: /home/dx/test/irssi/irssi/con:1: Warning: missing '='
Error: /home/dx/test/irssi/irssi/con:2: Warning: missing ';'
Error: /home/dx/test/irssi/irssi/con:2: error: unexpected end of file, expected character '}'
Error:
```

And it saves to:

```
chatnets = { 0 = "0"; };
settings = {
  core = { real_name = "dequis"; user_name = "dx"; nick = "dx"; };
  "fe-text" = { actlist_sort = "refnum"; };
};
```

Which loads just fine afterwards.

So, merge both?

dequis:
>new commit

Well this is awkward.

No visible change in behavior compared to the configs in my previous comment. A breakpoint in line 102 (the g_critical() one) is never hit. It's possible that this PR might have nothing to do with the thing I tested.

LemonBoy:
The second commit tries to save the date by nuking the parent node and re-creating it in place as a list/block one. I'm not 100% sold on the usefulness of this approach since we're substituting a broken and syntactically invalid configuration with a now syntactically-valid but possibly still broken one.

eg. the code can't do much when the parent key is wrong

I'm using the following snippet as test vector
```
0 0
chatnets{0""
```

ailin-nemui:
it doesn't work properly for some reason (probably a thought error on my side)
starting with
```
0 0
chatnets{""0
```
turns into
```
0 = "0";
chatnets = {  = { type = "IRC"; }; };
```
on first run and into
```
0 = "0";
chatnets = { type = { type = "IRC"; }; };
```
on 2nd

LemonBoy:
(It does output the 2nd snippet for me, after a load + `/quit` cycle)
That's my point, it can't work, hence my suggestion to merge the 1st commit /and/ #570

Issue #583, opened 2016-12-11
Performance issues when closing lots of windows


dequis:
Ref #582 "Set a default irssi windows limit"

>Yesterday an attacker opened 1000+ private msg windows and my irssi completely froze for several minutes when I tried to close even one of these msg windows. When I tried to close many of them at once with `/window close 50 100` command irssi consumed 100% of CPU and didn't seem to recover even after one hour.

New ticket to fix that problem by optimizing the commands instead of limiting windows (which is probably a good idea too)

Made some tests with variations of this:

```
(echo 001 a; for i in $(seq 1000); do echo ':nick'$i' PRIVMSG a :a' $RANDOM; done) | pv -Wl -L 50 | nc -l -p 6667

/script exec use Time::HiRes qw {time}; my $start = time(); Irssi::command("window close 2 1001"); print time() - $start;
```

Possibly not the best way to test this, but i was aiming for something somewhat close to the real thing.

Ran inside valgrind (callgrind) with instrumentation disabled (for no particular reason), performance should be half than native hardware.

All times in seconds. [Data spreadsheet here](https://docs.google.com/spreadsheets/d/1NwMVdRErzKxbnTsyokqzh0PxAicBrfGtWTt-sy_1GSU/edit?usp=sharing)

## /window close with all the windows

100 to 1000 windows with steps of 100.

O(n) and O(n^2) reference lines are extrapolating from the time to close 100, where n means number of windows.

It's quadratic but the absolute times up to 1000 windows are pretty low, I think it's perfectly tolerable compared to the others. But 2000 takes a minute, 3000 takes two minutes, 4000 takes four minutes...

![](http://dump.dequis.org/Yn3eL.png)

## Closing the last 100 windows

100 to 1000 windows with steps of 100.

O(n) reference line is extrapolating from the difference between 100 and 200, where n means total number of windows.

This is pretty good. It takes 5 seconds to close the last 100 windows of 2000, 10 seconds on 4000 and 25 seconds on 10000, so if you're stuck with that many windows, doing this one is several times is the best way to get out of that mess (assuming we haven't fixed /window close to behave like that by default, anyway)

![](http://dump.dequis.org/3qto-.png)

## Leftover windows after range

Also known as "closing first X windows", but the windows after that hurt so much that I had to change the perspective.

Always starts from a total of 500 windows, but varying the number of windows that are not closed after the range. The number of windows that are closed decreases too but that doesn't seem to be significant

[...]

PR #585, opened 2016-12-12
g_sequence backing for window list

ailin-nemui:
improve performance for #583

### DIFF
--- a/src/fe-common/core/fe-windows.c
+++ b/src/fe-common/core/fe-windows.c
@@ -165,7 +165,7 @@ static void windows_pack(int removed_refnum)
 	iter = windows_seq_refnum_lookup(refnum);
 	if (iter == NULL) return;
 
-	for (; iter != end; refnum++, iter = g_sequence_iter_next(iter)) {
+	while (iter != end) {
 		window = g_sequence_get(iter);
 
 		if (window == NULL || window->sticky_refnum || window->refnum != refnum)
@@ -173,6 +173,9 @@ static void windows_pack(int removed_refnum)
 
 		window_set_refnum0(window, refnum - 1);
 		windows_seq_changed(iter);
+
+		refnum++;
+		iter = g_sequence_iter_next(iter);
 	}
 }
 
@@ -556,12 +559,10 @@ GSList *windows_get_sorted(void)
 	GSList *sorted;
 
         sorted = NULL;
-	for (iter = windows_seq_end(),
-	     begin = windows_seq_begin();
+	iter = windows_seq_end();
+	begin = windows_seq_begin();
 
-	     iter != begin;
-
-	     ) {
+	while (iter != begin) {
 		iter = g_sequence_iter_prev(iter);
 		WINDOW_REC *rec = g_sequence_get(iter);
 
--- a/src/fe-common/core/fe-windows.c
+++ b/src/fe-common/core/fe-windows.c
@@ -42,6 +42,12 @@ static int daytag;
 static int daycheck; /* 0 = don't check, 1 = time is 00:00, check,
                         2 = time is 00:00, already checked */
 
+static int window_refnum_lookup(WINDOW_REC *window, void *refnum_p)
+{
+	int refnum = GPOINTER_TO_INT(refnum_p);
+	return window->refnum == refnum ? 0 : window->refnum < refnum ? -1 : 1;
+}
+
 static GSequenceIter *windows_seq_begin(void)
 {
 	return g_sequence_get_begin_iter(windows_seq);
@@ -50,6 +56,42 @@ static GSequenceIter *windows_seq_begin(void)
 static GSequenceIter *windows_seq_end(void)
 {
 	return g_sequence_get_end_iter(windows_seq);
+}
+
+static GSequenceIter *windows_seq_insert(WINDOW_REC *rec)
+{
+	return g_sequence_insert_sorted(windows_seq, rec, (GCompareDataFunc)window_refnum_cmp, NULL);
+}
+
+static GSequenceIter *windows_seq_refnum_lookup(int refnum)
+{
+	return g_sequence_lookup(windows_seq, GINT_TO_POINTER(refnum), (GCompareDataFunc)window_refnum_lookup, NULL);
+}
+
+static void windows_seq_changed(GSequenceIter *iter)
+{
+	g_sequence_sort_changed(iter, (GCompareDataFunc)window_refnum_cmp, NULL);
+}
+
+static GSequenceIter *windows_seq_window_lookup(WINDOW_REC *rec)
+{
+	return g_sequence_lookup(windows_seq, rec, (GCompareDataFunc)window_refnum_cmp, NULL);
+}
+
+/* search to the numerically right iterator of refnum */
+static GSequenceIter *windows_seq_refnum_search_right(int refnum)
+{
+	return g_sequence_search(windows_seq, GINT_TO_POINTER(refnum), (GCompareDataFunc)window_refnum_lookup, NULL);
+}
+
+/* we want to find the numerically left iterator of refnum, so we
+   search the right of the previous refnum. but we need to figure out
+   the case where the iterator is already at the beginning, i.e
+   iter->refnum >= refnum */
+static GSequenceIter *windows_seq_refnum_search_left(int refnum)
+{
+	GSequenceIter *iter = windows_seq_refnum_search_right(refnum - 1);
+	return iter == windows_seq_begin() ? NULL : g_sequence_iter_prev(iter);
 }
 
 static int window_get_new_refnum(void)
@@ -73,11 +115,6 @@ static int window_get_new_refnum(void)
 	}
 
 	return refnum;
-}
-
-static GSequenceIter *windows_seq_insert(WINDOW_REC *rec)
-{
-	return g_sequence_insert_sorted(windows_seq, rec, (GCompareDataFunc)window_refnum_cmp, NULL);
 }
 
 WINDOW_REC *window_create(WI_ITEM_REC *item, int automatic)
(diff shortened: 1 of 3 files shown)


ailin-nemui:
doesn't work yet, still crashes ;-) remedy for #583

src/fe-common/core/fe-windows.c
```diff
@@ -84,30 +88,60 @@ WINDOW_REC *window_create(WI_ITEM_REC *item, int automatic)
...
 static void windows_pack(int removed_refnum)
 {
 	WINDOW_REC *window;
 	int refnum;
+	GSequenceIter *iter, *end;
+
+	refnum = removed_refnum+1;
+	end = g_sequence_get_end_iter(windows_seq);
+	iter = g_sequence_lookup(windows_seq, GINT_TO_POINTER(refnum), (GCompareDataFunc)window_refnum_lookup, NULL);
+	if (iter == NULL) iter = end;
```
LemonBoy: Just `return` here, the `for` won't be executed.

src/fe-common/core/fe-windows.c
```diff
@@ -84,30 +88,60 @@ WINDOW_REC *window_create(WI_ITEM_REC *item, int automatic)
...
    leave the windows behind it alone. */
 static void windows_pack(int removed_refnum)
 {
 	WINDOW_REC *window;
 	int refnum;
+	GSequenceIter *iter, *end;
+
+	refnum = removed_refnum+1;
+	end = g_sequence_get_end_iter(windows_seq);
+	iter = g_sequence_lookup(windows_seq, GINT_TO_POINTER(refnum), (GCompareDataFunc)window_refnum_lookup, NULL);
```
LemonBoy: There are many `lookup` operations with a somewhat constant set of parameters, let's use a macro or a `static inline` helper ?

src/fe-common/core/fe-windows.c
```diff
@@ -84,30 +88,60 @@ WINDOW_REC *window_create(WI_ITEM_REC *item, int automatic)
...
+	if (iter == NULL) iter = end;
 
-	for (refnum = removed_refnum+1;; refnum++) {
-		window = window_find_refnum(refnum);
-		if (window == NULL || window->sticky_refnum)
+	for (; iter != end; refnum++, iter = g_sequence_iter_next(iter)) {
+		window = g_sequence_get(iter);
+
+		if (window == NULL || window->sticky_refnum || window->refnum != refnum)
```
LemonBoy: Doesn't the last condition stop this loop when it reaches a hole ?

src/fe-common/core/fe-windows.c
```diff
@@ -84,30 +88,60 @@ WINDOW_REC *window_create(WI_ITEM_REC *item, int automatic)
...
 void window_destroy(WINDOW_REC *window)
 {
+	GSequenceIter *iter;
 	g_return_if_fail(window != NULL);
 
 	if (window->destroying) return;
 	window->destroying = TRUE;
 	windows = g_slist_remove(windows, window);
+	iter = g_sequence_lookup(windows_seq, window, (GCompareDataFunc)window_refnum_cmp, NULL);
+	if (iter != NULL) g_sequence_remove(iter);
```
LemonBoy: Perhaps an `assert(iter != NULL)` is better as it shouldn't happen.

src/fe-common/core/fe-windows.c
```diff
@@ -189,26 +223,32 @@ void window_change_server(WINDOW_REC *window, void *server)
...
+	GSequenceIter *oiter, *witer;
 	int old_refnum;
 
 	g_return_if_fail(window != NULL);
 	g_return_if_fail(refnum >= 1);
 	if (window->refnum == refnum) return;
 
-	for (tmp = windows; tmp != NULL; tmp = tmp->next) {
-		WINDOW_REC *rec = tmp->data;
+	oiter = g_sequence_lookup(windows_seq, GINT_TO_POINTER(refnum), (GCompareDataFunc)window_refnum_lookup, NULL);
```
LemonBoy: Please add some comments about what the `witer` and `oiter` are for, I guess they're the **W**indow iter and the **O**rigin (maybe **R**efnum **iter** is a clearer name).

src/fe-common/core/fe-windows.c
LemonBoy: And can `g_sequence_get` return `NULL`, if it does I guess something went wrong and an `assert` should be used instead.

src/fe-common/core/fe-windows.c
```diff
@@ -189,26 +223,32 @@ void window_change_server(WINDOW_REC *window, void *server)
...
 	}
 
 	old_refnum = window->refnum;
 	window->refnum = refnum;
 	signal_emit("window refnum changed", 2, window, GINT_TO_POINTER(old_refnum));
+
+	if (witer != NULL && oiter != NULL) {
+		g_sequence_swap(oiter, witer);
+	} else {
+		g_sequence_sort_changed(witer, (GCompareDataFunc)window_refnum_cmp, NULL);
```
LemonBoy: I guess `witer` cannot be `NULL`, `assert` ?

src/fe-common/core/fe-windows.c
```diff
@@ -346,13 +386,13 @@ WINDOW_REC *window_find_closest(void *server, const char *name, int level)
 
 WINDOW_REC *window_find_refnum(int refnum)
 {
-	GSList *tmp;
+	GSequenceIter *iter;
 
-	for (tmp = windows; tmp != NULL; tmp = tmp->next) {
-		WINDOW_REC *rec = tmp->data;
+	iter = g_sequence_lookup(windows_seq, GINT_TO_POINTER(refnum), (GCompareDataFunc)window_refnum_lookup, NULL);
```
LemonBoy: Common idiom.

src/fe-common/core/fe-windows.c
```diff
@@ -400,71 +440,112 @@ WINDOW_REC *window_find_item(SERVER_REC *server, const char *name)
...
 int window_refnum_prev(int refnum, int wrap)
 {
-	GSList *tmp;
-	int prev, max;
+	GSequenceIter *iter, *end;
 
-	max = prev = -1;
-	for (tmp = windows; tmp != NULL; tmp = tmp->next) {
-		WINDOW_REC *rec = tmp->data;
+	iter = g_sequence_search(windows_seq, GINT_TO_POINTER(refnum-1), (GCompareDataFunc)window_refnum_lookup, NULL);
```
LemonBoy: Bar the different traversal direction the `_prev` and `_next` variants look pretty similar, can you perhaps merge the two under a `window_refnum_direction` and then make `window_refnum_prev` and `window_refnum_next` dumb wrappers over that function ?

src/fe-common/core/fe-windows.c
```diff
@@ -400,71 +440,112 @@ WINDOW_REC *window_find_item(SERVER_REC *server, const char *name)
...
+	end = g_sequence_get_end_iter(windows_seq);
+
+	if (iter != NULL) {
+		WINDOW_REC *rec;
+		if (iter != end) {
```
LemonBoy: If I understand this correctly this is a trick to avoid doing the `lookup(refnum)` and `iter_prev` in the common case (is it a common case ?) where you jump to the next/prev window. If so it needs to be documented.

src/fe-common/core/fe-windows.c
```diff
@@ -400,71 +440,112 @@ WINDOW_REC *window_find_item(SERVER_REC *server, const char *name)
...
+		}
 
-		if (rec->refnum < refnum && (prev == -1 || rec->refnum > prev))
-			prev = rec->refnum;
-		if (wrap && (max == -1 || rec->refnum > max))
-			max = rec->refnum;
+		iter = g_sequence_iter_prev(iter);
+		rec = g_sequence_get(iter);
+
+		if (rec->refnum < refnum)
```
LemonBoy: Is this needed ?

src/fe-common/core/fe-windows.c
ailin-nemui: that's the point of this method, stop at holes

ailin-nemui:
I implemented some of your suggestions

src/fe-common/core/fe-windows.c
ailin-nemui: we cannot be sure that the window with the refnum even exists. in that case we should search for the nearest window. lookup will only locate exact matches

ailin-nemui:
I added your other suggestions except for the asserts which I think aren't needed here

src/fe-common/core/fe-windows.c
```diff
@@ -42,15 +42,25 @@ static int daytag;
 static int daycheck; /* 0 = don't check, 1 = time is 00:00, check,
                         2 = time is 00:00, already checked */
 
+static GSequenceIter *windows_seq_begin(void)
```
LemonBoy: If you could stash all the `static` helpers together at the top of the file it'd be tidier.

src/fe-common/core/fe-windows.c
```diff
@@ -42,15 +42,25 @@ static int daytag;
...
 static int window_get_new_refnum(void)
 {
 	WINDOW_REC *win;
 	GSequenceIter *iter, *end;
 	int refnum;
 
 	refnum = 1;
-	for (iter = g_sequence_get_begin_iter(windows_seq),
-	     end = g_sequence_get_end_iter(windows_seq);
+	for (iter = windows_seq_begin(),
```
LemonBoy: Minor stylistic nit, a 7-line `for` doesn't look that nice

src/fe-common/core/fe-windows.c
```diff
@@ -117,9 +142,9 @@ static void windows_pack(int removed_refnum)
 	GSequenceIter *iter, *end;
 
 	refnum = removed_refnum+1;
```
LemonBoy: Space around the operators (ditto for a few more places)

src/fe-common/core/fe-windows.c
```diff
@@ -438,35 +468,34 @@ WINDOW_REC *window_find_item(SERVER_REC *server, const char *name)
 	return window_item_window(item);
 }
 
+static GSequenceIter *windows_seq_refnum_search_right(int refnum)
```
LemonBoy: Please add a small comment about the different implementation of the `left` vs `right` cases for our future selves and/or other devs.

ailin-nemui:
do you like it better now

PR #586, opened 2016-12-15
Process the nick changes in queries before the PRIVMSG is handled.

LemonBoy:
The `event privmsg` handler in `irc-queries.c` kicks in too late and renames the query so the privmsg ends up in the wrong window.
Tested with a local ratbox instance, had a lengthy conversation between me and myself over two irssi instances.
PS: We might need something for the `CTCP` too.


ailin-nemui:
any idea about the query_change_address code path which may now not be executed?

LemonBoy:
I think we can just add that check too, `g_strcmp0(query->address, address)` should suffice if I'm not mistaken. I didn't add it right away because I couldn't find a quick way to test that path and ran out of time ¯\\\__(ツ)\__/¯

PR #589, opened 2016-12-15
"Native" OTR Support

ahf:
This is a slightly different PR than our usual, small, PR contributions. This PR imports the code from the irssi OTR plug-in made by David Goulet, Uli Meis and with contributions from the rest of the irssi-OTR community.

Ignore the `README.otr` file for now, those changes can come as we go and will just make the experience better. I'll zap this file before we eventually merge this.


ailin-nemui:
- maybe commit these fixes to the upstream module first (even if it is going to be substituted by this)
- set want_otr to module if otr lib is found (like perl)
- backport libotr into travis environment so we get compile coverage

I won't review the individual otr code but it's fair to assume that we can accept it as long as you can maintain it

ahf:
1. I'll hear if upstream is interested in that. I suspect not though, since this is pretty much their way out of maintaining it :-)
2. Hm. I'm not sure this is a good idea. From my days doing packages and architecture work with Gentoo we always highly disliked these "magical" dependencies that are auto-discovered. Auto-discovery means that if I, as a package maintainer, misses that it does auto-detection by default I might end up with a dependency that Irssi adds to itself, but which isn't "encoded" in my package description.
3. Agreed. Will fix.

dequis:
>backport libotr into travis environment so we get compile coverage

https://github.com/bitlbee/bitlbee/blob/3.5.1/.travis.yml#L9-L20

Hacky but travis-built packages don't get uploaded anywhere, so it doesn't matter. Also doesn't require sudo. You can skip the debian parts.

Nowadays you could just tell travis to use ubuntu 14.04, but i like keeping it in 12.04 to ensure old libraries don't break the build.

ailin-nemui:
@ahf please reopen once ready

PR #590, opened 2016-12-18
Minor cleanup in the highlighting signal.

LemonBoy:
¯\\\_(ツ)_/¯

### DIFF
--- a/src/fe-common/core/hilight-text.c
+++ b/src/fe-common/core/hilight-text.c
@@ -379,34 +379,29 @@ static void sig_print_text(TEXT_DEST_REC *dest, const char *text,
 		char *middle;
 		int pos, color_pos, color_len;
 
-		tmp = g_string_new(NULL);
-
 		/* start of the line */
 		pos = strip_real_length(text, hilight_start, NULL, NULL);
-		g_string_append(tmp, text);
-		g_string_truncate(tmp, pos);
+		tmp = g_string_new_len(text, pos);
 
 		/* color */
 		g_string_append(tmp, color);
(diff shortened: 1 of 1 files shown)


src/fe-common/core/hilight-text.c
```diff
@@ -379,34 +379,29 @@ static void sig_print_text(TEXT_DEST_REC *dest, const char *text,
 		char *middle;
 		int pos, color_pos, color_len;
 
-		tmp = g_string_new(NULL);
-
 		/* start of the line */
 		pos = strip_real_length(text, hilight_start, NULL, NULL);
-		g_string_append(tmp, text);
-		g_string_truncate(tmp, pos);
+		tmp = g_string_new_len(text, pos);
```
ailin-nemui: this does not feel save if strlen(text) < pos

LemonBoy: Is that even possible? I expect `pos` to be `<= strlen`

ailin-nemui: maybe it isn't... :)

dequis:
```
/* Return how many characters in `str' must be skipped before `len'
   characters of text is skipped. */
int strip_real_length(const char *str, int len,
		      int *last_color_pos, int *last_color_len)
{
	const char *start = str;
	/*...*/
	while (*str != '\0') {
		/*...*/
	}
	return (int) (str-start);
}
```

`strip_real_length` would need to read past the end of the string to return something bigger than its length.

PR #594, opened 2016-12-23
Make sure SASL was actually requested before failing.

LemonBoy:
Tested against Efnet and Freenode, 10/10 would test again

### DIFF
--- a/src/fe-common/irc/fe-sasl.c
+++ b/src/fe-common/irc/fe-sasl.c
@@ -1,7 +1,7 @@
 /*
     fe-sasl.c : irssi
 
-    Copyright (C) 2015 The Lemon Man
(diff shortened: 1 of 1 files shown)


src/fe-common/irc/fe-sasl.c
```diff
@@ -43,6 +44,7 @@ static void sig_cap_end(IRC_SERVER_REC *server)
 	/* The negotiation has now been terminated, if we didn't manage to
 	 * authenticate successfully with the server just disconnect. */
 	if (!server->sasl_success &&
+	    gslist_find_string(server->cap_active, "sasl") != NULL &&
```
dequis: How about `server->sasl_mechanism != SASL_MECHANISM_NONE`?

PR #602, opened 2017-01-06
make irssi --with-perl build with separate object directory

McDutchie:
irssi 1.0.0 will not build if Perl is enabled and a separate object code directory is used. The problem is a relative path to an internal Perl dependency in four `Makefile.PL.in` files. This commit fixes the problem by making these paths absolute using `@top_srcdir@`.


ailin-nemui:
Thanks

Issue #519, opened 2016-08-02
/help SERVER ADD -port and -! examples incorrect


johannes87:
To reproduce:

```
/network add hackint
/server add -ssl -network hackint -port 9999 irc.hackint.org
/connect hackint
```

Irssi then connects to irc.hackint.org on port 6667. Tested with irssi 0.8.19 on OS X 10.11.

dequis:
Quoting from #256:

>not saying it makes sense, but the reason for this is that `-port` is for modifying an entry and `<address> <port>` is for adding an entry. Example:
>
>`/server add -network net1 example.com 6697` # add server of network 1
>`/server add -network net2 example.com 6698` # add server of network 2
>`/server add -port 6667 example.com` # modify port of server in network **2**
>`/server add -port 6668 example.com 6697` # modify port of server in network **1**

johannes87:
Thanks. Still, the documentation seems to be wrong. It lists the following command in the examples:

`/SERVER` ADD -! -auto -host staff.irssi.org -port 6667 -4 -network Freenode -noproxy orwell.freenode.net`

This suggests that "-port" does set the initial port. At least the example should be changed to remove the "-port 6667" IMHO:

`/SERVER` ADD -! -auto -host staff.irssi.org -4 -network Freenode -noproxy orwell.freenode.net 6667`

ailin-nemui:
the help is totally wrong

PR #606, opened 2017-01-06
Fix syntax on /help SERVER example

or4n:
Fix syntax on /help SERVER resolve issue #519

### DIFF
--- a/docs/help/in/server.in
+++ b/docs/help/in/server.in
@@ -64,7 +64,7 @@
     /SERVER CONNECT chat.freenode.net
     /SERVER CONNECT +chat.freenode.net
     /SERVER ADD -network Freenode -noautosendcmd orwell.freenode.net
-    /SERVER ADD -! -auto -host staff.irssi.org -port 6667 -4 -network Freenode -noproxy orwell.freenode.net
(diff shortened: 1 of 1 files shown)


ailin-nemui:
thanks

dequis:
Actually that's valid syntax to modify the port of an existing server

dequis:
Eh nevermind this was mentioned in the other ticket. But I don't think this change clarifies that

or4n:
dequis, you're right. Maybe we should add "-port 6667" to /SERVER MODIFY example?

ailin-nemui:
we need to double check if irssi has a bug in not actually allowing to add multiple same named servers with different ports

iirc the syntax is supposed to be /server -modify newport servername oldport, but /server add only supports /server add servername port

maybe server add should also be fixed to accept -port?

or4n:
I tried to add server with same name but different port. Irssi just says "18:15 -!- Irssi: Server orwell.freenode.net saved" but still doesn't do anything. It doesn't modify old entry or add new one.

I think it would be most logical solution to fix server add and then revert this pr.

or4n:
@ailin-nemui I did some more tinkering around and now have branch where server add accept -port (https://github.com/or4n/irssi/tree/server-add-cmd). Do you think I should make PR from that?

PR #608, opened 2017-01-07
Follow g_strsplit with call to g_strfreev

tijko:


src/irc/core/modes.c
```diff
@@ -748,6 +748,7 @@ static char *get_nicks(IRC_SERVER_REC *server, WI_ITEM_REC *item,
 				signal_emit("error command", 1,
 					    GINT_TO_POINTER(CMDERR_NOT_GOOD_IDEA));
 				signal_stop();
+                                g_strfreev(matches);
```
LemonBoy: This would be better placed after the `g_string_free` so that it matches the final sequence and the allocation order, other than this small nit the patch looks good to me.

tijko:
I made the edit and squashed the commits into one

src/irc/core/modes.c
```diff
@@ -743,6 +743,7 @@ static char *get_nicks(IRC_SERVER_REC *server, WI_ITEM_REC *item,
 			    g_hash_table_lookup(optlist, "yes") == NULL) {
                                 /* too many matches */
 				g_string_free(str, TRUE);
+                                g_strfreev(matches);
```
dequis: While the indentation is a mess (yep, there's another wrong line in the context part of this diff), we try to avoid making the problem worse. Tabs instead of spaces please.

tijko: My fault, edited to tabs.  I usually make small commits on a project if I'm not sure of how they will be received.  You and @LemonBoy have been cool about this small off-by-1 change, thanks!  I have really enjoyed the project too :)

dequis: <3

PR #610, opened 2017-01-08
Add frontend for fuzzing

josephbisch:
Use the following configure command:

$ ./configure --with-fuzzer --with-fuzzer-lib=/path/to/libFuzzer.a \
      CC=clang CXX=clang++

Places an irssi-fuzz in src/fe-fuzz/ after build.

Also can specify SANFLAGS to override the chosen sanitizer flags
(defaults to "-g -fsanitize=address -fsanitize-coverage=trace-pc-guard").


src/fe-fuzz/module-formats.c
```diff
@@ -0,0 +1,82 @@
+/*
```
ailin-nemui: I wonder if we could #include the other formats instead of having top copy*paste?

configure.ac
```diff
@@ -611,6 +650,7 @@ echo
 
 echo "Building text frontend ........... : $want_textui"
 echo "Building irssi bot ............... : $want_irssibot"
+echo "Building irssi fuzzer ............ : $want_irssifuzzer"
```
ahf: I think this is a slightly internal thing, so I don't think we should "confuse" the average user by making them wonder what "irssi fuzzer" is if they give compiling irssi for themselves a try.

src/fe-fuzz/Makefile.am
```diff
@@ -0,0 +1,25 @@
+bin_PROGRAMS = irssi-fuzz
+
+# Force link with clang++ for libfuzzer support
+CCLD=clang++ $(CXXFLAGS)
```
ahf: Probably do a check if `clang++` is available if fuzzing is enabled in `configure.ac`?

src/fe-fuzz/irssi.c
```diff
@@ -0,0 +1,58 @@
...
+#include "modules-load.h"
+#include "levels.h"
+#include "../fe-text/module-formats.h" // need to explicitly grab from fe-text
+#include "themes.h"
+#include "core.h"
+#include "fe-common-core.h"
+#include "args.h"
+#include "printtext.h"
+
+#include <stdbool.h>
```
ahf: Is this needed? It's a C99 feature and I don't think we build irssi as a C99 application right now - we might, but I'm not sure.

josephbisch:
Okay, fixed nits other than clang++ check.

Let me know if I should squash commits.

ahf:
LGTM.

PR #612, opened 2017-01-09
Allow selection of what kind of activity targets to ignore

vague666:
<pre>
Syntax:
 *	 	Ignore activity in all windows
 #	 	Ignore activity in all channels
 @	 	Ignore activity in all queries
 =	 	Ignore activity in all dcc chats
 #chan|nick	Ignore activity in named target(channel, query, dcc chat)
 tag/*		Ignore all activity on network 'tag'
 tag/#		Ignore activity in all channels on network 'tag'
 tag/@		Ignore activity in all queries on network 'tag'
 tag/=		Ignore activity in all dcc chats on network 'tag'
 tag/#chan|nick	Ignore activity in named channel/query/dcc chat on network 'tag'
</pre>

### DIFF
--- a/src/core/misc.c
+++ b/src/core/misc.c
@@ -168,6 +168,23 @@ int strarray_find(char **array, const char *item)
 	return -1;
 }
 
+int strarray_find_prefix(char **array, const char *item)
+{
+	char **tmp;
+	int index;
+
+	g_return_val_if_fail(array != NULL, -1);
+	g_return_val_if_fail(item != NULL, -1);
+
+	index = 0;
+	for (tmp = array; *tmp != NULL; tmp++, index++) {
+		if (g_str_has_prefix(*tmp, item))
+			return index;
+	}
+
+	return -1;
+}
+
 GSList *gslist_find_string(GSList *list, const char *key)
 {
 	for (; list != NULL; list = list->next)
--- a/src/core/misc.c
+++ b/src/core/misc.c
@@ -178,7 +178,7 @@ gboolean strarray_find_prefix(char **array, const char *item)
 
 	index = 0;
 	for (tmp = array; *tmp != NULL; tmp++, index++) {
-		if (g_str_has_prefix(*tmp, item))
+		if (g_str_has_prefix(g_ascii_strdown(*tmp, -1), g_ascii_strdown(item, -1)))
 			return TRUE;
 	}
 
--- a/src/core/misc.c
+++ b/src/core/misc.c
@@ -168,7 +168,7 @@ int strarray_find(char **array, const char *item)
 	return -1;
 }
 
-gboolean strarray_find_prefix(char **array, const char *item)
+int strarray_find_prefix(char **array, const char *item)
 {
 	char **tmp;
 	int index;
@@ -176,14 +176,14 @@ gboolean strarray_find_prefix(char **array, const char *item)
 	g_return_val_if_fail(array != NULL, -1);
 	g_return_val_if_fail(item != NULL, -1);
 
-	item = g_ascii_strdown(item, -1);
 	index = 0;
 	for (tmp = array; *tmp != NULL; tmp++, index++) {
-		if (g_str_has_prefix(g_ascii_strdown(*tmp, -1), item))
-			return TRUE;
+		//if (g_str_has_prefix(g_ascii_strdown(*tmp, -1), item))
+		if (g_ascii_strncasecmp(*tmp, item, length(item)) == 0)
+			return index;
 	}
 
-	return FALSE;
+	return -1;
 }
 
 GSList *gslist_find_string(GSList *list, const char *key)
--- a/src/core/misc.c
+++ b/src/core/misc.c
@@ -168,7 +168,7 @@ int strarray_find(char **array, const char *item)
 	return -1;
 }
 
-int strarray_find_prefix(char **array, const char *item)
+gboolean strarray_find_prefix(char **array, const char *item)
 {
 	char **tmp;
 	int index;
@@ -179,10 +179,10 @@ int strarray_find_prefix(char **array, const char *item)
 	index = 0;
 	for (tmp = array; *tmp != NULL; tmp++, index++) {
 		if (g_str_has_prefix(*tmp, item))
-			return index;
+			return TRUE;
 	}
 
-	return -1;
+	return FALSE;
 }
 
 GSList *gslist_find_string(GSList *list, const char *key)
--- a/src/core/misc.c
+++ b/src/core/misc.c
@@ -176,9 +176,10 @@ gboolean strarray_find_prefix(char **array, const char *item)
 	g_return_val_if_fail(array != NULL, -1);
 	g_return_val_if_fail(item != NULL, -1);
 
+	item = g_ascii_strdown(item, -1);
 	index = 0;
 	for (tmp = array; *tmp != NULL; tmp++, index++) {
-		if (g_str_has_prefix(g_ascii_strdown(*tmp, -1), g_ascii_strdown(item, -1)))
+		if (g_str_has_prefix(g_ascii_strdown(*tmp, -1), item))
 			return TRUE;
 	}
 
--- a/src/core/misc.c
+++ b/src/core/misc.c
@@ -178,7 +178,6 @@ int strarray_find_prefix(char **array, const char *item)
 
 	index = 0;
 	for (tmp = array; *tmp != NULL; tmp++, index++) {
-		//if (g_str_has_prefix(g_ascii_strdown(*tmp, -1), item))
 		if (g_ascii_strncasecmp(*tmp, item, strlen(item)) == 0)
 			return index;
 	}
--- a/src/core/misc.c
+++ b/src/core/misc.c
@@ -179,7 +179,7 @@ int strarray_find_prefix(char **array, const char *item)
 	index = 0;
 	for (tmp = array; *tmp != NULL; tmp++, index++) {
 		//if (g_str_has_prefix(g_ascii_strdown(*tmp, -1), item))
-		if (g_ascii_strncasecmp(*tmp, item, length(item)) == 0)
+		if (g_ascii_strncasecmp(*tmp, item, strlen(item)) == 0)
 			return index;
 	}
 
--- a/src/fe-common/core/fe-common-core.c
+++ b/src/fe-common/core/fe-common-core.c
@@ -19,6 +19,7 @@
 */
 
 #include "module.h"
+#include "modules.h"
 #include "module-formats.h"
 #include "args.h"
 #include "misc.h"
@@ -462,25 +463,59 @@ void fe_common_core_finish_init(void)
 gboolean strarray_find_dest(char **array, const TEXT_DEST_REC *dest)
 {
 	g_return_val_if_fail(array != NULL, FALSE);
+	const char *type = module_find_id_str("WINDOW ITEM TYPE", dest->window->active->type);
 
-	if (strarray_find(array, "*") != -1)
+	if ((strarray_find(array, "*") != -1) || // we ignore all targets
+		(g_ascii_strcasecmp(type, "CHANNEL") == 0 && strarray_find(array, "#") != -1) || // we ignore all channels
+		(g_ascii_strcasecmp(type, "QUERY") == 0 && // Is this a query?
+		(g_str_has_prefix("=", dest->target) ? // is it a dcc chat?
+			strarray_find(array, "=") != -1 : // are we ignoring dcc chat?
+			strarray_find(array, "@") != -1)) || // are we ognoring regular queries?
+		(strarray_find(array, dest->target) != -1)) // we ignore all channels with specific name
 		return TRUE;
+	else if (dest->server_tag != NULL) {
+		char *prefix = g_strdup_printf("%s/", dest->server_tag);
+		if (strarray_find_prefix(array, prefix)) {
+			GSList *targets = NULL, *iterator = NULL;
+			gboolean found = FALSE;
 
-	if (strarray_find(array, dest->target) != -1)
-		return TRUE;
+			if (type != NULL) {
+				// create a list of types to look for
+				targets = g_slist_append(targets, g_strdup("*"));
+				if (g_ascii_strcasecmp(type, "CHANNEL") == 0) {
+					targets = g_slist_append(targets, g_strdup("#"));
+					targets = g_slist_append(targets, g_strdup(dest->target));
+				}
+				else if (g_ascii_strcasecmp(type, "QUERY") == 0) {
+					if (g_str_has_prefix("=", dest->target))
+						targets = g_slist_append(targets, g_strdup("="));
+					else
+						targets = g_slist_append(targets, g_strdup("@"));
+				}
 
-	if (dest->server_tag != NULL) {
-		char *tagtarget = g_strdup_printf("%s/%s", dest->server_tag, "*");
-		int ret = strarray_find(array, tagtarget);
-		g_free(tagtarget);
-		if (ret != -1)
-			return TRUE;
+				for (iterator = targets; iterator; iterator = iterator->next) {
+					char *tagtarget = g_strdup_printf("%s/%s", dest->server_tag, (char *) iterator->data);
+					int ret = strarray_find(array, tagtarget);
+		        	        g_free(tagtarget);
+                			if (ret != -1) {
+						found = TRUE;
+						break;
+					}
+				}
 
-		tagtarget = g_strdup_printf("%s/%s", dest->server_tag, dest->target);
-		ret = strarray_find(array, tagtarget);
-		g_free(tagtarget);
-		if (ret != -1)
-			return TRUE;
+
+			}
+
+			g_slist_foreach(targets, (GFunc)g_free, NULL);
+			g_slist_free(targets);
+			g_free(prefix);
+
+			return found;
+		}
+		else {
+			g_free(prefix);
+		}
 	}
+
 	return FALSE;
 }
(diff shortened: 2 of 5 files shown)


ailin-nemui:
don't like this very much, we shoulld think of some better way. It's like a quick hack :-)

vague666:
I'm not sure there is a better way. Anything else would probably involve major changes to the codebase as I see it.

LemonBoy:
Perhaps by modifying `activity_hide_targets` syntax to allow for some finer-grained control over the filtering (eg: `Freenode/*` matches everything but `Freenode/#` matches only channels and `Freenode/@` might just match the non-channels if this option makes sense), just my 2cents.

vague666:
That change wouldn't be too taxing, sure... it just needs more cowbell

src/fe-common/core/fe-common-core.c
```diff
@@ -470,16 +470,23 @@ gboolean strarray_find_dest(char **array, const TEXT_DEST_REC *dest)
 		return TRUE;
 
 	if (dest->server_tag != NULL) {
-		char *tagtarget = g_strdup_printf("%s/%s", dest->server_tag, "*");
+		char *tagtarget = g_strdup_printf("%s/%s", dest->server_tag, "#");
```
LemonBoy: All the possible `tagtarget` values are in the form `server_tag/{#,@,*,target}` so it might make sense to have a `strarray_find_prefix` function so that you find the first entry matching `server_tag/` (the first one is the winner in case of duplicates) and then parse the part after the slash.
(please drop the previous commit and rebase this against master)

vague666: You'll have to explain the git steps one by one :) Close the PR? revert? I do not know!

vague666:
I should also `make` before `git commit` and `git push`
I thought the change was so simple there was no chance for errors :P

src/fe-common/core/fe-common-core.c
LemonBoy: You can `git reset` to the commit that diverged from `master` and then cherry-pick the changes on a pristine copy of the code 🔨

vague666: Can you expand what you mean with strarray_find_prefix when used on server_tag?

LemonBoy: All the entries in the list for a given tag start with `tagname/` so it makes sense to do a prefix search on the list and then check whether the part after the slash is a `@`,`#`,`*` or a valid `target` name.
One more detail jumped to my mind, if the user sets the list to `Freenode/@,Freenode/#` should we respect both even though that's better expressed as `Freenode/*`?

src/fe-common/core/fe-common-core.c
```diff
@@ -462,25 +463,59 @@ void fe_common_core_finish_init(void)
...
+			strarray_find(array, "@") != -1)) || // are we ognoring regular queries?
+		(strarray_find(array, dest->target) != -1)) // we ignore all channels with specific name
 		return TRUE;
+	else if (dest->server_tag != NULL) {
+		char *prefix = g_strdup_printf("%s/", dest->server_tag);
+		if (strarray_find_prefix(array, prefix)) {
+			GSList *targets = NULL, *iterator = NULL;
+			gboolean found = FALSE;
+
+			if (type != NULL) {
```
LemonBoy: Is it possible for `type` to be `NULL`? If so when?
You didn't check for it being `NULL` above.

src/fe-common/core/fe-common-core.c
```diff
@@ -462,25 +463,59 @@ void fe_common_core_finish_init(void)
 gboolean strarray_find_dest(char **array, const TEXT_DEST_REC *dest)
 {
 	g_return_val_if_fail(array != NULL, FALSE);
-
-	if (strarray_find(array, "*") != -1)
+	const char *type = module_find_id_str("WINDOW ITEM TYPE", dest->window->active->type);
+
+	if ((strarray_find(array, "*") != -1) || // we ignore all targets
```
LemonBoy: Put the check for `*` in its own line please (and comments on the line above the one they refer to)

src/fe-common/core/fe-common-core.c
```diff
@@ -462,25 +463,59 @@ void fe_common_core_finish_init(void)
...
+		(g_ascii_strcasecmp(type, "CHANNEL") == 0 && strarray_find(array, "#") != -1) || // we ignore all channels
```
LemonBoy: Having a small helper function taking a `TEXT_DEST_REC` and returning an enum depending on whether it is a channel, a dcc query or a regular one would make the code easier to read and avoid many string comparisons.

src/fe-common/core/fe-common-core.c
```diff
@@ -462,25 +463,59 @@ void fe_common_core_finish_init(void)
...
+		(g_ascii_strcasecmp(type, "QUERY") == 0 && // Is this a query?
+		(g_str_has_prefix("=", dest->target) ? // is it a dcc chat?
+			strarray_find(array, "=") != -1 : // are we ignoring dcc chat?
+			strarray_find(array, "@") != -1)) || // are we ognoring regular queries?
+		(strarray_find(array, dest->target) != -1)) // we ignore all channels with specific name
 		return TRUE;
+	else if (dest->server_tag != NULL) {
+		char *prefix = g_strdup_printf("%s/", dest->server_tag);
+		if (strarray_find_prefix(array, prefix)) {
```
LemonBoy: Apply the early exit pattern here, if the prefix isn't there then we have no business to do (and avoid a whole level of indentation)

src/fe-common/core/fe-common-core.c
```diff
@@ -462,25 +463,59 @@ void fe_common_core_finish_init(void)
...
+					targets = g_slist_append(targets, g_strdup(dest->target));
+				}
+				else if (g_ascii_strcasecmp(type, "QUERY") == 0) {
+					if (g_str_has_prefix("=", dest->target))
+						targets = g_slist_append(targets, g_strdup("="));
+					else
+						targets = g_slist_append(targets, g_strdup("@"));
+				}
+
+				for (iterator = targets; iterator; iterator = iterator->next) {
```
LemonBoy: The whole `targets` list is quite wasteful as you just need 4 bits to represent the set of all the possible checks to do (`*`, `#`, `=`, `dest->target`)

src/fe-common/core/fe-common-core.c
vague666: It might be a left-over from an earlier attempt

src/fe-common/core/fe-common-core.c
vague666: I think it makes it easier to build the list for which types to check for

src/fe-common/core/window-items.h
```diff
@@ -3,6 +3,14 @@
 
 #include "fe-windows.h"
 
+typedef enum {
+	WITEM_TYPE_CHANNEL = 1,
+	WITEM_TYPE_QUERY = 2,
+	WITEM_TYPE_PRIVMSG = 4,
+	WITEM_TYPE_DCCCHAT = 8,
+	WITEM_TYPE_OTHER = 16
+} WindowType;
```
LemonBoy: `WI_TYPE` ?

src/fe-common/core/fe-common-core.c
```diff
@@ -462,25 +463,57 @@ void fe_common_core_finish_init(void)
...
+	// we ignore all channels
+	else if (type & WITEM_TYPE_CHANNEL && strarray_find(array, "#") != -1)
+		return TRUE;
+	// Is this a dcc chat?
+	else if (type & WITEM_TYPE_DCCCHAT && strarray_find(array, "=") != -1)
+		return TRUE;
+	// Is this a private query window?
+	else if (type & WITEM_TYPE_PRIVMSG && strarray_find(array, "@") != -1)
 		return TRUE;
+	else if (dest->server_tag != NULL) {
```
LemonBoy: If `server_tag` is `NULL` we can just exit

src/core/misc.c
```diff
@@ -168,6 +168,24 @@ int strarray_find(char **array, const char *item)
...
+	char **tmp;
+	int index;
+
+	g_return_val_if_fail(array != NULL, -1);
+	g_return_val_if_fail(item != NULL, -1);
+
+	item = g_ascii_strdown(item, -1);
+	index = 0;
+	for (tmp = array; *tmp != NULL; tmp++, index++) {
+		if (g_str_has_prefix(g_ascii_strdown(*tmp, -1), item))
```
LemonBoy: `g_ascii_strncasecmp`

src/core/misc.c
```diff
@@ -168,6 +168,24 @@ int strarray_find(char **array, const char *item)
 	return -1;
 }
 
+gboolean strarray_find_prefix(char **array, const char *item)
```
LemonBoy: Please make this match the prototype of `strarray_find`

src/fe-common/core/fe-common-core.c
```diff
@@ -462,25 +463,57 @@ void fe_common_core_finish_init(void)
 gboolean strarray_find_dest(char **array, const TEXT_DEST_REC *dest)
 {
 	g_return_val_if_fail(array != NULL, FALSE);
+	WindowType type = window_item_get_type(dest->window->active);
```
LemonBoy: Either make it `const` or split the declaration and the initializer.

src/fe-common/core/fe-common-core.c
```diff
@@ -462,25 +463,57 @@ void fe_common_core_finish_init(void)
...
+	else if (type & WI_TYPE_PRIVMSG && strarray_find(array, "@") != -1)
 		return TRUE;
 
-	if (dest->server_tag != NULL) {
-		char *tagtarget = g_strdup_printf("%s/%s", dest->server_tag, "*");
-		int ret = strarray_find(array, tagtarget);
-		g_free(tagtarget);
-		if (ret != -1)
-			return TRUE;
+	if (dest->server_tag == NULL)
```
LemonBoy: Use `!`

src/fe-common/core/fe-common-core.c
```diff
@@ -462,25 +463,57 @@ void fe_common_core_finish_init(void)
...
+
+	char *prefix = g_strdup_printf("%s/", dest->server_tag);
+	if (!strarray_find_prefix(array, prefix)) {
+		g_free(prefix);
+		return FALSE;
+	}
 
-		tagtarget = g_strdup_printf("%s/%s", dest->server_tag, dest->target);
-		ret = strarray_find(array, tagtarget);
+	GSList *targets = NULL, *iterator = NULL;
```
LemonBoy: Declarations at the top please

src/fe-common/core/fe-common-core.c
```diff
@@ -462,25 +463,57 @@ void fe_common_core_finish_init(void)
...
+		if (ret != -1) {
+			found = TRUE;
+			break;
+		}
 	}
-	return FALSE;
+
+	g_slist_foreach(targets, (GFunc)g_free, NULL);
+	g_slist_free(targets);
+	g_free(prefix);
```
LemonBoy: Get rid of `prefix` when it is not needed anymore

src/fe-common/core/fe-common-core.c
```diff
@@ -462,25 +463,57 @@ void fe_common_core_finish_init(void)
 gboolean strarray_find_dest(char **array, const TEXT_DEST_REC *dest)
 {
 	g_return_val_if_fail(array != NULL, FALSE);
+	WindowType type = window_item_get_type(dest->window->active);
 
+	// we ignore all targets
```
LemonBoy: If the array is empty then we could return early

src/fe-common/core/fe-common-core.c
```diff
@@ -461,26 +462,63 @@ void fe_common_core_finish_init(void)
...
+	else if (type & WI_TYPE_PRIVMSG && strarray_find(array, "@") != -1)
 		return TRUE;
 
-	if (dest->server_tag != NULL) {
-		char *tagtarget = g_strdup_printf("%s/%s", dest->server_tag, "*");
-		int ret = strarray_find(array, tagtarget);
-		g_free(tagtarget);
-		if (ret != -1)
-			return TRUE;
+	g_return_val_if_fail(dest->server_tag != NULL, FALSE);
```
LemonBoy: The `g_return` macros are only used at the top of the function to do some basic checks on the parameters, the explicit if block was fine.

src/fe-common/core/fe-common-core.c
```diff
@@ -461,26 +462,63 @@ void fe_common_core_finish_init(void)
...
+		g_free(prefix);
+		return FALSE;
+	}
+
+	g_free(prefix);
 
-		tagtarget = g_strdup_printf("%s/%s", dest->server_tag, dest->target);
-		ret = strarray_find(array, tagtarget);
+	// create a list of types to look for
+	targets = g_slist_append(targets, g_strdup("*"));
```
LemonBoy: There's no need to strdup anything since the string literals have a ∞ lifetime and `dest->target` is definitely alive when it's feed into printf never to be used again.

src/fe-common/core/fe-common-core.c
vague666: I guess I misunderstood your 'use !' comment

src/fe-common/core/window-items.c
```diff
@@ -138,6 +138,23 @@ void window_item_set_active(WINDOW_REC *window, WI_ITEM_REC *item)
 	}
 }
 
+WindowType window_item_get_type(WI_ITEM_REC *item)
+{
+        g_return_val_if_fail(item != NULL, WI_TYPE_OTHER);
+
+	const char *type = module_find_id_str("WINDOW ITEM TYPE", item->type);
```
LemonBoy: Alignment

src/fe-common/core/window-items.h
```diff
@@ -3,6 +3,14 @@
 
 #include "fe-windows.h"
 
+typedef enum {
+	WI_TYPE_CHANNEL = 1,
+	WI_TYPE_QUERY = 2,
+	WI_TYPE_PRIVMSG = 4,
+	WI_TYPE_DCCCHAT = 8,
+	WI_TYPE_OTHER = 16
+} WindowType;
```
LemonBoy: `WI_TYPE` maybe ?

src/core/misc.c
```diff
@@ -168,6 +168,24 @@ int strarray_find(char **array, const char *item)
...
+{
+	char **tmp;
+	int index;
+
+	g_return_val_if_fail(array != NULL, -1);
+	g_return_val_if_fail(item != NULL, -1);
+
+	index = 0;
+	for (tmp = array; *tmp != NULL; tmp++, index++) {
+		//if (g_str_has_prefix(g_ascii_strdown(*tmp, -1), item))
```
LemonBoy: Remove

src/fe-common/core/window-items.h
vague666: I pulled that name from SettingType

src/core/misc.c
vague666: Ah yes, forgot :)

src/fe-common/core/fe-common-core.c
```diff
@@ -461,26 +462,63 @@ void fe_common_core_finish_init(void)
 
 gboolean strarray_find_dest(char **array, const TEXT_DEST_REC *dest)
 {
+	const WindowType type = dest->window->active ? window_item_get_type(dest->window->active) : WI_TYPE_OTHER;
```
ailin-nemui: remove the whole WindowType

ailin-nemui: use `item = window_item_find_window(dest->window, dest->server, dest->target)`

src/fe-common/core/fe-common-core.c
```diff
@@ -461,26 +462,63 @@ void fe_common_core_finish_init(void)
...
+	// we ignore all targets
 	if (strarray_find(array, "*") != -1)
 		return TRUE;
-
-	if (strarray_find(array, dest->target) != -1)
+	// exit if not a channel or query
+	else if (type & WI_TYPE_OTHER)
+		return FALSE;
+	// we ignore all channels
+	else if (type & WI_TYPE_CHANNEL && strarray_find(array, "#") != -1)
```
ailin-nemui: use
```
static int channel_type = module_get_uniq_id_str("WINDOW ITEM TYPE", "CHANNEL");
... if (item->type == channel_type && ...
```

src/fe-common/core/fe-common-core.c
```diff
@@ -461,26 +462,63 @@ void fe_common_core_finish_init(void)
...
+	gboolean found = FALSE;
+
 	g_return_val_if_fail(array != NULL, FALSE);
 
+	// we ignore all targets
 	if (strarray_find(array, "*") != -1)
 		return TRUE;
-
-	if (strarray_find(array, dest->target) != -1)
+	// exit if not a channel or query
```
ailin-nemui: instead of `strarray_find`,, `strarray_find_prefix` and GSLists, rewrite the code like this:

- loop **once** over array
- in each iteration, check if anything that needs to match, does match

ailin-nemui: example check:
```
if (item != NULL && item->type == channel_type && g_strcmp0(current_element, "#"))
```

src/fe-common/core/window-items.c
```diff
@@ -138,6 +138,21 @@ void window_item_set_active(WINDOW_REC *window, WI_ITEM_REC *item)
 	}
 }
 
+WindowType window_item_get_type(WI_ITEM_REC *item)
```
ailin-nemui: remove it

src/fe-common/core/window-items.h
ailin-nemui: remove it

LemonBoy:
Superseded by #779

PR #613, opened 2017-01-09
fix regression in completion

### LINKED ISSUE #609: Regression in tab completion of setting values (`/set <name> <space><tab>`)
Pointed out by vague. 1.0.0 bad, 0.8.20 good. Someone bisect this later.

ailin-nemui:
fixes #609


LemonBoy:
Looks right but in what case exactly the separator is a character different than a space?

ailin-nemui:
good question... I know that we should be able to complete eg /join #abc,#def,#ghi with , as the separator but I wonder if it matters in this branch (or if it gets it right even)

LemonBoy:
Good to know, I've tried that out and it seems to complete the `/j` with a comma-separated list of channels just fine with this patch.

PR #618, opened 2017-01-13
Print a warning if the server doesn't support SASL at all.

### LINKED ISSUE #616: Display reason why when quitting because of sasl_disconnect_on_failure
It is not always obvious when sasl_disconnect_on_failure disconnects you, like when a server doesn't support SASL at all (there is no SASL authentication failure message).

LemonBoy:
Fixes #616 (untested ofc)
And what if the the server doesn't support `CAP` at all?

### DIFF
--- a/src/fe-common/irc/fe-sasl.c
+++ b/src/fe-common/irc/fe-sasl.c
@@ -54,6 +54,11 @@ static void sig_cap_end(IRC_SERVER_REC *server)
 		 * halted and when the control goes back to irc_parse_incoming
 		 * the server object is safely destroyed. */
 		signal_stop();
+		/* If the server doesn't support the sasl cap then we print a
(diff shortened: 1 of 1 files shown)


dequis:
I'd prefer to print the error whenever `sasl_disconnect_on_failure` fails. Just add an else part to that if. Also would be nice to mention the `sasl_disconnect_on_failure` setting

>And what if the the server doesn't support CAP at all?

Hm, then this wouldn't trigger at all.

LemonBoy:
Either way it's fine with me, do you have any idea about the failure message? I can't seem to come up with a clear and concise one.

> Hm, then this wouldn't trigger at all.

Should we cover that case too?

ailin-nemui:
would be desirable imo

PR #621, opened 2017-01-16
Fix glib-2.0.m4 so that $PKG_CONFIG doesn't break configure script

josephbisch:
If $PKG_CONFIG contains a space, then the test may not work, so
surround with quotation mark characters. (for example,
PKG_CONFIG="pkg-config --static" breaks configure before this commit).


ailin-nemui:
if you can afford it please report upstream and refbug here

PR #622, opened 2017-01-16
provide net_start_ssl api

### LINKED ISSUE #615: Make irssi_ssl_get_iochannel accessible for plugins
The Quassel protocol doesn't start straight as TLS, but has a protocol-negociation before-hand.
Hence, quassel-irssi can't just use standard use_tls=1 of irssi (unless I've missed something, in which case, please tell!)

At the moment, to handle this, quassel-irssi has a copy/paste of irssi's network-openssl.c, which is obviously ugly, and isn't update-proof (it has been broken by irssi 1.0.0)

Would it be possible to make this symbol public?

ailin-nemui:
fixes #615

### DIFF
--- a/src/core/network-openssl.c
+++ b/src/core/network-openssl.c
@@ -20,6 +20,7 @@
 
 #include "module.h"
 #include "network.h"
+#include "net-sendbuffer.h"
 #include "misc.h"
 #include "servers.h"
 #include "signals.h"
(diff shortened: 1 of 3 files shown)


src/core/network-openssl.c
```diff
@@ -701,6 +702,22 @@ GIOChannel *net_connect_ip_ssl(IPADDR *ip, int port, IPADDR *my_ip, SERVER_REC *
...
+	GIOChannel *handle, *ssl_handle;
+
+	g_return_val_if_fail(server != NULL, NULL);
+
+	handle = net_sendbuffer_handle(server->handle);
+	if (handle == NULL)
+		return NULL;
+	ssl_handle  = irssi_ssl_get_iochannel(handle, server->connrec->port, server);
+	if (ssl_handle == NULL)
+		g_io_channel_unref(handle);
```
ailin-nemui: should the channel be unreffd here? I think probably not

LemonBoy: Nope, when the call above fails the original handle is left untouched since `g_io_channel_unix_get_fd` is the only function that operates on the `handle` parameter of `irssi_ssl_get_iochannel` before it reaches the function tail where it can't fail anymore.

PR #626, opened 2017-01-22
support storing and replaying the monospace attribute in textbuffer

ailin-nemui:
I want to support this formatting command (%#) also in terminal, even if it doesn't do anything. objections?


dequis:
Don't see why not, go ahead.

Looking forward to your implementations of `IRSSI_GUI_GTK`, `IRSSI_GUI_GNOME`, `IRSSI_GUI_QT` and `IRSSI_GUI_KDE` (which are four different things and not just two)

ailin-nemui:
lololol

PR #627, opened 2017-01-22
Check whether the client certificate is expired.

### LINKED ISSUE #211: Warn if the client certificate is expired
At the moment, using an expired client certificate to connect to an IRC server doesn't use the certificate (without a warning or anything) or uses it and the server ignores it. I've tested this on Rizon and I'm not sure which of the two it actually is, but I don't speak much C and OpenSSLs API, so I won't even try to make wild guesses :-) The only thing I've been able to figure out is that none of the checks in https://github.com/irssi/irssi/blob/8bd575df2ec0d4a17b6f59116d555b64f5bb1312/src/core/network-openssl.c#L487-L492 error out. It would be great if irssi would at least show a warning in this case or didn't try to establish a connection to the server at all.

LemonBoy:
Right now we only warn the user, the connection keeps going.
Fixes #211

### DIFF
--- a/src/core/network-openssl.c
+++ b/src/core/network-openssl.c
@@ -420,16 +420,38 @@ static GIOChannel *irssi_ssl_get_iochannel(GIOChannel *handle, int port, SERVER_
 
 	if (mycert && *mycert) {
 		char *scert = NULL, *spkey = NULL;
+		FILE *fp;
 		scert = convert_home(mycert);
 		if (mypkey && *mypkey)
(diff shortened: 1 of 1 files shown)


dequis:
Nice

ailin-nemui:
is this in any way affected by openssl 1.1?

ailin-nemui:
@LemonBoy

LemonBoy:
It does just work since the compat. patch has been merged

Issue #611, opened 2017-01-09
SASL broken due to ison fix


ailin-nemui:
#605 seems to break sasl. reported by athan and vague on #irssi

```
<< CAP LS
<< NICK vague
<< USER vague vague chat.freenode.net :vague
>> :verne.freenode.net NOTICE * :*** Looking up your hostname...
>> :verne.freenode.net NOTICE * :*** Checking Ident
>> :verne.freenode.net NOTICE * :*** Found your hostname
>> :verne.freenode.net NOTICE * :*** No Ident response
>> :verne.freenode.net CAP * LS :account-notify extended-join identify-msg multi-prefix sasl
<< CAP REQ :multi-prefix sasl
>> :verne.freenode.net 433 * vague :Nickname is already in use.
<< NICK vague_
>> :verne.freenode.net CAP * ACK :multi-prefix sasl 
<< AUTHENTICATE PLAIN
>> AUTHENTICATE +
<< AUTHENTICATE *
<< CAP END
>> :verne.freenode.net 906 vague_ :SASL authentication aborted
```

ailin-nemui:
@dequis

vague666:
I was able to connect if I skipped SASL auth and commenting out the section in #605 let me connect with SASL again

dequis:
well, fuck

LemonBoy:
Probably due to `wait_cmd` being off-sync ¯\\\_(ツ)_/¯

dequis:
#614 reverts it

Also, pretty cool to have real users testing things for us. Maybe we should get a test suite some day so we only bother users with the complex bugs that tests won't catch, not the trivial ones.

PR #647, opened 2017-02-11
notify-ison: Don't send ison before the connection is done

### LINKED ISSUE #596: ISON sent too early when using SASL
```
>> :bear.freenode.net CAP * LS :account-notify extended-join identify-msg multi-prefix sasl
<< CAP REQ :multi-prefix sasl
>> :bear.freenode.net 433 * vague :Nickname is already in use.
<< NICK vague_
>> :bear.freenode.net CAP * ACK :multi-prefix sasl
<< AUTHENTICATE PLAIN
<< ISON :vague
>> AUTHENTICATE +
<< AUTHENTICATE xxxxxxxxxxxxxxxxxxxxxxxxx
>> :bear.freenode.net 451 * :You have not registered
>> :bear.freenode.net 900 vague_ vague_!vague@xxx.xxx.xxx vague :You are now logged in as vague.
>> :bear.freenode.net 903 vague_ :SASL authentication successful
```

reported by @vague666

dequis:
Fixes #596 

Alternative fix to #605, this time much more boring and limited in scope.

If anything else relies on the same 120 second wait_cmd delay we can apply this same fix. Relying on the delay isn't great anyway.

### DIFF
--- a/src/irc/notifylist/notify-ison.c
+++ b/src/irc/notifylist/notify-ison.c
@@ -80,6 +80,10 @@ static void ison_send(IRC_SERVER_REC *server, GString *cmd)
 {
 	MODULE_SERVER_REC *mserver;
 
+	if (!server->connected) {
+		return;
+	}
+
 	mserver = MODULE_DATA(server);
 	mserver->ison_count++;
 


ailin-nemui:
> If anything else relies on the same 120 second wait_cmd delay we can apply this same fix. Relying on the delay isn't great anyway.

can you explain this a bit more detailed

dequis:
It's what I mentioned in https://github.com/irssi/irssi/issues/596#issuecomment-270929486

The fix based on that was #605 and trying to fix it based on that idea resulted in #611. Therefore, that idea is wrong and it's better to do the obvious thing instead.

Issue #651, opened 2017-02-15
Problem to parse options with captial letters with Irssi::command_parse_options


varazir:
Hello, 

I have this code in my script 

```perl
command_set_options('join_msg' => '+title +deviceId +deviceIds +deviceNames +url clipboard +smsnumber smstext +priority noencrypt +tasker text debug');

my ($join_args, $join_rest) = Irssi::command_parse_options('join_msg', $data);

ref $join_args or return 0
```

It doesn't return a error if the options has a uppercase like above. 
If I change the options to lowercase I get a error if the +options are missing a argument

IRSSI: Client: irssi 1.0.1 (20170203 2046)

data dump 
https://gist.github.com/varazir/03ac68bba36c197def700cbb66116fcc

LemonBoy:
I'm still not sure about what the problem is here, can you show how you're invoking `join_msg` and also what you're seeing and what you expect it to show?

varazir:
Close this, it's working as it should now.

LemonBoy:
Perhaps you're hitting the problem #652 solves where `/join_msg -deviceNames -text` shows a warning about the missing argument but `/join_msg -text -deviceNames` doesn't (and `deviceNames` is set to `''`) in the hash.

varazir:
I build a workaround in my script. checking that the argument isn't 0  length

PR #652, opened 2017-02-15
Properly check the command arguments in tail place.

LemonBoy:
A command requiring an argument and given in tail position would not
raise an error but silently set the value to the empty string ''.

Spotted by trying to figure out what #651 is about.

### DIFF
--- a/src/core/commands.c
+++ b/src/core/commands.c
@@ -567,13 +567,14 @@ static int get_cmd_options(char **data, int ignore_unknown,
 
 	option = NULL; pos = -1;
 	for (;;) {
-		if (**data == '-') {
+		if (**data == '\0' || **data == '-') {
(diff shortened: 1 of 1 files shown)


ailin-nemui:
mind boggling... good job but I still need to understand it

LemonBoy:
The code is a bit hairy but it boils down to this: say we've got `-foo bar -baz` where `foo` and `bar` require an argument, the code checks whether an argument has been provided every time it parses a command (every time it encounters a `-`) so when it reaches `-bar` it knows that the previous command has an argument and keeps parsing the string, eventually reaching the `\0`. You now see the problem: since there are no more `-` after `-baz` the code won't check the argument for the last command.

This can be noticed from the perl side of the moon as you'd get no error and spurious entries in the resulting hash.

PR #653, opened 2017-02-16
Enable UTF8 in GRegex

### LINKED ISSUE #636: GRegex not utf8 compliant?
Trying to match utf8 characters with a single "." in the lastlog fails. 
Example: having "RÖYH" in the lastlog, using "/lastlog -regexp R.YH" returns nothing.
Using "/lastlog -regexp R..YH" returns the relevant line.
Affects 1.0.0., works as expected in eg. 0.8.17-1+deb8u3.

ailin-nemui:
fixes #636 

up for discussion! this remaps invalid utf8 bytes into Private Use Area of unicode

### DIFF
--- a/src/core/iregex-gregex.c
+++ b/src/core/iregex-gregex.c
@@ -2,7 +2,7 @@
 
 #include "iregex.h"
 
-const gchar *
+static const gchar *
 make_valid_utf8(const gchar *text, gboolean *free_ret)
 {
 	GString *str;
@@ -82,7 +82,7 @@ i_regex_match (const Regex *regex,
 	return ret;
 }
 
-gsize
+static gsize
 strlen_pua_oddly(const char *str)
 {
 	const gchar *ptr;
--- a/src/core/iregex-gregex.c
+++ b/src/core/iregex-gregex.c
@@ -1,0 +1,137 @@
+#include <string.h>
+
+#include "iregex.h"
+
+const gchar *
+make_valid_utf8(const gchar *text, gboolean *free_ret)
+{
+	GString *str;
+	const gchar *ptr;
+	if (g_utf8_validate(text, -1, NULL)) {
+		if (free_ret)
+			*free_ret = FALSE;
+		return text;
+	}
+
+	str = g_string_sized_new(strlen(text) + 12);
+
+	ptr = text;
+	while (*ptr) {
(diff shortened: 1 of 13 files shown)


ailin-nemui:
also I didn't bother to write the regex.h implementation

ailin-nemui:
I intend to merge this since there was noone rushing forward to implement something else, and I am heavily in favour of keeping the platform independent gregex/pcre (thus disagree to revert that change)

dequis:
Sure, that bikeshedding wasn't going anywhere and this method is good enough.

It does need some review, I didn't look at this at all before. Also needs fuzzing.

LemonBoy:
Since people have shown renewed interest in running irssi on shoeboxes and similar underpowered hardware I think we should also think about the possible performance implications of "sanitizing" the input string every time we're going to run a match against it.

dequis:
>shoeboxes and similar underpowered hardware

Heheh.

I dunno, the performance issues we've fixed recently were O(n^2)-like things that broke for large values of n. This change looks linear on message length to me.

src/core/iregex-gregex.c
```diff
@@ -0,0 +1,137 @@
+#include <string.h>
+
+#include "iregex.h"
+
+const gchar *
```
dequis: static

src/core/iregex.h
```diff
@@ -0,0 +1,52 @@
...
+               GRegexMatchFlags match_options,
+               MatchInfo **match_info,
+               const gchar **new_string);
+
+gboolean
+i_match_info_fetch_pos (const MatchInfo *match_info,
+                        gint match_num,
+                        gint *start_pos,
+                        gint *end_pos,
+                        const gchar *new_string);
```
dequis: What is this new_string for? It seems to behave differently to the i_regex_match one, and one of the two implementations doesn't use it. I assume it's something related to the PUA stuff, but a comment would help

src/core/iregex-gregex.c
ailin-nemui: indeed

src/core/iregex.h
ailin-nemui: i_regex_match allocates the new_string if string is invalid utf8, and i_match makes used of it to do the reverse offset translation. I added a few comments but maybe we can get rid of new_string instead

PR #654, opened 2017-02-18
Do not expand the doubly-escaped sequences.

### LINKED ISSUE #520: /set expand_escapes ON doesn't handle double-escaping ("\\n")
`/set expand_escapes = ON` is great. 
It allows things like line-breaks to be embedded in regular messages (with `\n`) for instance.

The issue is that it currently cannot be double-escaped (`\\n`) to prevent expansion, as is the common convention. (i.e. `\\n` should render as simply `\n` and so on).

That is all. 
Cheers.

LemonBoy:
Closes #520.

### DIFF
--- a/src/core/misc.c
+++ b/src/core/misc.c
@@ -719,6 +719,13 @@ int expand_escape(const char **data)
 			}
 		}
 		return strtol(digit, NULL, 8);
+
+	case '\\':
+		/* escaped sequence, if the next character isn't part of a valid
+		 * sequence then we don't skip the leading slash. */
+		if (strchr("trnexc01234567", (*data)[1]) != NULL)
+			*data += 1;
+		return -1;
 	default:
 		return -1;
 	}


ailin-nemui:
I don\'t think this is a proper fix...

LemonBoy:
> I don't think this is a proper fix...

What constitutes a "proper" fix? What's wrong with this?
Was it really that **hard** to write something more than such a snarky remark?

ailin-nemui:
was it really hard to write some more details on your PR and how you believe it fixes the issue?

LemonBoy:
Said the one who has just pushed a PR whose only commit reads `make foreach send commands` and with such an informative message as `fixes /foreach query unquery [oops!]`.

This said, the PR simply outputs the `\\<char>` sequence as `\<char>` in the output string which is also explained in a comment I've added.

ailin-nemui:
I could not believe it at first, but I would say you're completely right. Thanks for the explanation. I only wonder now, if this would be better in `expand_escape` function. Probably doesn't matter, thoughts?

dequis:
>Was it really that hard to write something more than such a snarky remark?

Whoa whoa whoa, assume good faith.

I don't really understand why the `/* unknown escape, add it as-is */` block isn't doing what it's intended to do, or maybe a caller is messing with it. brb debugging

LemonBoy:
That makes sense if we want this behaviour for the other two or three places where `expand_escape` is called.

The `/* unknown escape, add it as-is */` block doesn't fire because of how the escape sequence gets chunked in the loop.

ailin-nemui:
yes, this was one of the reasons why I didn't understand this patch at first. The issue is not that it doesn't fire, it adds back `\\n` (instead of the "desired" `\n`)

ailin-nemui:
...The title would thus better be "do expand escaped escape sequences"

dequis:
I just did some testing, totally confused myself, wrote a very long comment with lots of test cases and then realized that bitlbee's command parser was messing with me and stripping slashes from commands. I shouldn't have used its control channel for testing.

Aaaaanyway this looks good to me.

>The issue is not that it doesn't fire, it adds back `\\n` (instead of the "desired" `\n`)

And this explains it perfectly.

So this PR adds an exception for a double slash, so that `\\n` can become `\n`, while `\a` can remain as `\a` using the `unknown escape` block. Makes sense, ship it.

src/core/misc.c
```diff
@@ -719,6 +719,13 @@ int expand_escape(const char **data)
 			}
 		}
 		return strtol(digit, NULL, 8);
+
+	case '\\':
+		/* escaped sequence, if the next character isn't part of a valid
```
ailin-nemui: why can't we just return "\\" here?

dequis:
Huh this looks completely different to what I reviewed before.

src/core/misc.c
LemonBoy: I decided to return `-1` because we didn't expand any escape in this case.

ailin-nemui: I mean instead of all that code, technically we should be expanding "\\\\" into "\\"

ailin-nemui:
on IRC I said something about "`\\k` should remain `\\k`," which leads to convoluted code here and also wasn't what I intended to say. I think it's fine if `\\` turns into `\` everywhere (I was thinking about `\k` should stay `\k` as it is currently and not become `k`). Thus the code can be simplified to `case '\\' : return '\\'` from what I can tell

PR #655, opened 2017-02-20
Tab completion for /format values.

LemonBoy:
I've tested this very briefly and it does seem to be consistent with how the completion for the other commands behaves.

### DIFF
--- a/src/fe-common/core/themes.c
+++ b/src/fe-common/core/themes.c
@@ -1263,7 +1263,7 @@ static void cmd_save(const char *data)
 	cmd_params_free(free_arg);
 }
 
-static void complete_format_list(THEME_SEARCH_REC *rec, const char *key, GList **list)
+static void complete_format_list(THEME_SEARCH_REC *rec, const char *key, const int get_value, GList **list)
 {
 	FORMAT_REC *formats;
 	int n, len;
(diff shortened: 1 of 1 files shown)


vague666:
Good work \m/ :)

src/fe-common/core/themes.c
```diff
@@ -1289,12 +1290,12 @@ static GList *completion_get_formats(const char *module, const char *key)
 	list = NULL;
 
 	modules = get_sorted_modules();
-	if (*module == '\0' || theme_search(modules, module) != NULL) {
+	if (module == NULL || theme_search(modules, module) != NULL) {
 		for (tmp = modules; tmp != NULL; tmp = tmp->next) {
 			THEME_SEARCH_REC *rec = tmp->data;
 
-			if (*module == '\0' || g_ascii_strcasecmp(rec->short_name, module) == 0)
```
ailin-nemui: question from ahf what motivates this change?

LemonBoy: I do not like throwing around an empty string when `NULL` could do just fine.

ailin-nemui:
`/format irc <TAB>` no longer completes to `netsplit`, `netsplit_more`, ... `away`...

LemonBoy:
The easy fix would be to check whether `irc` matches a module name but I'd prefer to change the format accepted by the command to a IMO saner `/format [<module>/]<name>`.

ailin-nemui:
the old code somehow did that.. but I really like having completion for format values, too! I am also afraid that some private script may be using /format module name, in general would be nice if we can keep the syntax imo.. although I do not feel very strongly

LemonBoy:
I don't care enough about this to keep the PR open, if anyone wants to take care of this should feel free to pick this up.

PR #658, opened 2017-02-21
Quote the filename when dcc requests are auto accepted.

### LINKED ISSUE #656: DCC autoget stopped working
When a DCC transfer is initiated by another party it does display the name and size of the file that is to be transferred, but then i always get "DCC no file offered by [name]". To make sure that the issue was not on the other end i used another client (pidgin) and it works just fine there.

If there is some debug information i could provide, let me know.

**EDIT:**
So the issue only appears to affect the autoget functionality if the file name contains spaces.
Version used: `1.0.1`

LemonBoy:
PR #453 forces the user to quote the filenames given to /DCC commands
when they contain spaces but the autoget functionality didn't get
updated so the filename was always passed without quotes.

Closes #656.

### DIFF
--- a/src/core/misc.c
+++ b/src/core/misc.c
@@ -724,18 +724,20 @@ int expand_escape(const char **data)
 	}
 }
 
-/* Escape all '"', "'" and '\' chars with '\' */
-char *escape_string(const char *str)
+/* Escape all the characters in `what' with a backslash */
+char *escape_string(const char *str, const char *what)
 {
-	char *ret, *p;
+	const char *p;
+	char *ret;
 
-	p = ret = g_malloc(strlen(str)*2+1);
-	while (*str != '\0') {
-		if (*str == '"' || *str == '\'' || *str == '\\')
-			*p++ = '\\';
-		*p++ = *str++;
+	ret = g_malloc(strlen(str) * 2 + 1);
+	for (p = str; *p != '\0'; p++, ret++) {
+		if (strchr(what, *p) != NULL) {
+			*ret++ = '\\';
+		}
+		*ret = *p;
 	}
-	*p = '\0';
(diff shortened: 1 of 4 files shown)


ailin-nemui:
what about filenames containing "

ailin-nemui:
how does the parser handle e.g \? does it need to be escaped as well?

dequis:
I don't really like this, I'd rather allow dcc get to accept unquoted args. (see https://github.com/irssi/irssi/issues/656#issuecomment-281400609)

ailin-nemui:
or we could further improve /dcc get to accept multiple files

dequis:
ehhhh

martinetd:
Hi!

Just reacting here because I was debugging the same issue just now, and found that however cmd_get_quoted_param works you didn't actually need to escape the internal quotes - e.g. '"foo "with quotes""' passed to it would return a proper 'foo "with quotes"' somehow.
The second patch in this PR now makes irssi segfault if I try to send a file with a quote in the filename -- that's rare enough, but still not exactly an expected behaviour :) (I get an invalid free in sig_dcc_request)

Should I open a new issue for it?
I'm not quite sure what's the best action to take now, quoting properly seems like an obvious thing to do but we also need to fix parsing in that case.

martinetd:
Ugh, actually that invalid free is just that escape_string doesn't return the right pointer, e.g. it returns the address for \0 at the end of the string in the memory it just allocated, and not the g_malloc address... So this won't work for any file regardless of wether escapement is ok or not.

I'll go with "ehhhh" too, will fix that + check if escaping works and if it does post a new PR with just that.

ailin-nemui:
thanks for hanging in here and sorry for merging the bad code

PR #659, opened 2017-02-27
make foreach send commands

ailin-nemui:
fixes /foreach query unquery [oops!]


dequis:
A bit copypastey, but I also looked into this before and didn't have any better ideas so sure.

ailin-nemui:
the problem would be that /foreach sends input instead of commands, so it's easy to mess up and spam all your windows/channels/queries

PR #661, opened 2017-03-01
Fix for #660

### LINKED ISSUE #660: Segfault when unloading a script that registers for 'script destroyed' signals
Script unloading is done in the wrong order:
1. Destroy the Perl package for the script
2. Emit 'signal destroyed' signal
3. Unhook signal handlers for the script

If a script adds a 'script destroyed' signal handler, unloading the script will cause step 2 to attempt to invoke the 'signal destroyed' handler in the script (because the signal handlers aren't yet removed).  However, because the Perl package has been destroyed, this will cause a script error, which will in turn attempt to unload the script (again). This will continue recursively until either stack or heap space is exhausted and the program segfaults.

To reproduce, simply load this script, then unload it:
````perl
use strict;
use Irssi qw/signal_add signal_register active_win/;
use Data::Dumper;

signal_register({ 'script error' => [ 'Irssi::Script', 'string' ] });
signal_register({ 'script created' => [ 'Irssi::Script' ] });
signal_register({ 'script destroyed' => [ 'Irssi::Script' ] });

sub scriptevent {
eval {
	active_win->command("SAY ScriptEvent: ($_[0]) $_[1]{package}");
};
}

signal_add('script error', sub { scriptevent('script error', $_[0]); });
signal_add('script created', sub { scriptevent('script created', $_[0]); });
signal_add('script destroyed', sub { scriptevent('script destroyed', $_[0]); });

````

Stevie-O:
This fixes #660

### DIFF
--- a/src/perl/perl-core.c
+++ b/src/perl/perl-core.c
@@ -67,10 +67,10 @@ static void perl_script_destroy(PERL_SCRIPT_REC *script)
 {
 	perl_scripts = g_slist_remove(perl_scripts, script);
 
-	signal_emit("script destroyed", 1, script);
-
 	perl_signal_remove_script(script);
 	perl_source_remove_script(script);
+
+	signal_emit("script destroyed", 1, script);
 
 	g_free(script->name);
 	g_free(script->package);


LemonBoy:
It makes sense to do so because the `script destroyed` signal is only used internally and it doesn't call into the Perl side. The only caveat is that the script won't see the signal at all (do we care about this?)

Stevie-O:
@LemonBoy I'm not sure what you mean when you say that it is 'only used internally and doesn't call into the Perl side', but you are right that the script itself can't get the 'script destroyed' signal, but it never could! By the time the `perl_script_destroy` function is called, pretty much everything the script needs in order to run has already been removed by `perl_script_destroy_package`.

I considered the idea of allowing a script to receive it's own destroy signal, but rejected it for the following reasons:
1. Such a thing would make it impossible for a user to unload a script without executing any code on it.
2. If a script wants a to have a 'safe unload' mechanism, the script may register a custom command that cleans up and then unloads itself; if the user wants to do a clean unload, they can just run that command instead.  (That should work fine; I think I'll test that next.)
3. If the 'script destroyed' handler causes a script error, we're right back where we started.
All in all, it does not seem like a good idea to try to automatically notify scripts that they're being unloaded.

ailin-nemui:
in general I would not expect scripts to do such mischievous things as registering this signal. It would be nice to have a safe scripting api but we do not at the moment

Issue #101, opened 2014-07-18
Calling die in 'script error' handler causes segfault


Hello71:
/script exec sub fail { Irssi::signal_remove('script error', 'fail'); die; }; Irssi::signal_add('script error', 'fail'); die

    #0  0x00000034a934cda1 in __strlen_sse2_pminub () from /lib64/libc.so.6
    #1  0x000000000048f65e in perl_script_destroy_package (script=0x5722d0) at perl-core.c:56
    #2  perl_script_unload (script=0x5722d0) at perl-core.c:310
    #3  0x000000000048f8b3 in sig_script_error (script=0x5722d0, error=<optimized out>) at perl-core.c:431
    #4  0x0000000000488bc2 in signal_emit_real (rec=rec@entry=0x551070, params=params@entry=2, va=va@entry=0x7fffffffc818, first_hook=<optimized out>) at signals.c:242
    #5  0x0000000000489035 in signal_emit (signal=signal@entry=0x49e1f4 "script error", params=params@entry=2) at signals.c:286
    #6  0x000000000048fbb5 in perl_script_eval (script=0x5722d0) at perl-core.c:244
    #7  script_load (name=<optimized out>, path=path@entry=0x0, data=<optimized out>) at perl-core.c:278
    #8  0x0000000000490163 in perl_script_load_data (data=<optimized out>) at perl-core.c:302
    #9  0x00000000004954ad in cmd_script_exec (data=<optimized out>) at perl-fe.c:56
    #10 0x0000000000488bc2 in signal_emit_real (rec=rec@entry=0x5a18d0, params=params@entry=3, va=va@entry=0x7fffffffc9f8, first_hook=<optimized out>) at signals.c:242
    #11 0x0000000000489035 in signal_emit (signal=signal@entry=0x58c2d0 "command script exec", params=params@entry=3) at signals.c:286
    #12 0x0000000000475da8 in command_runsub (cmd=0x49e848 "script", 
        data=0x58f30f "exec sub fail { Irssi::signal_remove('script error', 'fail'); die; }; Irssi::signal_add('script error', 'fail'); die;", server=0x0, item=0x0) at commands.c:329
    #13 0x0000000000488bc2 in signal_emit_real (rec=rec@entry=0x576540, params=params@entry=3, va=va@entry=0x7fffffffcba8, first_hook=<optimized out>) at signals.c:242
[...]

PR #662, opened 2017-03-02
Fix for segfault when a 'script error' handler dies

Stevie-O:
I have put together a well and true fix for #101. (I saw the comment in perl.txt and took it as a challenge :grin:) Not sure if it's the best solution, but it does work.

The fix is relatively straightforward and only touches four source files.
It adds two fields to `PERL_SCRIPT_REC`: `script_error_status` and `disable_signals`.

The overhead is negligible for the common case (i.e. when scripts aren't experiencing errors.)

`script_error_status` serves two purposes:
1. It is used to detect nested `script error` signals for the same script.  Barring certain *very* unlikely situations, such a situation indicates that a script's own `script error` handler has died.
2. It serves as a crude reference count on the script, to prevent both use-after-free errors and memory leaks.

`disable_signals` has only one purpose: To prevent `script error` signals from being delivered to a script.  It actually does more than that -- it prevents *any* signals from being delivered to the script.  I don't consider this much of a problem; unless the user has some *very* fancy scripts loaded, the only signal that would ever be delivered to a script with this flag set is `script error`.

For example, here's what it looks like on my screen when I run a variation of the example command from #101:

````
/script exec sub fail { print(__PACKAGE__ . "::fail('$_[1]')"); die 'goodbye, cruel world'; }; Irssi::signal_add('script error', 'fail'); die 'hasta la vista'

22:07 Irssi::Script::data1::fail('hasta la vista at (eval 317) line 1.
22:07 ')
22:07 -!- Irssi: Error in script data1:
22:07 goodbye, cruel world at (eval 317) line 1.
22:07
````

This is what happens behind the scenes:
1. A script error occurs (the `die 'hasta la vista'`)
2. `perl_report_script_error` (a new function) sets the script's `script_error_status` to `SES_ERROR`, then emits a `script error` signal.
3. The script's error handler prints the message about 'fail' being called (the first two output lines)
4. A second script error occurs (the `die 'goodbye, cruel world'`)
5. `perl_report_script_error` sees that we're already handling a script error.
    1. The `script error` signal that was emitted in step 2 is stopped
    2. The script's `disable_signals` flag is set to `TRUE`.
    3. A new `script error` signal is raised.  Note that our script will not receive that notification, because we shut its signals off.
    4. The sig_script_error handler in perl-fe.c displays the error notification (the last three output lines)
    5. The sig_script_error handler in perl-core.c unloads the script, but does *not* delete the `PERL_SCRIPT_REC` structure.  Instead, it sets the script's `script_error_status` to `SES_DESTROY`.
    6. The `script error` signal_emit call from step 2 completes and control is returned to `perl_report_script_error`.
6. `perl_report_script_error` sees that the script's `script_error_status` is now `SES_DESTROY`, and frees the `PERL_SCRIPT_REC` structure.

### DIFF
--- a/src/perl/perl-core.c
+++ b/src/perl/perl-core.c
@@ -35,6 +35,17 @@
 #include "XSUB.h"
 #include "irssi-core.pl.h"
 
+/* Values for script_error_status */
+
+/* No script error */
+#define SES_NO_ERROR 0
+/* Script error being emitted */
+#define SES_EMIT_ERROR 1
+/* Script destroyed; free memory */
+#define SES_DESTROY 2
+/* Used to help detect use-after-free */
+#define SES_DEAD 127
+
 extern char **environ;
 
 GSList *perl_scripts;
@@ -45,6 +56,8 @@ static char *perl_args[] = {"", "-e", "0", NULL};
 
 #define IS_PERL_SCRIPT(file) \
 	(strlen(file) > 3 && g_strcmp0(file+strlen(file)-3, ".pl") == 0)
+
+static void perl_script_free(PERL_SCRIPT_REC *script);
 
 static void perl_script_destroy_package(PERL_SCRIPT_REC *script)
 {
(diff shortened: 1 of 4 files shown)


Stevie-O:
TBH, you probably don't want to merge this (I've got a better solution in the works); however, I would greatly appreciate any feedback you guys have on my design.

ailin-nemui:
hi @Stevie-O   thanks for taking an interest in tackling this interesting issue, it seems much work for little gain. Also most scripts work fine without using this signal

PR #663, opened 2017-03-04
Use-after-free and other script stability fixes

Stevie-O:
This commit eliminates several use-after-free situations relating to
unloading scripts.

It also eliminates some segfaults stemming from recursive signals/commands.

### DIFF
--- a/src/fe-common/core/fe-settings.c
+++ b/src/fe-common/core/fe-settings.c
@@ -191,9 +191,9 @@ static void cmd_set(char *data)
 				/* Unpossible! */
 				break;
 			}
-			signal_emit("setup changed", 0);
 			printformat(NULL, NULL, MSGLEVEL_CLIENTCRAP, TXT_SET_TITLE, rec->section);
 			set_print(rec);
+			signal_emit("setup changed", 0);
 		} else
 			printformat(NULL, NULL, MSGLEVEL_CLIENTERROR, TXT_SET_UNKNOWN, key);
 	}
--- a/src/perl/perl-core.c
+++ b/src/perl/perl-core.c
@@ -35,6 +35,12 @@
 #include "XSUB.h"
 #include "irssi-core.pl.h"
 
+#ifdef TRACE_SCRIPT_UNLOADS
+#define SCRIPT_UNLOAD_DEBUG g_message
+#else
+#define SCRIPT_UNLOAD_DEBUG while(0)g_message
+#endif
+
 extern char **environ;
 
 GSList *perl_scripts;
@@ -71,6 +77,14 @@ static void perl_script_destroy(PERL_SCRIPT_REC *script)
 	perl_source_remove_script(script);
 
 	signal_emit("script destroyed", 1, script);
+
+	perl_script_unref(script);
+}
+
+
+static void perl_script_free(PERL_SCRIPT_REC *script)
+{
+	g_return_if_fail(script->refcount == 0);
 
 	g_free(script->name);
 	g_free(script->package);
@@ -250,7 +264,7 @@ static int perl_script_eval(PERL_SCRIPT_REC *script)
 
 		if (error != NULL) {
 			error = g_strdup(error);
-			signal_emit("script error", 2, script, error);
+			perl_script_error(script, error);
 			g_free(error);
 		}
 	}
@@ -258,7 +272,7 @@ static int perl_script_eval(PERL_SCRIPT_REC *script)
 	FREETMPS;
 	LEAVE;
 
-        return error == NULL;
+        return (!script->destroyed);
 }
 
 /* NOTE: name must not be free'd */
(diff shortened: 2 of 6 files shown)


Stevie-O:
This commit fixes three kinds of problems:

Use-after-free of `PERL_SCRIPT_REC` and `PERL_SIGNAL_REC`
--------------------

Consider a script, ignore_master, that manages ignore settings. It some custom commands registered, including one named `ignore_cloned`, and signal handlers for `ignore created`, `ignore destroyed`, and `ignore changed`.

You type `ignore_cloned` in order to ignore any *!user@host shared by 3 or more users.
This causes the 'command ignore_cloned' signal to be delivered to ignore_master (via `perl_call_signal` in src/perl/perl-signals.c).

The ignore_cloned command finds someone and adds them to the ignore list by executing the "ignore" command. This is processed by `cmd_ignore` in src/fe-common/core/fe-ignore.c, which does a lot of prep work before calling `ignore_add_rec` in src/core/ignore.c, which causes an `ignore created` signal to be emitted.

This causes the `ignore created` to be delivered to ignore_master (again, via `perl_call_signal` in src/perl/perl-signals.c.)

Now consider the situation when an error occurs in the `ignore created` signal handler.  Perl terminates the function call, leading to this code path:

````
                signal_emit("script error", 2, script, error);
                rec = NULL;
````
The `script error` signal is emitted, which will cause the script to be unloaded.  To be precise, it causes this:
1. The package namespace is wiped by `perl_script_destroy_package`. This means that none of the code or variables that the script loaded can be accessed through the package namespace.
2. With that, nearly all of the script is wiped. However, not all of it is gone -- the `ignore_cloned` command handler is still executing.  Perl ensures that that won't malfunction
3. All `PERL_SOURCE_REC`s and `PERL_SIGNAL_REC`s associated with the script are freed, as well as the `PERL_SCRIPT_REC` for the script.  This includes `rec`.

Because the `script error` signal unloaded the script, `rec` now points to freed memory. We set it to NULL so that the following code does not attempt to touch it.


Then, with the signal_emit finished, control is returned by `ignore_add_rec`, which then returns to its caller, which is the `ignore_cloned` command handler in the script we just unloaded.  What happens next depends on how the script is written.

#### Possibility 1: The script tries to call a function in its own package.

Because the package namespace was wiped by `perl_script_destroy_package`, this will fail with an 'Undefined subroutine' error.  This will then cause this to run:

````
                signal_emit("script error", 2, script, error);
````

However, `script` refers to memory that was freed by step 3 above.  If only the default handlers are registered, this will then dereference a whole lot of freed memory, culminating in a double-free bug.

#### Possibility 2: The script finishes executing without trying to call any functions in its own package.

In this instance, the call completes without error -- but `rec` still points to freed memory. And we then run the second half of `perl_call_signal` on that freed memory.


Protection from recursive scripts
--------------------

As part of some code that protects against integer wraparound in the reference counting code, the new code gets protection from this situation for free:

````
use strict;
use Irssi qw/command_bind command/;

command_bind('recursive-command', sub {
	command('recursive-command');
});
````

Under the old code, that would recurse until heap or stack space was exhausted, and then the program would segfault. 
Obviously, apart from contrived circumstances, this would only happen if someone made a mistake in their script. (Perhaps they meant to invoke a different command.).  As things go, "don't make mistakes or it will segfault" isn't a very user-friendly arrangement ;)
The new code will terminate the script well before that happens.

Protection from recursion in `script error`
--------------------

While the preceding also gives protecteion against `script error`, there is specific code included to detect recursion in `script error` events, and if that happens, the script's signal and command handlers are temporarily disabled.

**Comment by ailin-nemui** [comment]
thanks for your submission, it will take a while before I get to take a look at it

**Comment by ailin-nemui** [comment]
I found another nice use case which I believe is not covered by your patch,  thanks to @vague666 

```perl

use strict;
use warnings;

sub sig_setup_changed {
    die "xxx";
}

Irssi::settings_add_str('crash', 'change_this_to_crash', '');
Irssi::signal_add('setup changed', 'sig_setup_changed');
```
`/set change_this_to_crash boom`

vague666:
@ailin-nemui gee thanks, make my bad coding practices known to the world! :P

Stevie-O:
@ailin-nemui There, that fixes that:

````
16:32 -!-  ___           _
16:32 -!- |_ _|_ _ _____(_)
16:32 -!-  | || '_(_-<_-< |
16:32 -!- |___|_| /__/__/_|
16:32 -!- Irssi v1.0.1-1 - http://www.irssi.org
16:32 -!- Irssi: Loaded script vague666
16:32 [crash]
16:32 change_this_to_crash = boom
16:32 -!- Irssi: Error in script vague666:
16:32 xxx at /home/stevie/.irssi-dev/scripts/vague666.pl line 5.
16:32
````

ailin-nemui:
Thanks for your work! Note to self: some scripts try to force a value in a range when using /set, so they want to overwrite this in `setup changed`. Verify if this is broken by this fix

Stevie-O:
@ailin-nemui Do you have an example of such a script so I know how they would be doing it?

Stevie-O:
Okay, I dug up an example "pushover.pl" by using Github's source code search functionality, and created a simple test:

````perl
use strict;
use warnings;

use Irssi qw/settings_add_int settings_get_int settings_set_int signal_add/;

use constant { LESS_THAN_10 => "less_than_10" };

# test case based on pushover.pl
# https://github.com/hansnielsen/irssi-scripts/blob/108490bd3d2ad6a1b91ca58e5f866d01b64a9962/pushover.pl

sub sig_setup_changed {
	if (settings_get_int(LESS_THAN_10) >= 10) {
		settings_set_int(LESS_THAN_10, 9);
	}
}

settings_add_int('test', LESS_THAN_10, 5);
signal_add('setup changed', \&sig_setup_changed);
````

When I do `/set less_than_10 42` the output is:

````

ailin-nemui:
thanks for fixing the crash, in my opinion the old behaviour was nicer, immediately giving visual feedback that a value could not be changed. mhm

Stevie-O:
I do agree with you that the old behavior was nicer, and I'm currently working on a good way to fix it.

ailin-nemui:
I think it is basically good, maybe back out that last change that I asked for since you are working on an even better fix

ailin-nemui:
@Stevie-O still interested in doing anything here?

ailin-nemui:
closing for inactivity

PR #664, opened 2017-03-05
Fix dc99f8d7a of PR #658 (LemonBoy/dcc-autoaccept)

martinetd:
escape_string should return the alloc'ed address, not a pointer to the
NUL at the end of the string


martinetd:
Sorry used wrong commiter mail, sending again as I can't seem to figure how to update the very same pull req. Also softened commit message.

PR #665, opened 2017-03-05
Fix escape_string returning wrong address

martinetd:
escape_string should return the alloc'ed address, not a pointer to the
NUL at the end of the string

This fixes dc99f8d7a of PR #658 (LemonBoy/dcc-autoaccept)


ailin-nemui:
thanks, I reverted the broken code merge

PR #667, opened 2017-03-07
fix dcc get

### LINKED ISSUE #656: DCC autoget stopped working

ailin-nemui:
fixes #656

### DIFF
--- a/src/irc/dcc/dcc-autoget.c
+++ b/src/irc/dcc/dcc-autoget.c
@@ -23,6 +23,7 @@
 #include "masks.h"
 #include "settings.h"
 #include "servers.h"
+#include "misc.h"
 
 #include "dcc-get.h"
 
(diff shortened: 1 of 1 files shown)


ailin-nemui:
we should escape characters, this is a regression of consistently using quotes around file names

dequis:
Looks good

Odd but not necessarily bad behavior: if a filename doesn't have spaces but has double quotes, all of these are valid:

```
/dcc get bob asd"f
/dcc get bob "asd"f"
/dcc get bob 'asd"f'
/dcc get bob "asd\"f"
/dcc get bob 'asd\"f'
```

But this isn't:

```
/dcc get bob asd\"f
```

Also, only the first argument is taken. Given a filename with spaces, `asd f`, this is valid now:

```
/dcc get bob "asd f" garbage
```

`garbage` is just ignored. But this is invalid:

```
/dcc get bob asd f
```

(and it was previously valid which is the feature-regression i keep complaining about and you're probably sick of hearing about by now)

ailin-nemui:
thanks, let's do this (for now) then

PR #672, opened 2017-03-12
Update irssi.conf

xnite:
The address for IRCSource is wrong, so I've corrected it.


ailin-nemui:
The network this referred to originally was http://www.ircsource.net/servers.php

seems to be different from irc-source with dash. not sure if it still exists?

xnite:
Not sure, tried connecting and it seemed dead. I thought it was meant to
be a reference to my network since I couldn't find a network called
IRCSource on IRC-Source.com. If you don't mind I would still like it
added to the list but feel free to keep both. Officially the network is
BuddyIM (irc.buddy.im) but irc.irc-source.com points to it and we have
valid IV certificates for all the domains pointing to the network.

PR #676, opened 2017-03-20
fix initial screen draw (bdo#856201)

ailin-nemui:


ailin-nemui:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=856201

src/common.h
```diff
@@ -6,7 +6,7 @@
 #define IRSSI_GLOBAL_CONFIG "irssi.conf" /* config file name in /etc/ */
 #define IRSSI_HOME_CONFIG "config" /* config file name in ~/.irssi/ */
 
-#define IRSSI_ABI_VERSION 8
+#define IRSSI_ABI_VERSION 9
```
LemonBoy: Why does such a small change deserve an ABI bump?

src/fe-text/irssi.c
```diff
@@ -313,6 +316,7 @@ int main(int argc, char **argv)
 	g_log_set_always_fatal(loglev);
 	textui_finish_init();
 	main_loop = g_main_new(TRUE);
+	dirty_check(); /* draw the screen prior to the first wait poll */
```
LemonBoy: I'd move this at the end of `textui_finish_init` where the other `dirty_check` is.
The comment isn't very helpful as-is as it just states where this piece of code is, should be accompanied by some words about /why/ we're redrawing the screen.

src/common.h
ailin-nemui: it doesn't, I created the fork at the wrong time

ailin-nemui:
thanks for review, so far I don't understand why irssi would have behaved differently in older versions or what's going on at all? anyway actually drawing the screen does seem like the right thing to do before starting to wait on events

src/fe-text/irssi.c
ailin-nemui: I was considering that also but since we do dirty_check in the while loop it seems more appropriate here. another idea is to simply move the dirty_check before the poll inside the loop buut I didn't  want to evaluate the implications of that...?

dequis:
Is this something anyone has managed to reproduce, or just that guy in the debian ticket?

ailin-nemui:
I can actually confirm this -- start irssi with a fresh config and I have a small delay before I can see the whole ui

Stevie-O:
Can confirm that the change appears to be in glib; built/run 1.0.1 irssi against several Glib versions on Debian and the change appears to be between 2.49.2-2 (no delay visible at startup) and 2.49.3-1 (noticeable delay at startup).

I looked at the changelog for Glib between v2.49.2 to v2.49.3 and nothing jumps out at me as a possible cause for this; as soon as I figure out how, I'll bisect it. (Debian's build system is giving me grief trying to build v2.49.2 from a git clone.)

Stevie-O:
@ailin-nemui , @LemonBoy  Okay, I've tracked down the specifics of this issue:

In https://github.com/GNOME/glib/commit/e4ee3079c5afc3c1c3d2415f20c3e8605728f074 there were changes to three functions related to fd monitoring in order to reduce unnecessary wakeups: `g_source_set_ready_time`, `g_main_context_add_poll_unlocked`, and `g_main_context_remove_poll_unlocked`.  Previously, these functions would always cause a wakeup of the main loop; this commit changes them to only wake up the main thread if they're not being called from the main thread.

`g_main_context_add_poll_unlocked` is called unconditionally by `g_main_context_new`, which is the implementation for `g_main_new`, which is called by src/fe-text/irssi.c.  This means that the first call to `g_main_iteration` (src/fe-text/irssi.c line 327) would always complete immediately, causing an immediate 'reload_config' check followed by dirty_check.  Under the new glib logic, however, that doesn't happen anymore; so it waits until some sort of timer expires before it starts rendering the screen.

It seems to me that the most appropriate way fix this would be to rearrange the loop as follows:

````c
        while (!quitting) {
                if (reload_config) {
                        /* SIGHUP received, do /RELOAD */
                        reload_config = FALSE;
                        signal_emit("command reload", 1, "");
                }

                dirty_check();

                term_refresh_freeze();
                g_main_iteration(TRUE);
                term_refresh_thaw();
        }
````

ailin-nemui:
Sounds good to me, want to make the pr?

LemonBoy:
Thanks for the thorough investigation @Stevie-O, +1 for the proposed change.

Stevie-O:
Okay, doing one last sanity check before I make a PR.

Stevie-O:
There we go; the new PR is #680.

dequis:
We should probably open an issue on glib's side, at least to let them know. It's not a big deal for us but they may consider this a regression.

PR #677, opened 2017-03-20
OpenSSL support detection and documentation fixes

rrebello:
This commit series improves the detection of OpenSSL through the use of pkg-config in the configure script and removes an outdated information from INSTALL.


dequis:
Relevant 6300dfec71d376c96351708f76a6c4ee4a187eb5 @ahf

PR #678, opened 2017-03-20
Parse GEN_IPADD records in subjectAltName.

LemonBoy:
Needs testing, don't have time right now to do so.
Fixes [bdo#583544](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=583544)

### DIFF
--- a/src/core/network-openssl.c
+++ b/src/core/network-openssl.c
@@ -81,20 +81,57 @@ static const char *tls_dns_name(const GENERAL_NAME * gn)
 {
 	const char *dnsname;
 
-	/* We expect the OpenSSL library to construct GEN_DNS extension objects as
-	   ASN1_IA5STRING values. Check we got the right union member. */
-	if (ASN1_STRING_type(gn->d.ia5) != V_ASN1_IA5STRING) {
-		g_warning("Invalid ASN1 value type in subjectAltName");
-		return NULL;
-	}
+	if (gn->type == GEN_DNS) {
+		/* We expect the OpenSSL library to construct GEN_DNS extension objects as
+		   ASN1_IA5STRING values. Check we got the right union member. */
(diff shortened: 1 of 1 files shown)


dequis:
Crazy stuff. It sounds like the sort of thing openssl should be handling for us, and it really would be better to find a way to solve this with less code, not more.

dequis:
```c
static gboolean irssi_ssl_verify(SSL *ssl, SSL_CTX *ctx, const char* hostname, int port, X509 *cert, SERVER_REC *server, TLS_REC *tls_rec)
{
	long result;

	result = SSL_get_verify_result(ssl);
	if (result != X509_V_OK) {
		g_warning("Could not verify TLS servers certificate: %s", X509_verify_cert_error_string(result));
		return FALSE;
	} else if (! irssi_ssl_verify_hostname(cert, hostname)){
		return FALSE;
	}
	return TRUE;
}
```

Sounds like SSL_get_verify_result is OK but our extra checks are failing.

ahf:
We should ask ourselves if we really want this feature and, if we do want it, whether it should be protected by a flag somehow that is disabled by default.

dequis:
Are certificates issued to IP addresses something that is usually unsupported everywhere?

Hmm I wonder if this is the sort of thing where you could throw a CSR with fields the CA doesn't recognize, have it signed because the CA sucks, and pass validation

Also, from the ticket:

> 0: it's not from an irc server, but that doesn't matter here.
>   Also, details redacted since it's not available on the public internet
>   anyway.

So... no real use case? Just submitting bugs for the sake of annoying devs?

dequis:
Heh that makes my last comment sound bad. Context:

>16:50 < TheLemonMan> I'll close the openssl pr, we'd better to entirely offload the verification process to openssl
>16:50 < TheLemonMan> any objection?

PR #685, opened 2017-04-05
Fix strange history behavior when history is empty

### LINKED ISSUE #462: Window input history works strangely with window history cleared
to reproduce:
- join a new channel or issue /window history -clear
- enter text but don't press enter
- press arrow up
- return to the text you entered... you would assume you'd get to it by arrow down but you are now at 
  the start of the history buffer and have to press up to get to the text which is a bit confusing

not sure if this can be improved though, maybe arrow up should be 'stopped' if there's no more items in history?

josephbisch:
If text is being entered and then the user presses the up arrow
followed by the down arrow, the expected behavior is to return to the
text being entered. Prior to this commit that was not the case.

Fixes #462

### DIFF
--- a/src/fe-common/core/command-history.c
+++ b/src/fe-common/core/command-history.c
@@ -116,6 +116,8 @@ const char *command_history_prev(WINDOW_REC *window, const char *text)
 		history->pos = history->pos->prev;
 		if (history->pos == NULL)
                         history->over_counter++;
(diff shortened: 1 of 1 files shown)


LemonBoy:
IMO we should get rid of this counted-wraparound thing and just stop when we reach the top/bottom.

PR #686, opened 2017-04-07
Don't allow command history to wrap around

josephbisch:
This changes the behavior of the command history to avoid wrapping back
to the bottom once the top of the history is reached.

Tested of course.


LemonBoy:
We can get rid of `over_counter` now, don't we?

josephbisch:
Yes, I removed `over_counter` now and tested again.

PR #694, opened 2017-04-26
Accept -lcurses as well

coypoop:
NetBSD curses is called just 'curses' and works as well


ailin-nemui:
afaik we cannot be sure that curses without n makes use of terminfo

coypoop:
sorry for having variants of everything, someone in netbsd passionately hates ncurses

coypoop:
Thanks!

PR #695, opened 2017-04-26
Remove some dead assignments

coypoop:
I'm poking at it via clang static analyzer.

### DIFF
--- a/src/fe-common/core/fe-help.c
+++ b/src/fe-common/core/fe-help.c
@@ -232,7 +232,6 @@ static void show_help(const char *data)
 		if (!header) {
 			printtext(NULL, NULL, MSGLEVEL_CLIENTCRAP,
 				  "Irssi commands:");
-			header = TRUE;
 		}
 
 		if (last->category != NULL) {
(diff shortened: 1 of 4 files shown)


ailin-nemui:
thanks for getting involved, however please be wary of static analysis

src/irc/core/irc-servers-setup.c
```diff
@@ -138,8 +138,6 @@ static void init_userinfo(void)
 	if (nick == NULL || *nick == '\0') {
 		str = g_getenv("IRCNICK");
 		settings_set_str("nick", str != NULL ? str : user_name);
-
-		nick = settings_get_str("nick");
```
ailin-nemui: for example this statement here will resolve the precondition nick == NULL so that nick is guaranteed to be not NULL after this code block, in case of a future code addition which needs to make use of that variable

src/fe-common/core/fe-help.c
```diff
@@ -232,7 +232,6 @@ static void show_help(const char *data)
 		if (!header) {
 			printtext(NULL, NULL, MSGLEVEL_CLIENTCRAP,
 				  "Irssi commands:");
-			header = TRUE;
```
ailin-nemui: similarly here, if future code would need to display the header, we get the facts right by storing it

coypoop:
OK.

PR #696, opened 2017-04-26
Null check rec on process_destroy

coypoop:
rec=NULL may be a possible code path from handle_exec, if so then we might dereference null.
I am not sure if this is possible, but it looks like it may be.

Let me know if you disagree!

### DIFF
--- a/src/fe-common/core/fe-exec.c
+++ b/src/fe-common/core/fe-exec.c
@@ -175,6 +175,8 @@ static PROCESS_REC *process_find(const char *name, int verbose)
 
 static void process_destroy(PROCESS_REC *rec, int status)
 {
+	if (rec == NULL) return;
+
 	processes = g_slist_remove(processes, rec);
 
 	signal_emit("exec remove", 2, rec, GINT_TO_POINTER(status));


dequis:
```c
        /* options that require process ID/name as argument */
	if (rec == NULL &&
	    (signum != -1 || g_hash_table_lookup(optlist, "close") != NULL)) {
		printtext(NULL, NULL, MSGLEVEL_CLIENTERROR,
			  "Unknown process name: %s", args);
		return;
	}
	if (g_hash_table_lookup(optlist, "close") != NULL) {
		/* forcibly close the process */
                process_destroy(rec, -1);
                return;
	}
```

So we have:

a. `rec == NULL`
b. `signum != -1`
c. `g_hash_table_lookup(optlist, "close") != NULL`

And these conditions

1. `if (a && (b || c)) return;` (can be simplified to `if (a && c) return;`)
2. `if (c) return;`

Sounds like 2 passes if `!a` (`rec != NULL`)

Typical static analyzer false positive (it doesn't have a way to see both hash table lookups as the same thing)

coypoop:
oops, sorry for the noise

PR #697, opened 2017-05-01
sideways split support for Irssi

ailin-nemui:
warning:

* may be buggy

This commit adds support for sideways splits into Irssi. To that regard,
there are a number of new commands available

the "R" commands:

	/window new rsplit - make a new sideways split
	/window rshow - show an existing window to the right
	/window rgrow/rshrink/rsize/rbalance
	 - manipulate the size of your sideways split windows

the "D" commands:

	/window dup/ddown/dleft/dright
	 - navigate the windows directionally, as an alternative to
	   /window up/down that you can bind to some key
	/window move dleft/dright - the same for moving

Enjoy!


ailin-nemui:
Hi,

I am posting this here so it can be potentially discussed. The code is based on work done by LemonBoy in #431 

I won't have much time so I'm looking for other people who can help out

That said it seems to be working fine, but there are some general trade-offs implemented in this current approach which may need to be decided upon

ailin-nemui:
If you need some help debugging I have a commit in https://github.com/ailin-nemui/irssi/tree/rsplit-debug-code that will print the window sizes to stderr. With that patch applied, you can do `irssi 2>windowsizes.log` to inspect the screen layout during the resize progress. That may help in better understanding some issues visually

ailin-nemui:
I think it's mostly done

ailin-nemui:
one thing that has come up in discussions is the increase in bandwidth which is caused by sideways splits.

- Irssi updates visible buffers immediately, there is no delayed refresh implemented
- Irssi refuses to start on terminals without DECSTBM support
- For full width buffers, adding a line to a buffer requires ~line-length bytes
- So width×height bytes to fill the screen

For sideways splits we have a problem:

- Most terminals do not support DECSLRM :-(
- Thus, the whole split needs to be redrawn
- That requires ~width×height bytes - per line
- i.e width×height² bytes to fill the screen!

vvug:
This has been for years the most needed feature in irssi, especially to have a nicklist/chanlist bar on the side. Thanks for implementing it! I wouldn't care too much about the increased bandwidth: in a constrained environment one could chose not to use splits. Or you could enable the feature only in terminals that support DECSLRM.

timofonic:
@ahf @erdnaxeli  @ailin-nemui @fmmedeiros @josephbisch  @LemonBoy @trasz  @vague666 @vvug What's the status of this PR? Has it been splitted in other PRs? Can anyone explain it to us mere users? 

Things like things are what make me always been in doubt to consider choosing between @irssi and @WeeChat, each one has it's different strong and weak points. 

My ideal world would be massive code sharing and something like [mpv](@mpv-player) to happen, but I know it's a pipe dream. First there needs to exist a FFMpeg-like library for communications

dequis:
Didn't I tell you a while ago to stop pinging everyone? You're just bothering random people. Don't do that.

ailin-nemui:
@irssi/developers

PR #702, opened 2017-05-11
Added support for -notls and -notls_verify

vague666:


src/fe-common/core/fe-server.c
```diff
@@ -156,6 +156,10 @@ static void cmd_server_add_modify(const char *data, gboolean add)
 
 	if (g_hash_table_lookup(optlist, "tls") || g_hash_table_lookup(optlist, "ssl"))
 		rec->use_tls = TRUE;
+	else if (g_hash_table_lookup(optlist, "notls") || g_hash_table_lookup(optlist, "nossl")) {
```
ailin-nemui: please add { } around the `if` as well

ailin-nemui:
lgtm

ailin-nemui:
#117

src/fe-common/core/fe-server.c
```diff
@@ -154,8 +154,13 @@ static void cmd_server_add_modify(const char *data, gboolean add)
         else if (g_hash_table_lookup(optlist, "4"))
 		rec->family = AF_INET;
 
-	if (g_hash_table_lookup(optlist, "tls") || g_hash_table_lookup(optlist, "ssl"))
+	if (g_hash_table_lookup(optlist, "tls") || g_hash_table_lookup(optlist, "ssl")) {
 		rec->use_tls = TRUE;
+	}
+	else if (g_hash_table_lookup(optlist, "notls") || g_hash_table_lookup(optlist, "nossl")) {
+		rec->use_tls = FALSE;
+		rec->tls_verify = FALSE;
```
LemonBoy: Why does the `tls` option affect `tls_verify`? Afaics the latter is only used when the former is true

vague666: If the latter exists in rec it will set use_tls = TRUE later on in the code

LemonBoy: I see, better document this with a comment

src/fe-common/core/fe-server.c
```diff
@@ -159,6 +159,8 @@ static void cmd_server_add_modify(const char *data, gboolean add)
 	}
 	else if (g_hash_table_lookup(optlist, "notls") || g_hash_table_lookup(optlist, "nossl")) {
 		rec->use_tls = FALSE;
+		/* if rec has tls_verify = TRUE then use_tls will be set to true on lines 224-225
```
LemonBoy: Ehh, I guess something like `/* tls_verify implies use_tls, disable it explicitly */` would cut it better

vague666: It's there now! Next time say what you want me to write :P

LemonBoy: (You probably forgot to push)
You've committed the capital sin of embedding the LoCs, the comment is gonna get stale as soon as a single line is added :)
LGTM

PR #703, opened 2017-05-11
Allow -port <num> or irc.host.tld <num> in /server add 

vague666:
Also in /server modify
irc.host.tld &lt;num&gt; has preference if both are specified

### DIFF
--- a/src/fe-common/core/fe-server.c
+++ b/src/fe-common/core/fe-server.c
@@ -124,7 +124,8 @@ static void cmd_server_add_modify(const char *data, gboolean add)
 		port = atoi(portstr);
 	else if (value != NULL && *value != '\0')
 		port = atoi(value);
-	else if (g_hash_table_lookup(optlist, "tls"))
+	else if (g_hash_table_lookup(optlist, "tls") ||
+		 g_hash_table_lookup(optlist, "ssl"))
 		port = DEFAULT_SERVER_ADD_TLS_PORT;
 	else
 		port = DEFAULT_SERVER_ADD_PORT;
--- a/src/fe-common/core/fe-server.c
+++ b/src/fe-common/core/fe-server.c
@@ -120,7 +120,8 @@ static void cmd_server_add_modify(const char *data, gboolean add)
 
 	value = g_hash_table_lookup(optlist, "port");
 	port = *portstr == '\0' ?
-		(value != NULL && *value != '\0' ? atoi(value) : DEFAULT_SERVER_ADD_PORT)
+			(value != NULL && *value != '\0' ?
+			atoi(value) : DEFAULT_SERVER_ADD_PORT)
 		: atoi(portstr);
 
 	chatnet = g_hash_table_lookup(optlist, "network");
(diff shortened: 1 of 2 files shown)


src/fe-common/core/fe-server.c
```diff
@@ -117,7 +117,11 @@ static void cmd_server_add_modify(const char *data, gboolean add)
 		return;
 
 	if (*addr == '\0') cmd_param_error(CMDERR_NOT_ENOUGH_PARAMS);
-	port = *portstr == '\0' ? DEFAULT_SERVER_ADD_PORT : atoi(portstr);
+
+        value = g_hash_table_lookup(optlist, "port");
+        port = *portstr == '\0' ?
+                (value != NULL && *value != '\0' ? atoi(value) : DEFAULT_SERVER_ADD_PORT)
```
dequis: spaces instead of tabs

dequis:
Other than the indentation nit, sounds good.

ailin-nemui:
afaics this change will unconditionally modify the port even if you do not specify any port, can you double check?

dequis:
Hm yeah, so something like changing `DEFAULT_SERVER_ADD_PORT` into `rec ? rec->port : DEFAULT_SERVER_ADD_PORT`. But that line has so many ternaries it's better to turn them into proper if blocks at this point.

ailin-nemui:
can you change the ?:?: to a if, it's too unreadable for me

src/fe-common/core/fe-server.c
```diff
@@ -117,7 +117,17 @@ static void cmd_server_add_modify(const char *data, gboolean add)
...
 	if (*addr == '\0') cmd_param_error(CMDERR_NOT_ENOUGH_PARAMS);
-	port = *portstr == '\0' ? DEFAULT_SERVER_ADD_PORT : atoi(portstr);
+
+	value = g_hash_table_lookup(optlist, "port");
+
+	if (*portstr != '\0')
+		port = atoi(portstr);
+	else if (value != NULL && *value != '\0')
+		port = atoi(value);
+	else if (g_hash_table_lookup(optlist, "tls"))
```
ailin-nemui: does we need to add "ssl" also for backward compat?

ailin-nemui:
note to self: test the case of modifying a port and if irssi can handle multiple ports of the same host

vague666:
@ailin-nemui, how did that work? Was it possible to have several servers with the same name but different ports? That's something I haven't tested, but I've tested all scenarios I could think of when it comes to single servers

vague666:
There should be a test document that describes a set of operations that should be verified to work if something is changed. It's a big job to create such a document though

ailin-nemui:
@vague666 in the past the idea was that you can have the same named server, with different ports, eg. (A) irc.example.net 6660, (B) irc.example.net 6661, and so on, and then you could modify the port of (A) with /server add -port 6662 irc.example.net 6660

but idk if it ever worked (I currently think it didn't)

vague666:
I can't add the same server with another port in 1.0.2

PR #706, opened 2017-05-17
Add parse_uint function to improve integer overflow handling

dequis:
Originally found by oss-fuzz (issue 525) in `get_ansi_color` using ubsan. After a lot of analysis I'm 99% sure this isn't security relevant so it's fine to handle this publicly.

The fix is mainly adding a function that does it right and use it everywhere. This is harder than it seems because the `strtol()` family of functions doesn't have the friendliest of interfaces.

Aside from `get_ansi_color()`, there were other pieces of code that used the same `(out*10+(*in-'0'))` pattern, like the `parse_size()` and `parse_time_interval()` functions, which are mostly used for settings. Those are interesting cases, since they multiply the parsed number (resulting in more overflows) and they write to a signed integer parameter (which can accidentally make the uints negative without UB)

Thanks to Pascal Cuoq for enlightening me about the undefined behavior of `parse_size` (and, in particular, the implementation-defined behavior of one of the WIP versions of this commit, where something like signed integer overflow happened, but it was legal). Also for writing tis-interpreter, which is better than ubsan to verify these things.

### DIFF
--- a/src/core/misc.c
+++ b/src/core/misc.c
@@ -750,10 +750,42 @@ int nearest_power(int num)
 	return n;
 }
 
-int parse_time_interval(const char *time, int *msecs)
+/* Parses unsigned integers from strings with decent error checking.
+ * Returns true on success, false otherwise (overflow, no valid number, etc)
+ * There's a 31 bit limit so the output can be assigned to signed positive ints */
+int parse_uint(const char *nptr, char **endptr, int base, guint *number)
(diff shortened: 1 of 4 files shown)


ailin-nemui:
well good that you did this and I think it's great -- except I don't feel like I can review it with a certainty. just fyi. maybe just merge and yolo?

Issue #635, opened 2017-02-02
Means to restore a layout


madduck:
After I ran `/LAYOUT SAVE` and made a few changes to the windows, there is no way for me to restore the saved layout. `/LAYOUT RESET` apparently just removes the layout altogether. I can then `/RELOAD`, but that doesn't actually reapply the layout (i.e. doesn't move windows that have changed position, or recreate missing ones).

This is obviously a hard problem in ways, because what should we do with windows that aren't in the layout. By default, I'd say to just leave them untouched.

But maybe a `-force` option could be provided which would actually destroy unwanted windows (moving the window items to provisioned windows according to bounds, or destroying those items if they're not expected as part of the layout, or split them out into new windows (as `autocreate_*` would…).

Soundtoxin:
Accidentally closed my split window for hilights and was looking up how to restore my layout when I came across this. Ended up manually recreating the window, which I had to look up how to do. I would like it if this feature got added, just in case.

PR #708, opened 2017-05-18
Restore layout

LemonBoy:
⚠️ DO NOT MERGE UNTIL THIS LINE HAS BEEN REMOVED ⚠️ 

This addresses #635 once and forever.
Please give a look at the changes made to `sig_layout_restore` wrt the GUIs.

cc @madduck @Soundtoxin is this ok?

### DIFF
--- a/docs/help/in/layout.in
+++ b/docs/help/in/layout.in
@@ -7,6 +7,7 @@
 
     SAVE:     Saves your layout to the configuration.
     RESET:    Removes the saved layout from the configuration.
+    RESTORE:  Restores the saved layout from the configuration.
 
 %9Description:%9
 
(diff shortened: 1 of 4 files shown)


LemonBoy:
You're the one that knows the windowing system the most, if you feel this is fine then smash that merge button!

ailin-nemui:
I don't think it will really "restore" the layout, right? only recreate missing refnums

LemonBoy:
The code that restores the layout was meant to be run without any window open, all this patch does is recreate the missing windows and then hands them all to the code that shuffles them around.

ailin-nemui:
that's why I was thinking we should forcibly "destroy" all the main windows and then run "the code" (and also run it -- in parts -- again on existing windows, to put them in the right location on screen)

would be more complex than this ;-)

Issue #693, opened 2017-04-23
Nick completion should escape nicks when expand_escapes is on


ailin-nemui:
Example:

JC<Tab> -> JCaesa\r: 

when you send it, the \r results in a line break

PR #709, opened 2017-05-25
Escape nicks during nick completion when expand_escapes is enabled

osm:
This commit fixes issue #693.

I am not sure what the protocol for sending pull requests to this project are, so I just send it. Let me know if I have done something stupid so I can fix it, my C skills feels a bit rusty.

### DIFF
--- a/src/fe-common/core/completion.c
+++ b/src/fe-common/core/completion.c
@@ -137,8 +137,9 @@ char *word_complete(WINDOW_REC *window, const char *line, int *pos, int erase, i
         int old_startpos, old_wordlen;
 
 	GString *result;
-	char *word, *wordstart, *linestart, *ret;
-	int continue_complete, want_space;
+	const char *cmdchars;
+	char *word, *wordstart, *linestart, *ret, *data;
+	int continue_complete, want_space, expand_escapes;
 
 	g_return_val_if_fail(line != NULL, NULL);
 	g_return_val_if_fail(pos != NULL, NULL);
@@ -241,14 +242,24 @@ char *word_complete(WINDOW_REC *window, const char *line, int *pos, int erase, i
 	if (complist == NULL)
 		return NULL;
 
+        /* get the cmd char */
+	cmdchars = settings_get_str("cmdchars");
+
+	/* get the expand_escapes setting */
+	expand_escapes = settings_get_bool("expand_escapes");
+
+	/* escape if the word doesn't begin with '/' and expand_escapes are turned on */
+	data = strchr(cmdchars, *line) == NULL && expand_escapes ?
+		escape_string(complist->data) : g_strdup(complist->data);
+
 	/* word completed */
-	*pos = startpos+strlen(complist->data);
+	*pos = startpos + strlen(data);
 
 	/* replace the word in line - we need to return
 	   a full new line */
 	result = g_string_new(line);
 	g_string_erase(result, startpos, wordlen);
-	g_string_insert(result, startpos, complist->data);
+	g_string_insert(result, startpos, data);
 
 	if (want_space) {
 		if (!isseparator(result->str[*pos]))
@@ -256,13 +267,17 @@ char *word_complete(WINDOW_REC *window, const char *line, int *pos, int erase, i
 		(*pos)++;
 	}
 
-	wordlen = strlen(complist->data);
+	wordlen = strlen(data);
 	last_line_pos = *pos;
 	g_free_not_null(last_line);
 	last_line = g_strdup(result->str);
 
 	ret = result->str;
 	g_string_free(result, FALSE);
+
+	/* free the data */
+	g_free(data);
+
 	return ret;
 }
 


ailin-nemui:
hi, thanks for getting involved! I thought there was a function to do escaping already implemented somewhere in the source code, I will search for it tomorrow. And please remove your vim swap file from the commit ;-)

osm:
Hi,
Yeah you are of course right, I just found the escape_string function (and sorry about the swap file :-)

src/fe-common/core/chat-completion.c
```diff
@@ -639,6 +639,29 @@ static void complete_window_nicks(GList **list, WINDOW_REC *window,
...
+
+	/* verify that the list isn't NULL */
+	g_return_if_fail(list != NULL);
+
+	/* iterate over the nick list */
+	for (l = *list; l != NULL; l = l->next) {
+		/* only escape if we find either \\r or \\n in the nick */
+		if (l->data != NULL &&
+		    (strstr(l->data, "\\r") != NULL ||
+		     strstr(l->data, "\\n") != NULL)) {
```
dequis: `expand_escapes`/`expand_escape` handle much more than just `\r` and `\n`, there's \xFF, `\cA`, `\777` and so on.

dequis:
What's up with your commit timestamps? They seem to be off by a whole day at least.

I'm not sure it's a good idea to iterate over the whole completion list. It just seems like a bad place to introduce the escaping (also, there are other places in the same function that manipulate completion lists and would require the same thing). Would be better to do it at a point where the completion word is chosen from the list and it's added to the input box.

src/fe-common/core/chat-completion.c
osm: Yeah, my bad. So what would you prefer me to do here? A new function that handles just \r and \n?

osm:
Thanks for the feedback, much appreciated.

I was developing inside a virtual box running OpenBSD, apparently it didn't handle the time so well. Will fix it before I commit anything else.

I agree that it is not optimal to iterate over the whole list, I decided to do this because it was the only place I could find that only handled nick completions for messages in a channel. But this might just be because I'm too unfamiliar with your code base, would you mind pointing me to a file and line where you think it would be more suitable to place the escape?

I am not sure that we need to escape the nick somewhere else though. For example, when you do a /msg you don't want to escape the nick. I think it is only valid to escape the nick when sending something in a channel, right?

src/fe-common/core/chat-completion.c
osm: I added a new function now with the latest commit that takes a pattern and escapes all occurrences of the pattern in the submitted string. I hope this is an OK solution :)

dequis: Uh, sorry, I meant the opposite. You shouldn't be limiting this to just `\r` and `\n`

osm: Ah, yeah I was thrown by the initial issue, so I didn't think about that, rookie mistake :-).

I have removed the escape_pattern func now and the if statement you pointed out, so now it uses escape_string on all nicks.

I guess the only question now is the one below here, if you would like me to change where the escaping occurs or not. Let me know if you want me to fix this as well.

ailin-nemui:
hi, just to clarify, we need to escape the nicks when they are inserted on the input line, but *not* when they are part of a command. I haven't checked (yet) if your code does that or not

osm:
Hi,

Yeah I only tested with the /msg command and that worked, but just tried with /ctcp and then it applied my escaping, so that's not good. I will try to find another place to put the escaping instead.

osm:
So, here's a new stab on doing the escaping the right way, thanks for your patience :-)

I have done it like this now:
1) Moved the escaping to the word_complete function and removed my escape_nicks function I created in an earlier commit.
2) Stopped iterating over the complete list of nicks and just escape the word that will be displayed in the input line. 
3) Only escape words if the line doesn't begin with a / (this is how I determine if it's a command or not).
4) Added a check to the expand_escapes function so it doesn't expand our previously escaped nicks (or other text).

Question: Would you like me to create a new PR with just this one last commit if this last attempt is OK? I'm not sure how github works with merges, but if all my failed attempts will show up in the log it will be cleaner if I just abandon this PR and create a new one, what do you think?

ailin-nemui:
Hi, I still need to take a look at your changes but thanks for the effort. It sounds like you found a good solution. In the meantime you can grep the source for cmdchars, which you need to check instead of /. And once we're all happy with your code, you can "force push" your final commit into this PR, thereby removing the earlier tries from history

src/core/misc.c
```diff
@@ -726,7 +726,6 @@ int expand_escape(const char **data)
 	}
 }
 
-/* Escape all '"', "'" and '\' chars with '\' */
```
LemonBoy: What happened to this comment?

src/fe-common/core/completion.c
```diff
@@ -241,28 +241,37 @@ char *word_complete(WINDOW_REC *window, const char *line, int *pos, int erase, i
 	if (complist == NULL)
 		return NULL;
 
+	/* escape if the word doesn't begin with '/' and expand_escapes are turned on */
+	data = *line != '/' && settings_get_bool("expand_escapes") ?
```
LemonBoy: Are you trying to check whether `line` is a command?
If so you should honour the `cmdchars` option instead of checking for `/` only.

src/fe-common/core/completion.c
```diff
@@ -241,28 +241,37 @@ char *word_complete(WINDOW_REC *window, const char *line, int *pos, int erase, i
...
+		escape_string(complist->data) : complist->data;
+
 	/* word completed */
-	*pos = startpos+strlen(complist->data);
+	*pos = startpos+strlen(data);
```
LemonBoy: Spaces around the `+`, let it breathe!

src/fe-common/core/completion.c
```diff
@@ -241,28 +241,37 @@ char *word_complete(WINDOW_REC *window, const char *line, int *pos, int erase, i
...
+	wordlen = strlen(data);
 	last_line_pos = *pos;
 	g_free_not_null(last_line);
 	last_line = g_strdup(result->str);
 
 	ret = result->str;
 	g_string_free(result, FALSE);
+
+	/* free data if we have been escaping something */
+	if (data != complist->data)
```
LemonBoy: I'm not a great fan of such pointer comparisons, some other places in the code handle this situation by adding a `g_strdup` and always call `g_free` when they're done.

osm: Should also be fixed now :)

src/fe-common/core/completion.c
osm: Fixed.

src/fe-common/core/completion.c
osm: Fixed in the latest commit.

src/core/misc.c
osm: Ah, that was removed by accident. I have reverted this in the latest commit.

osm:
Everything should be fixed now according to the comments from @LemonBoy and @ailin-nemui. If you think everything looks OK now I will squash all commits into one and force push it again so I don't mess up the commit log completely :)

ailin-nemui:
can you explain to me again how (4) Added a check to the expand_escapes function works,  and why it is necessary?

osm:
The check added in expand_escapes checks if there are two backslashes in a row, if this is the case it indicates that we deliberately have been escaping something, so it moves past the two backslashes and continues to expand any other \r or \n that might exist.

Consider the following text that is going to be sent to a channel:
foo\\nbar: hello\nworld

Without this check the following message to a channel would produce this output:
21:03 < osm> foo
21:03 < osm> bar: hello
21:03 < osm> world

With the check the output will be like this, which I think is better (even though the nick will be incorrectly padded with an extra \, but I don't think there is any way around this).
21:04 < osm> foo\\nbar: hello
21:04 < osm> world

ailin-nemui:
hi @osm I think there is something odd about it, there is probably another bug somewhere which expands the codes multiple times(?), so in any case I think we need to fix them independently, but your code for the *tab completion* seems good

ailin-nemui:
maybe you can test my patch (and remove that bit from your code)

osm:
I have removed my fix for the #723 PR that you created now. I have also tested #723 and verified that it worked together with this PR and as you can see I squashed all the previous commits into one nice looking thing :)

ailin-nemui:
thanks!

PR #711, opened 2017-05-26
do not reset true colour bit on colour reset

### LINKED ISSUE #710: 24-bit colours don't seem to reset properly
I'm trying to write a theme that mimics the "Powerline" system. I have a [working version in 16-colour mode](https://github.com/darac/irssi-powerline), but I have been [asked](https://github.com/darac/irssi-powerline/issues/9) to extend it to 256-colour or 24-bit colour mode.

So, I have rearranged my theme to pre-define the colours, which I can then call on in the abstracts. The problem I'm having is that in 24-bit mode, I can't seem to switch from black-on-yellow to yellow-on-default without a blank space in between.

https://gist.github.com/darac/bed7e46141db6f0def5b44c9228a6796 is a minimal theme in 16-colour mode which *only* styles `ownmsgnick` (so it will look a little odd). This produces a correct line (timestamp in black-on-white, then a white-to-yellow chevron, then user mode in white-on-yellow, then nick in black-on-yellow, then a yellow-to-default chevron, then the message).

https://gist.github.com/darac/2d0465588d91197741866bac5efd6094 is the same theme in 24-bit-colour mode. The issue I have is on the last chevron. Where I have `{solf_yellow }`, this should produce a yellow-to-default (transparent) chevron. However, it *only* does so if there is (unstyled) text beforehand. That is, abutting the expandos doesn't seem to work correctly.

For reference, I am running:
 * "irssi 1.0.2-1 (20170310 1741)" with `colors_ansi_24bit = "yes";`
 * inside tmux 2.3 (inner `TERM=tmux-256color`, Tc bit added to terminfo)
 * inside mintty 2.7.4 (outer `TERM=xterm-256color`)

which should give me full 24-bit colour support, I think.

I'd appreciate if you could confirm if this is a bug in irssi, or in my theming code.

ailin-nemui:
fixes #710

### DIFF
--- a/src/fe-text/textbuffer-view.c
+++ b/src/fe-text/textbuffer-view.c
@@ -114,7 +114,6 @@ static void update_cmd_color(unsigned char cmd, int *color)
 		if (cmd & LINE_COLOR_BG) {
 			/* set background color */
 			*color &= FGATTR;
-			*color &= ~ATTR_FGCOLOR24;
 			if ((cmd & LINE_COLOR_DEFAULT) == 0)
 				*color |= (cmd & 0x0f) << BG_SHIFT;
 			else {
@@ -123,7 +122,6 @@ static void update_cmd_color(unsigned char cmd, int *color)
 		} else {
 			/* set foreground color */
 			*color &= BGATTR;
-			*color &= ~ATTR_BGCOLOR24;
 			if ((cmd & LINE_COLOR_DEFAULT) == 0)
 				*color |= cmd & 0x0f;
 			else {


ailin-nemui:
I think this is the right thing to do because we want to keep the flags, so in case of a partial reset (background reset, or foreground reset, or the terminfo parser returning to normal mode) at the same time of a 24bit colour code, Irssi would otherwise forget from which register to restore the colour (base colours or 24bit) and always pick the base one. I just hope I'm not overlooking something but it was probably simply a bug in my original implementation

ailin-nemui:
(it works when a space is added because then Irssi will send `%n%o<space>%ZB58900` instead of `%ZB58900%o`)

PR #719, opened 2017-06-17
Setting sasl_mechanism to '' disables the auth

### LINKED ISSUE #718: Add -nosasl flag to /network modify
This will make it easier to disable SASL if user requires

LemonBoy:
There was no easy way for the user to disable the SASL authentication
once the network was created.
Closes #718

### DIFF
--- a/docs/help/in/network.in
+++ b/docs/help/in/network.in
@@ -36,6 +36,7 @@
     -sasl_mechanism  Specifies the mechanism to use for the SASL authentication.
                      At the moment irssi only supports the 'plain' and the
                      'external' mechanisms.
+                     Use '' to disable the authentication.
     -sasl_username   Specifies the username to use during the SASL authentication.
     -sasl_password   Specifies the password to use during the SASL authentication.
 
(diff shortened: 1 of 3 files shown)


dequis:
Sounds good

ailin-nemui:
could we use empty string instead? many other settings work like that iirc

dequis:
How do you set settings to empty strings? There's "/set -clear" for global ones but network/server settings need a parameter and a pair of empty quotes won't help.

ailin-nemui:
try: /network add -autosendcmd 'xxx' testnet, /network add -autosendcmd '' testnet, works for me

LemonBoy:
Not quite, see [here](https://github.com/irssi/irssi/blob/master/src/fe-common/irc/fe-ircnet.c#L166).
I also don't really like using an empty string.

ailin-nemui:
so let's allow both none and ''? afai can see simply remove the check for *opt=='\0' to allow comparison of it?

LemonBoy:
I don't really like seeing `sasl_mechanism = none` in the output of `/network`, using `''` is fine.

dequis:
Should probably change the commit message too

ailin-nemui:
is this good to merge? should it maybe also be possible to clear username/password?

LemonBoy:
I've extended this behaviour to the `sasl_username` and `sasl_password` fields too.
Feel free to merge once Travis gives the green.

PR #723, opened 2017-06-19
fix weird n-fold unescaping in expand_escapes

ailin-nemui:
noticed by @osm in #709

### DIFF
--- a/src/fe-common/core/chat-completion.c
+++ b/src/fe-common/core/chat-completion.c
@@ -1041,9 +1041,9 @@ static char *expand_escapes(const char *line, SERVER_REC *server,
 			   event to handle it (or actually the text before
 			   the newline..) */
 			if (prev != line) {
-				const char *prev_line = g_strndup(prev, (line - prev) - 1);
+				char *prev_line = g_strndup(prev, (line - prev) - 1);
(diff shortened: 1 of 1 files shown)


osm:
I tested this together with the PR in #709 and it works correctly.

src/fe-common/core/chat-completion.c
```diff
@@ -1036,9 +1040,11 @@ static char *expand_escapes(const char *line, SERVER_REC *server,
 			/* newline .. we need to send another "send text"
 			   event to handle it (or actually the text before
 			   the newline..) */
-			if (ret != ptr) {
-				*ptr = '\0';
-				signal_emit("send text", 3, ret, server, item);
```
ailin-nemui: in any case calling "send text" again to implement this "hack" would mean
"send text" "abc\ndef"
"send text" "abc"
being emitted, which seems wrong. the fe-text handler on the other hand catches the send text first and thus expand_escapes doesn't work at all (also sad ;-(

dequis: Hmm. It's not affecting escapes at all in fe-exec either before or after this change. I guess this is fine.

PR #725, opened 2017-06-23
Subst vars

LemonBoy:
Extra wip.

Loosely inspired by `muttrc` and its backtick syntax, this is the nicest thing I could carve out from our not-so-friendly usage of `GScanner`: prepending a `&` before a quoted variable runs the content trough a user-supplied function, right now it is just a simple stub.

Nothing's set in stone so feel free to drop in with any suggestion about the design.

### Todo
* Decide if this approach is worth pursuing.
* Remove the `g_assert` and provide a nice way to recover from that error case.
* Ownership of the `new_value`.
* Should we cache the resulting values?
* Fuzz the shit out of this.

### DIFF
--- a/src/core/settings.c
+++ b/src/core/settings.c
@@ -703,6 +703,13 @@ int irssi_config_is_changed(const char *fname)
 		 config_last_checksum != file_checksum(fname));
 }
 
+static gboolean filter_stub(const char *key, const char *value, char **new_value)
+{
+	g_warning("Filtering %s in %s", value, key);
+	*new_value = (char *)value;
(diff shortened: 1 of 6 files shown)


ailin-nemui:
nice for working on this. My initial recommendation was a bit different:
* add /eval support to several variables (like password) (it is already present in e.g. -autosendcmd); 
* make a global /set eval_vars on to keep backwards compat;
* introduce function arguments using the currently broken (and thus unused) $(function,args) syntax

ailin-nemui:
related to: #6

PR #726, opened 2017-06-24
Add a startup warning if the TERM var is wrong inside tmux/screen

dequis:
One of the most common and confusing issues we get in #irssi, this should help identifying and mitigating it.

----

screnshot

> 12:34 -!- Irssi: warning The TERM environment variable is set to 'xterm' which can cause display glitches when running under tmux.
> 12:34 -!- Irssi: Consider changing TERM to 'screen' or 'screen-256color' instead.

Feel free to bikeshed the text of the message, not sure i'm entirely happy with it myself.

* The "can cause display glitches" has a bit too much certainty given how little this code knows about screen-compatible TERM values. Maybe it's ok.

* I like the tradeoff of suggesting only 'screen' or 'screen-256color' and not mentioning 'tmux' and 'tmux-256color' - while the latter is slightly better, I found it's a bit annoying to set globally and then get errors when you ssh to servers with older terminfo that don't know about it.

* Mentioning the name of the multiplexer hopefully helps mitigate the unavoidable "it tells me to use screen but i'm using tmux" but it's not foolproof. Not that anything is.

* I think half of the issues are about someone setting TERM inside .bashrc or equivalent, but some of them have the wrong TERM inside the multiplexer config itself. Could consider guiding users to look in some of those places. Or not.

Out of scope concern: Starting my main irssi connects me to a trillion servers and this message would be lost up there, but the same thing applies to script load failures and other early-startup errors.

### DIFF
--- a/src/fe-text/irssi.c
+++ b/src/fe-text/irssi.c
@@ -207,6 +207,8 @@ static void textui_finish_init(void)
 		fe_settings_set_print("nick");
 	if (user_settings_changed & USER_SETTINGS_HOSTNAME)
 		fe_settings_set_print("hostname");
+
+	term_environment_check();
 }
 
 static void textui_deinit(void)
--- a/src/fe-text/term-terminfo.c
+++ b/src/fe-text/term-terminfo.c
@@ -733,3 +733,31 @@ void term_gets(GArray *buffer, int *line_count)
 		}
 	}
 }
+
+static const char* term_env_warning =
+	"The TERM environment variable is set to '%s' which can cause display "
+	"glitches when running under %s.\n"
+	"Consider changing TERM to 'screen' or 'screen-256color' instead.";
+
+void term_environment_check(void)
+{
+	const char *term, *sty, *tmux, *multiplexer;
+
+	term = g_getenv("TERM");
+	sty = g_getenv("STY");
+	tmux = g_getenv("TMUX");
+
+	multiplexer = (sty && *sty) ? "screen" :
+	              (tmux && *tmux) ? "tmux" : NULL;
+
+	if (!multiplexer) {
+		return;
+	}
+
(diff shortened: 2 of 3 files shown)


ailin-nemui:
yea, weechat has the same warning, wonder if it helps?

ailin-nemui:
I'm indifferent but would suggest the help should mention the `tmux-`terms  when running under tmux as well

ailin-nemui:
@dequis what do to

dequis:
Oh hey totally forgot I wrote this. Saw the notification and thought "cool, someone wrote it for me".

Re: mentioning `TERM=tmux`

>I like the tradeoff of suggesting only 'screen' or 'screen-256color' and not mentioning 'tmux' and 'tmux-256color' - while the latter is slightly better, I found it's a bit annoying to set globally and then get errors when you ssh to servers with older terminfo that don't know about it.

I think it only exists since ncurses 6.0

So if you're asking for my opinion, I think my opinion is right :P

ailin-nemui:
the "problem" is that italics only works in tmux if tmux sees that TERM is set to tmux

ailin-nemui:
@dequis are you still interested in finishing this up?

dequis:
Define "finishing this up"?

I still prefer to mention something that works every time instead of adding more failure modes just to have a minor feature work.

dequis:
Rebased and fixed the conflict, feel free to just push stuff if you want to tweak the message.

ailin-nemui:
I changed the message because once this lands hopefully ncurses6 is already widespread

dequis:
Neat, positional formatting, don't see that often, clever. This is okay.

ailin-nemui:
I actualy screwed up with the space lol

PR #734, opened 2017-07-28
Don't compute log_dir_create_mode in three different places.

trasz:


dequis:
I guess that's ok. Not sure what motivated this.

src/core/rawlog.c
```diff
@@ -130,6 +130,7 @@ void rawlog_open(RAWLOG_REC *rawlog, const char *fname)
 			      log_file_create_mode);
 	if (rawlog->handle == -1) {
 		g_warning("rawlog open() failed: %s", strerror(errno));
+		return;
```
LemonBoy: You leak `path`, move the `g_free` before the if

trasz: You're right, thanks!  (That's what happens when you try to fix things before the morning coffee...)

PR #738, opened 2017-08-09
Fix indentation; no functional changes.

trasz:


josephbisch:
Thanks, it is good that this is being fixed, but there are still other lines in that file that use spaces instead of tab indent.

trasz:
Yay at giving up the coffee, which makes you able to fail at even the simplest things!

PR #741, opened 2017-08-29
Complete filenames ending with a slash

LemonBoy:


ailin-nemui:
please throw me some examples

LemonBoy:
* `/script load auto<tab>` -> `/script load autorun/` -> add a trailing slash so that the completion mechanism knows we want the content of the folder then press `<tab>` -> `/script load autorun/file.pl`
* `/script load autorun/<tab>` -> shows the content of the folder

Issue #740, opened 2017-08-28
Color codes which end in commas both reset background and print comma as literal


GinjaNinja32:
When irssi formats an IRC message containing a color code such as `^C2,`, it treats the `,` as both literal and part of the color code; that is:

`^C1,0black on white; ^C2blue on white` -> `black on white; `, `blue on white`, in the specified colors.
`^C1,0black on white; ^C2,blue on default` -> `black on white; `, `,blue on default`, again in the specified colors; note the comma before "blue" in the result.

`^C2,` resets the background, and sets the foreground to blue, then prints a comma.
`^C,` resets the background but not foreground, and prints the comma similarly.

It appears any color code which ends with a comma *both* resets the background and adds a literal comma to the string - it should likely only do one of these things; IMO it should reset the background, but I don't mind much either way.

Found on irssi 0.8.20; reproduced successfully on irssi 1.0.4.

PR #742, opened 2017-08-31
Reset foreground and background while only comma is specified

SilverRainZ:
Currently irssi can not correctly process `^C,` which is a valid mirc color code.

For `^C4,2ir^C,ssi`, the foreground color was inherited and left a comma: 

before

But `^C,` has neither foreground nor background, it should be treated as same as `^C`:

after

### DIFF
--- a/src/fe-common/core/formats.c
+++ b/src/fe-common/core/formats.c
@@ -1067,8 +1067,9 @@ static const char *get_ansi_color(THEME_REC *theme, const char *str,
 /* parse MIRC color string */
 static void get_mirc_color(const char **str, int *fg_ret, int *bg_ret)
 {
-	int fg, bg;
+	int fg, bg, hasfg;
 
+	hasfg = 0;
 	fg = fg_ret == NULL ? -1 : *fg_ret;
 	bg = bg_ret == NULL ? -1 : *bg_ret;
 
@@ -1078,6 +1079,7 @@ static void get_mirc_color(const char **str, int *fg_ret, int *bg_ret)
 	} else {
 		/* foreground color */
 		if (**str != ',') {
+			hasfg = 1;
 			fg = **str-'0';
 			(*str)++;
 			if (i_isdigit(**str)) {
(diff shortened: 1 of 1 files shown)


ailin-nemui:
does anyone have a mIRC ready to compare if it should be `ir,ssi` or `irssi`? does this fix issue #740  ?

SilverRainZ:
@ailin-nemui I have not a mIRC for testing so I only tested on hexchat and srain.
It really fixes #740.

ailin-nemui:
we had this bug: http://bugs.irssi.org/index.php?do=details&task_id=250

SilverRainZ:
The one on the right is mIRC? It does't regard the comma as part of color code in `^C,`, which means that the foreground color is never inherited -- I think it is unreasonable.

ailin-nemui:
I think mIRC is the "de facto" standard on mIRC colour codes so we may need to accept the fact that the foreground can not be inherited, `^C,` is `^C` then`,`...

dequis:
irssi:

scrot

mirc 7.47:

scrot

hexchat

scrot

weechat

scrot

After doing this I realized there are other more interesting test cases up there. Welp. I'll get to that later.

Regardless of the decisions made here: any changes to this code must go through fuzzing before they are merged. This is particularly fragile code.

ailin-nemui:
I think the changes itself are good code wise but they don't match mirc

SilverRainZ:
Well, now it act like mirc.

ailin-nemui:
how about this:
```diff
diff --git a/src/fe-common/core/formats.c b/src/fe-common/core/formats.c
index 005e6fb7..a340bffa 100644
--- a/src/fe-common/core/formats.c
+++ b/src/fe-common/core/formats.c
@@ -1072,7 +1072,8 @@ static void get_mirc_color(const char **str, int *fg_ret, int *bg_ret)
 	fg = fg_ret == NULL ? -1 : *fg_ret;
 	bg = bg_ret == NULL ? -1 : *bg_ret;
 
-	if (!i_isdigit(**str) && **str != ',') {
+	if (!i_isdigit(**str)) {
+		/* turn off color */
 		fg = -1;
 		bg = -1;
 	} else {
@@ -1085,11 +1086,8 @@ static void get_mirc_color(const char **str, int *fg_ret, int *bg_ret)
 				(*str)++;
 			}
 		}
-		if (**str == ',') {
+		if ((*str)[0] == ',' && i_isdigit((*str)[1])) {
 			/* background color */
-			if (!i_isdigit((*str)[1]))
-				bg = -1;
-			else {
 				(*str)++;
 				bg = **str-'0';
 				(*str)++;
@@ -1097,7 +1095,6 @@ static void get_mirc_color(const char **str, int *fg_ret, int *bg_ret)
 					bg = bg*10 + (**str-'0');
 					(*str)++;
 				}
-			}
 		}
 	}
 
```

PR #743, opened 2017-08-31
OS X install notes

nick96:
I was having problems building Irssi from source on OS X and found the issue was that the system openSSL library was the problem. I updated the INSTALL file so that hopefully people don't have to
search too far if they have this problem as well.

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -82,8 +82,8 @@ things that can go wrong:
  - If configure complains that it doesn't find some perl stuff, you're
    probably missing libperl.so or libperl.a. In debian, you'll need to do
    apt-get install libperl-dev
- - For unprivileged home directory installations, you probably do not want 
-   to specify --with-perl-lib=(site|vendor). Instead, you can use the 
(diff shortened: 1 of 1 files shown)


dequis:
Sounds more like a bug we should fix

nick96:
Fair enough, is that something that can be determined in the configure file? I used the path for a homebrew install but others might have a different path

nick96:
[config.log and output of export](https://gist.github.com/nick96/cddd05d4f99bfa36b8d098ffa8231038)

ailin-nemui:
I still can't figure it out, can you please show me also the broken and the working compile log (with `make V=1`)

and also double check if you provided the right files, as those 2 config.logs are 100% identical

ailin-nemui:
similar error description on another project: https://github.com/NagiosEnterprises/nrpe/issues/164

nick96:
I can no longer replicate this as I'm no longer on OS X, sorry I can't be of more help.

ailin-nemui:
closing for now

PR #749, opened 2017-09-10
Save backup config during /upgrade

### LINKED ISSUE #690: Have /upgrade either save the current layout/config in /tmp or prompt for confirmation
Dear all,

Today was the day I discovered the `/upgrade` irssi command. Today was also the day I used it for the first time. And incidentally, today was the day I lost most of my recent configuration.

This could have been avoided in two different ways:

1. By prompting the user (me) to confirm having understood that the command will effectively reload my current config
2. By saving the running config to a temporary location (for example in `/tmp`) and reloading that config after the `/upgrade` command is ran.

The first one doesn't require a lot of work, but the second one would be a really good feature.

josephbisch:
Fixes #690 by using a file named config.clientbackup in ~/.irssi.

This is just a suggestion for how to handle #690. There may be a better way to handle this.

This could use testing. Here's the expected flow of things:

1. Run irssi.
2. Edit config within irssi.
3. /upgrade
4. /save
5. /quit

After that, you should have ~/.irssi/config, which is the same as before, except whatever config options you changed before the upgrade are reflected in the config file. There was a ~/.irssi/config.clientbackup, but that should be deleted upon successful save.


josephbisch:
Ready for testing and review. Don't forget to backup your config before testing.

ailin-nemui:
just a random thought, haven't reviewed: iirc you can run irssi with --config=myconf, wonder if it would be better to have a backup "suffix" instead. but this is likely quite a rare case

josephbisch:
Right, so if someone is using --config=myconf to switch between two configs, this won't work. So you are right, it should be $config.clientbackup instead of always being config.clientbackup.

ailin-nemui:
maybe just check for a $config.clientbackup instead of explicitly passing the backup file in --config. When does the modified && clientbackup != NULL case happen?

ailin-nemui:
@josephbisch this still needs some changes

PR #754, opened 2017-09-13
add new function to set the position in bytes

### LINKED ISSUE #752: tab completion miscounts characters
report by wodim

> if you type (| is the caret)

> ññññññññ | hello

> and type the first letters of a nick and  then tab, the caret moves to the end of the line
> instead of staying at the end of the nick

ailin-nemui:
fixes #752

### DIFF
--- a/src/fe-text/gui-entry.c
+++ b/src/fe-text/gui-entry.c
@@ -936,6 +936,26 @@ void gui_entry_set_pos(GUI_ENTRY_REC *entry, int pos)
 	gui_entry_draw(entry);
 }
 
+void gui_entry_set_text_and_pos_bytes(GUI_ENTRY_REC *entry, const char *str, int pos_bytes)
+{
+	int pos;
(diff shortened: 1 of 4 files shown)


src/fe-text/gui-entry.c
```diff
@@ -936,6 +936,26 @@ void gui_entry_set_pos(GUI_ENTRY_REC *entry, int pos)
 	gui_entry_draw(entry);
 }
 
+void gui_entry_set_text_and_pos_bytes(GUI_ENTRY_REC *entry, const char *str, int pos_bytes)
+{
+	int pos;
+	const char *ptr;
+
+        g_return_if_fail(entry != NULL);
```
dequis: I see you're trying to stay consistent with the existing indentation style

ailin-nemui: copy&paste of bad indent. fixed

Issue #103, opened 2014-07-20
ssl passphrase support


darix:
I am not sure that saving the password in plain text in the irssi config kind of defeats the purpose of the password in first place. openSUSE had a patch that always asked for the passphrase but then would of course block in the reconnect case. not ideal either.

how about asking on the first connect and caching the passphrase in memory?
and maybe give the passphrase dialog a timeout.

sounds better than storing it in plaintext.

ailin-nemui:
closing this for now duplicate of #6

PR #757, opened 2017-09-14
ask user for ssl client cert password

TauPan:
Putting plaintext passwords in the irssi config might not conform to security requirements for some ssl client certificates.

see http://bugs.irssi.org/index.php?do=details&task_id=305&project=5

I'm not quite sure how to do this properly so I get a password prompt on the statusbar, which redraws afterwards.

Looking forward to your feedback on this.

### DIFF
--- a/src/core/network-openssl.c
+++ b/src/core/network-openssl.c
@@ -33,6 +33,8 @@
 #include <openssl/pem.h>
 #include <openssl/ssl.h>
 #include <openssl/err.h>
+#include <termios.h>
+#include <stdio.h>
 
 /* OpenSSL 1.1.0 introduced some backward-incompatible changes to the api */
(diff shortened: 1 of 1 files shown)


ailin-nemui:
your code would block irssi while waiting for the password, which is not acceptable

TauPan:
@ailin-nemui Oh. My use-case is logging into a single server with a configured ssl_cert on irssi startup.

You mean if the user receives a disconnect from a server irssi would block every other connection as well, while waiting for the password?

I certainly didn't anticipate that.

Since I'm not familiar with irssi's codebase I don't know how to proceed. If you see no other option to integrate this into irssi the proper way, I'm ok with closing this ticket.

ailin-nemui:
I'm afraid the best we have come up so far with is to put it in /server -ssl_pass (which may get saved to your config)

for some related discussions general to passwords, you may also read issue #6 (and previously #103)

ailin-nemui:
closing for inactivity

PR #762, opened 2017-10-05
allow access to global command history when using a specifc history

ailin-nemui:
also add history loading method

### DIFF
--- a/src/fe-common/core/command-history.c
+++ b/src/fe-common/core/command-history.c
@@ -307,6 +307,23 @@ const char *command_global_history_next(WINDOW_REC *window, const char *text)
 	return command_history_next_int(window, text, TRUE);
 }
 
+const char *command_history_delete_current(WINDOW_REC *window, const char *text)
+{
+	HISTORY_REC *history;
+	GList *pos;
+
+	history = command_history_current(window);
+	pos = history->pos;
+
+	if (pos != NULL && g_strcmp0(((HISTORY_ENTRY_REC *)pos->data)->text, text) == 0) {
+		((HISTORY_ENTRY_REC *)pos->data)->history->lines--;
+		history_list_delete_link_and_destroy(pos);
+	}
+
+	history->redo = 0;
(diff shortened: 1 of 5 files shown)


ailin-nemui:
usage: ctrl+up/down

ailin-nemui:
sample script for your autorun:

```perl
use strict;
use warnings;
use YAML::Tiny;

my $histfile = Irssi::get_irssi_dir()."/history";

my ($hist) = eval { YAML::Tiny::LoadFile($histfile) };
if ($hist) {
    Irssi::UI::Window::load_history_entries(undef, @$hist);
}

Irssi::signal_add 'gui exit' => sub {
    if (Irssi::settings_get_bool('settings_autosave')) {
	YAML::Tiny::DumpFile($histfile, [ Irssi::UI::Window::get_history_entries(undef) ] )
    }
}
```

src/fe-common/core/command-history.c
```diff
@@ -30,32 +30,108 @@
...
+static void history_entry_destroy(HISTORY_ENTRY_REC *entry)
+{
+	g_free((char *)entry->text);
+	g_free(entry);
+}
+
+#define history_list_last command_history_list_last
+#define history_list_first command_history_list_first
+#define history_list_prev command_history_list_prev
+#define history_list_next command_history_list_next
```
dequis: Huh. Why?

dequis:
Good stuff.

The commit log / PR title could use some more love. I mean, global window history is the default behavior, but being able to use global history while window history is enabled is a very good feature.

Not sure I fully understand what g_history is supposed to hold.

Leaving the job of persisting history to a perl script is a weird choice (not wrong, but not one i'd choose) but I guess you like perl, and more APIs don't hurt. It's a good example of adding more perl stuff, I guess.

src/fe-common/core/command-history.c
ailin-nemui: fixed

ailin-nemui:
I renamed g_history to history_entries

Issue #758, opened 2017-09-18
The /statusbar documentation is bad


dequis:
The whole thing is just not great at explaining how statusbars and statusbar items work, it's not helpful if you don't already know how to manipulate them.

The existing text has some conceptual errors too. Most glaring mistake is that `/STATUSBAR <name> ADD <item>` does not actually add statusbars, but items. I think enabling means adding statusbars, disabling means removing, unless it's a builtin statusbar

There's more text here https://irssi.org/documentation/startup/#statusbar - certainly better than the /help text.

A quadpoint-style illustrated explanation in a separate page of the website would definitely be cool.

Also the process of writing these docs will highlight what parts of statusbars themselves are bad.

ailin-nemui:
is this fixed now?

dequis:
Yeah, I was asking a bit too much in the description, that PR is good enough

PR #763, opened 2017-10-06
Improvements to statusbar documentation and help text.

rbisewski:
As per issue #758, I have made an attempt at improving both the HOWTO and the "/help statusbar" content to better reflect the current state of irssi statusbar functionality.

### DIFF
--- a/docs/help/in/statusbar.in
+++ b/docs/help/in/statusbar.in
@@ -17,8 +17,8 @@
     POSITION:     Sets the position of the statusbar. Represented as a
                   number, with 0 implying the first position.
     VISIBLE:      Sets the visibility of the statusbar or item. If set to
-                  to always it is visible on all screens, otherwise if set
-                  to inactive or active then it is only visible on inactive
+                  always it is visible on all screens, otherwise if set to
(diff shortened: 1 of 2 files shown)


dequis:
Thanks!

Looks like we totally forgot there was a copy of the startup doc in this repo. It's outdated relative to the one in https://github.com/irssi/irssi.github.io/blob/master/documentation/startup/index.markdown

Can you drop that part of this PR and send those changes to the other repo? We'll just use pandoc or something to generate a whole new startup-HOWTO.html

Haven't reviewed too much for correctness but sounds about right. Nitpicks on wording: I'd use "number" over "unsigned integer", and maybe change those "refers to" to "sets" (assuming all of them are commands to set attributes of the statusbar, right?)

I'm not sure disable completely removes the statusbar every time. I think that's just the non-builtin ones.

rbisewski:
I have since made the requested changes to the pull-request from your above feedback, and have created a fork of your site and added the HOWTO changes into it.

Let me know if you notice anything else.

ailin-nemui:
thanks

PR #764, opened 2017-10-06
Add color support for input bar

GinjaNinja32:
This is mostly a rebase of [this patch](http://bugs.irssi.org/index.php?do=details&task_id=621), with a few bugfixes (a segfault in `gui_entry_transpose_words`, and a pointer-to-int assignment).

Allows scripts to specify that ranges of characters in the input bar should be rendered in specific colors; the original intent for this was to implement a spellchecker, my use is to colorise nicks in the input bar.

Example script using this, mostly based on my rewritten [colorize_nicks.pl](https://github.com/irssi/scripts.irssi.org/blob/master/scripts/colorize_nicks.pl): https://thanatos.gn32.uk/f/colorise_input.pl (I'll likely clean up the code a little and submit for scripts.irssi.org, assuming this PR gets merged)

### DIFF
--- a/src/fe-text/gui-entry.c
+++ b/src/fe-text/gui-entry.c
@@ -65,8 +65,10 @@ static void entry_text_grow(GUI_ENTRY_REC *entry, int grow_size)
 	entry->text_alloc = nearest_power(entry->text_alloc+grow_size);
 	entry->text = g_realloc(entry->text,
 				sizeof(unichar) * entry->text_alloc);
-	entry->colors = g_realloc(entry->colors,
-	                        sizeof(int) * entry->text_alloc);
+
+	if (entry->uses_extents)
+		entry->extents = g_realloc(entry->extents,
+				   sizeof(char *) * entry->text_alloc);
 }
 
 GUI_ENTRY_REC *gui_entry_create(int xpos, int ypos, int width, int utf8)
@@ -79,10 +81,25 @@ GUI_ENTRY_REC *gui_entry_create(int xpos, int ypos, int width, int utf8)
 	rec->width = width;
 	rec->text_alloc = 1024;
 	rec->text = g_new(unichar, rec->text_alloc);
-	rec->colors = g_new(int, rec->text_alloc);
+	rec->extents = NULL;
 	rec->text[0] = '\0';
 	rec->utf8 = utf8;
 	return rec;
+}
+
+static void destroy_extents(GUI_ENTRY_REC *entry)
+{
+	if (entry->uses_extents) {
+		int i;
+		for (i = 0; i < entry->text_alloc; i++) {
+			if (entry->extents[i] != NULL) {
+				g_free(entry->extents[i]);
+			}
+		}
+	}
+	g_free(entry->extents);
+	entry->extents = NULL;
+	entry->uses_extents = FALSE;
 }
 
 void gui_entry_destroy(GUI_ENTRY_REC *entry)
@@ -103,8 +120,8 @@ void gui_entry_destroy(GUI_ENTRY_REC *entry)
 	}
 	g_slist_free(entry->kill_ring);
 
+	destroy_extents(entry);
 	g_free(entry->text);
-	g_free(entry->colors);
 	g_free(entry->prompt);
 	g_free(entry);
 }
@@ -168,15 +185,36 @@ void big5_to_unichars(const char *str, unichar *out)
 	*out = '\0';
 }
 
+/* Return screen length of plain string */
+static int scrlen_str(const char *str, int utf8)
+{
+	int len = 0;
+	char *stripped;
+	g_return_val_if_fail(str != NULL, 0);
+
+	stripped = strip_codes(str);
+	len = string_width(stripped, utf8 ? TREAT_STRING_AS_UTF8 : TREAT_STRING_AS_BYTES);
+	g_free(stripped);
+	return len;
+}
+
 /* ----------------------------- */
 
-static int pos2scrpos(GUI_ENTRY_REC *entry, int pos)
+static int pos2scrpos(GUI_ENTRY_REC *entry, int pos, int cursor)
 {
 	int i;
 	int xpos = 0;
 
+	if (!cursor && pos <= 0)
+		return 0;
+
+	if (entry->uses_extents && entry->extents[0] != NULL) {
+		xpos += scrlen_str(entry->extents[0], entry->utf8);
+	}
+
 	for (i = 0; i < pos; i++) {
 		unichar c = entry->text[i];
+		const char *extent = entry->uses_extents ? entry->extents[i+1] : NULL;
 
 		if (term_type == TERM_TYPE_BIG5)
 			xpos += big5_width(c);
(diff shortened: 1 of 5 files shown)


src/fe-text/gui-entry.c
```diff
@@ -248,10 +253,13 @@ static void gui_entry_draw_from(GUI_ENTRY_REC *entry, int pos)
 	if (xpos > end_xpos)
                 return;
 
+	color = ATTR_RESET;
 	term_set_color(root_window, ATTR_RESET);
 	term_move(root_window, xpos, entry->ypos);
 
-	for (i = entry->scrstart + pos; i < entry->text_len; i++) {
+	col = entry->scrstart + pos < entry->text_len ?
+			entry->colors + entry->scrstart + pos : 0;
```
LemonBoy: `s/0/NULL/`

src/fe-text/gui-entry.c
```diff
@@ -266,14 +274,19 @@ static void gui_entry_draw_from(GUI_ENTRY_REC *entry, int pos)
 		if (xpos > end_xpos)
 			break;
 
+		if (*col != color) {
```
LemonBoy: If `col` is set to `NULL` above you get a NPE.

src/fe-text/gui-entry.c
```diff
@@ -505,6 +522,10 @@ void gui_entry_insert_text(GUI_ENTRY_REC *entry, const char *str)
 		}
 	}
 
+	for(i = 0; i < len; i++) {
```
LemonBoy: Missing space after `for`.

src/fe-text/gui-entry.c
```diff
@@ -700,6 +726,9 @@ void gui_entry_erase(GUI_ENTRY_REC *entry, int size, CUTBUFFER_UPDATE_OP update_
 	g_memmove(entry->text + entry->pos - size, entry->text + entry->pos,
 		  (entry->text_len-entry->pos+1) * sizeof(unichar));
 
+	g_memmove(entry->colors + entry->pos - size, entry->colors + entry->pos,
```
LemonBoy: The `memmove` above has a `+1` that's missing here, why is that?

src/fe-text/gui-entry.c
```diff
@@ -719,7 +748,10 @@ void gui_entry_erase_cell(GUI_ENTRY_REC *entry)
 		       mk_wcwidth(entry->text[entry->pos+size]) == 0) size++;
 
 	g_memmove(entry->text + entry->pos, entry->text + entry->pos + size,
-		  (entry->text_len-entry->pos-size+1) * sizeof(unichar));
+	          (entry->text_len-entry->pos-size+1) * sizeof(unichar));
+
+	g_memmove(entry->colors + entry->pos, entry->colors + entry->pos + size,
```
LemonBoy: Ditto as before.

src/fe-text/gui-entry.c
```diff
@@ -830,31 +867,55 @@ void gui_entry_transpose_words(GUI_ENTRY_REC *entry)
...
+		}
 
 		g_free(first);
 		g_free(sep);
 		g_free(second);
 
+		g_free(first_color);
+		g_free(sep_color);
+		g_free(second_color);
+
```
LemonBoy: Stray newline.

src/fe-text/gui-entry.c
```diff
@@ -1042,3 +1103,31 @@ void gui_entry_redraw(GUI_ENTRY_REC *entry)
...
+	end = pos + len;
+
+	if (end > entry->text_len)
+		end = entry->text_len;
+
+	for (i = pos; i < end; i++) {
+		if (entry->colors[i] != color) {
+			entry->colors[i] = color;
+			update = 1;
```
LemonBoy: `s/1/TRUE/`

src/fe-text/gui-entry.c
GinjaNinja32: I think the intention was that `col` is only set to `NULL` if the loop won't run.

Perhaps this should instead use `entry->colors[i]`, since `col` is equal to `entry->colors + i` at all points during the loop?

src/fe-text/gui-entry.c
GinjaNinja32: I'm not entirely sure. This is straight from the original patch.

Is `entry->text` `NULL`-terminated, and `entry->colors` implicitly has the same number of elements?

src/fe-text/gui-entry.c
GinjaNinja32: I assume `FALSE` on L1109 too?

src/fe-text/gui-entry.c
```diff
@@ -1042,3 +1100,31 @@ void gui_entry_redraw(GUI_ENTRY_REC *entry)
 	gui_entry_fix_cursor(entry);
 	gui_entry_draw(entry);
 }
+
+void gui_entry_set_color(GUI_ENTRY_REC *entry, int pos, int len, int color)
```
LemonBoy: I think we should add a couple of checks to make sure `pos` and `len` are `>= 0` to avoid some potential memory corruption bugs due to careless scripts.

src/fe-text/gui-entry.c
```diff
@@ -1042,3 +1103,31 @@ void gui_entry_redraw(GUI_ENTRY_REC *entry)
...
```
LemonBoy: I think we should add a couple of checks to make sure `pos` and `len` are `>= 0` to avoid some potential memory corruption bugs due to careless scripts.

GinjaNinja32: Done in https://github.com/irssi/irssi/pull/764/commits/8326588c9f34f0670768856a5781800cd27cf3e3

ailin-nemui:
seems mostly ok, but I don't like having to use the internal colour attributes from XS -- maybe we can add in some functions to do colour code conversions between \004, % and this. And another thing: it doesn't do 24bit "true colour" (api is term_set_color2, didn't exist back then)

The more general question is if we like this implementation. Alternative would be to have in-line colours like in print text / gui prompt. But this code here is much simpler.

In general I think it would be a nice to have feature for your and the other use case

LemonBoy:
> maybe we can add in some functions to do colour code conversions between \004, % and this.

What if we reuse the same parsing logic that's used for themes formats? It's not that good but at least spares us from writing another parser and having the script writers learn a new set of specifiers.

> Alternative would be to have in-line colours

The principle of separation between data and metadata applies here.
A script may as well render the inline colours though, that's an interesting idea.

> other use case

I'm using it to turn the whole prompt red whenever there's a `cmdchar` with leading rubbish, no more accidental commands for me 😃

ailin-nemui:
> I'm using it to turn the whole prompt red whenever there's a cmdchar with leading rubbish, no more accidental commands for me

if we had something more generic like in line colours then it could offer more flexibility like also in line altering of the actual displayed text. think of a blinking "warning" appended to the prompt or an in line display of spelling suggestions after the unrecognised word

GinjaNinja32:
I'm not entirely sure how inline colors would work here.

Using format codes instead of ints sounds good, though. `get_nick_color2` appears to return \004-style codes, so it'd be good if it accepts those - perhaps a string containing either a `\cD..` code or a `%#`/`%X##` code?

I have no idea how I'd implement that, but I can take a look and poke around IRC and see if I can put something together.

ailin-nemui:
just conceptually speaking, one way to think about inline colours would be to take the input
```
 /whois xxx
```
then call a function (e.g. XS) which would change it to
```
[NOT COMMAND!]  /whois xxx
```
and that's what would then be visible. This would demand the callback to transform "input" into "displayed input" to be called at basically every change to the prompt

colours would then be realised by "returning"
```
%R /whois xxx
```
from the signal handler. There /should/ be some methods in irssi (which may need to be moved to some util/misc function) that turn %codes into the appropriate attributes

ailin-nemui:
todo: change XS to work on %-codes (maybe exposing some functions for \004 conversion at the same time)

ailin-nemui:
first version, still some errors, feel free to play with it and fix the remaining bugs

ailin-nemui:
it still has some problems with pos2scrpos and the redraw_from when the input is scrolled because of long texts

ailin-nemui:
Irssi::gui_input_clear_extents(pos, len)
Irssi::gui_input_get_extent(pos)
Irssi::gui_input_set_extents(pos, len, left, right)

ailin-nemui:
http://anti.teamidiot.de/static/nei/*/Code/Irssi/spellcheck.pl

ailin-nemui:
@GinjaNinja32 I changed the api again slightly,extent -> extents. hope you can test this again and discuss any usability issues that we may need to fix before merge

ailin-nemui:
@irssi/developers I think we should merge this

PR #766, opened 2017-10-08
Fix MODE parameter parsing

### LINKED ISSUE #601: `MODE #foo +o :user` gets parsed as `+o :user`, not `+o user`
See https://github.com/sorcix/irc/issues/26 for details (first part of the ticket has irssi-specific stuff)

> Placing a breakpoint on `modes_type_prefix()` and looking at the backtrace, I see:
> 
> ```
> #0  modes_type_prefix (channel=0x55e8a29a8e20, setby=0x55e8a29a8bc1 "baz", type=43 '+', mode=111 'o',
>     arg=0x55e8a29ae963 ":bar", newmode=0x55e8a2857200) at modes.c:301
> #1  0x000055e8a0e137dd in parse_channel_modes (channel=0x55e8a29a8e20, setby=setby@entry=0x55e8a29a8bc1 "baz",
>     mode=mode@entry=[…] "+o :bar", update_key=update_key@entry=1) at modes.c:372
> #2  0x000055e8a0e13f1d in event_mode (server=0x55e8a29a6880, data=<optimized out>, nick=0x55e8a29a8bc1 "baz")
>     at modes.c:498
> #3  0x000055e8a0e3de92 in signal_emit_real (rec=rec@entry=0x55e8a286d690, params=params@entry=4, va=va@entry=0x7fffafbae100,
>     first_hook=<optimized out>) at signals.c:242
> ```
> 
> AFAICT, irssi’s [`parse_channel_modes()`](https://github.com/irssi/irssi/blob/22ce6637c7acbdda27395521cc7dce23fa98b8b9/src/irc/core/modes.c#L337-L410) function just doesn’t expect any of the parameters to contain the prefix (`:`) character.
> 

event_mode:

```
	params = event_get_params(data, 2 | PARAM_FLAG_GETREST,
				  &channel, &mode);
```

event_get_params:

```c
	rest = count & PARAM_FLAG_GETREST;
	count = PARAM_WITHOUT_FLAGS(count);

	while (count-- > 0) {
		str = (char **) va_arg(args, char **);
		if (count == 0 && rest) {
			/* put the rest to last parameter */
			tmp = *datad == ':' ? datad+1 : datad;
```

So `:` is only removed if it's in `MODE #foo :+o user`, not in `MODE #foo +o :user`.

It's technically valid rfc1459 framing so we have to fix it.

horgh:
This is to fix #601.

I tested this with the added unit tests, as well as with an IRCD modified to always prefix the last parameter to MODE with a ":", whether needed or not.

### DIFF
--- a/src/fe-common/irc/fe-events.c
+++ b/src/fe-common/irc/fe-events.c
@@ -224,7 +224,7 @@ static void event_nick(IRC_SERVER_REC *server, const char *data,
 static void event_mode(IRC_SERVER_REC *server, const char *data,
 		       const char *nick, const char *addr)
 {
-	char *params = NULL, *channel = NULL, *mode = NULL;
+	char *params, *channel, *mode;
 
 	g_return_if_fail(data != NULL);
 
--- a/src/fe-common/irc/fe-events.c
+++ b/src/fe-common/irc/fe-events.c
@@ -224,7 +224,7 @@ static void event_nick(IRC_SERVER_REC *server, const char *data,
 static void event_mode(IRC_SERVER_REC *server, const char *data,
 		       const char *nick, const char *addr)
 {
-	char *params, *channel, *mode;
+	char *params = NULL, *channel = NULL, *mode = NULL;
 
 	g_return_if_fail(data != NULL);
 
--- a/src/irc/core/irc.c
+++ b/src/irc/core/irc.c
@@ -40,6 +40,8 @@ static int signal_server_incoming;
 #  define MAX_SOCKET_READS 5
 #endif
 
+static void strip_params_colon(char *const);
+
 /* The core of the irc_send_cmd* functions. If `raw' is TRUE, the `cmd'
    won't be checked at all if it's 512 bytes or not, or if it contains
    line feeds or not. Use with extreme caution! */
@@ -269,8 +271,9 @@ char *event_get_params(const char *data, int count, ...)
 	while (count-- > 0) {
 		str = (char **) va_arg(args, char **);
 		if (count == 0 && rest) {
-			/* put the rest to last parameter */
-			tmp = *datad == ':' ? datad+1 : datad;
+			/* Put the rest into the last parameter. */
+			strip_params_colon(datad);
+			tmp = datad;
 		} else {
 			tmp = event_get_param(&datad);
 		}
@@ -279,6 +282,31 @@ char *event_get_params(const char *data, int count, ...)
 	va_end(args);
 
 	return duprec;
+}
+
+/* Given a string containing <params>, strip any colon prefixing <trailing>. */
+static void strip_params_colon(char *const params)
+{
+	if (!params) {
+		return;
+	}
+
+	char *s = params;
+	while (*s != '\0') {
+		if (*s == ':') {
+			memmove(s, s+1, strlen(s+1)+1);
+			return;
+		}
+
+		s = strchr(s, ' ');
+		if (!s) {
+			return;
+		}
+
+		while (*s == ' ') {
+			s++;
+		}
+	}
 }
 
 static void irc_server_event(IRC_SERVER_REC *server, const char *line,
--- a/src/irc/core/irc.c
+++ b/src/irc/core/irc.c
@@ -287,11 +287,13 @@ char *event_get_params(const char *data, int count, ...)
 /* Given a string containing <params>, strip any colon prefixing <trailing>. */
 static void strip_params_colon(char *const params)
 {
-	if (!params) {
+	char *s;
+
+	if (params == NULL) {
 		return;
 	}
 
-	char *s = params;
+	s = params;
 	while (*s != '\0') {
 		if (*s == ':') {
 			memmove(s, s+1, strlen(s+1)+1);
@@ -299,7 +301,7 @@ static void strip_params_colon(char *const params)
 		}
 
 		s = strchr(s, ' ');
-		if (!s) {
+		if (s == NULL) {
 			return;
 		}
 
--- a/src/irc/core/modes.c
+++ b/src/irc/core/modes.c
@@ -480,8 +480,8 @@ static void event_user_mode(IRC_SERVER_REC *server, const char *data)
 static void event_mode(IRC_SERVER_REC *server, const char *data,
 		       const char *nick)
 {
-	IRC_CHANNEL_REC *chanrec = NULL;
-	char *params = NULL, *channel = NULL, *mode = NULL;
+	IRC_CHANNEL_REC *chanrec;
+	char *params, *channel, *mode;
 
 	g_return_if_fail(data != NULL);
 
--- a/src/irc/core/modes.c
+++ b/src/irc/core/modes.c
@@ -480,8 +480,8 @@ static void event_user_mode(IRC_SERVER_REC *server, const char *data)
 static void event_mode(IRC_SERVER_REC *server, const char *data,
 		       const char *nick)
 {
-	IRC_CHANNEL_REC *chanrec;
-	char *params, *channel, *mode;
+	IRC_CHANNEL_REC *chanrec = NULL;
+	char *params = NULL, *channel = NULL, *mode = NULL;
 
 	g_return_if_fail(data != NULL);
 


horgh:
Travis doesn't like the addition of trying to run `make check`. I'm working on it!

ailin-nemui:
we should discuss a bit more detailed how to best add tests into irssi

horgh:
Sure! I'm happy to change this. Mainly I wanted tests to try to understand what was going on with the relevant functions and this was an easy way.

These tests use automake's [test suite functionality](https://www.gnu.org/software/automake/manual/automake.html#Simple-Tests). It is pretty flexible and there can be several test programs that will run. Since the one added here is a regular C program, we can also run it under valgrind and gdb and the like.

Another way I sometimes like to write tests like these is to have the test code in the actual source files guarded by ifdefs, and conditionally compile a main function in there. That way static functions can be tested.

What do you think? I'm not sure of the best way that would work here. We could consider test suite libraries/frameworks as well, if we wanted to think about making tests easier to write.

ailin-nemui:
(I opened a separate discussion about this)

src/irc/core/modes.c
```diff
@@ -498,7 +500,82 @@ static void event_mode(IRC_SERVER_REC *server, const char *data,
...
+	params_ptr = params_copy;
+
+	// Get the target.
+
+	while (*params_ptr == ' ') {
+		params_ptr++;
+	}
+
+	s = event_get_param(&params_ptr);
+	if (s == NULL || strlen(s) == 0) {
```
dequis: `s == NULL || *s == '\0'`

src/irc/core/modes.c
```diff
@@ -498,7 +500,82 @@ static void event_mode(IRC_SERVER_REC *server, const char *data,
...
+// correct, but it has historically worked.
+//
+// For example:
+//
+// "#channel +o test" gets parsed to: target = "#channel" and modes = "+o test"
+// "#channel :+o test" gets parsed to: target = "#channel" and modes = "+o test"
+// "#channel +o :test" gets parsed to: target = "#channel" and modes = "+o test"
+//
+// The caller must free the target and modes strings.
+void get_mode_params(char const * const params, char * * target, char * * modes)
```
dequis: `(char const *const params, char **target, char **modes)`

src/irc/core/modes.c
```diff
@@ -498,7 +500,82 @@ static void event_mode(IRC_SERVER_REC *server, const char *data,
...
+{
+	char * params_copy = NULL;
+	char * params_ptr = NULL;
+	char * s = NULL;
+	char * mode_buf = NULL;
```
dequis: `char *mode_buf`

src/irc/core/modes.c
```diff
@@ -498,7 +500,82 @@ static void event_mode(IRC_SERVER_REC *server, const char *data,
...
+// position of <trailing> is not particularly important. This is not very RFC
+// correct, but it has historically worked.
+//
+// For example:
+//
+// "#channel +o test" gets parsed to: target = "#channel" and modes = "+o test"
+// "#channel :+o test" gets parsed to: target = "#channel" and modes = "+o test"
+// "#channel +o :test" gets parsed to: target = "#channel" and modes = "+o test"
+//
+// The caller must free the target and modes strings.
```
dequis: A block level comment would make sense. The code style generally prefers `/* */` comments everywhere, even if it's single line

src/irc/core/modes.c
```diff
@@ -498,7 +500,82 @@ static void event_mode(IRC_SERVER_REC *server, const char *data,
...
+
+		s = event_get_param(&params_ptr);
+		if (s == NULL || strlen(s) == 0) {
+			break;
+		}
+
+		if (strlen(mode_buf) > 0) {
+			strcat(mode_buf, " ");
+		}
+		strcat(mode_buf, s);
```
dequis: Use GString instead of g_malloc0 / strcat / etc

https://developer.gnome.org/glib/stable/glib-Strings.html

LemonBoy:
If I correctly understand what #601 is about all we have to do is preprocess the `rest` string to make sure an eventual  `:`  is weeded out: this leads to a much simpler fix IMO that only needs a `memmove`. Maybe not even that since, after a quick glance at the code, the whitespace is ignored when the `mode` argument is parsed, you could just replace the `:` with a ` ` and call it a day.

Just my 0.02€

ailin-nemui:
let's investigate if there is a simpler solution àla LemonBoy's suggestion, and move the test to a separate PR. As per #767 I would like to see something based on glib test for that in tests/irc

horgh:
Thanks for the reviews. I force pushed stripping the ":" as suggested by @LemonBoy. I opted to strip it rather than replace it with a space as this function is used for more than mode parsing, and I don't know that all uses would ignore it. I'll move the tests to a separate PR.

dequis:
This will just look for any : anywhere and remove it. Colons are only meaningful at the beginning of a parameter, not in the middle.

horgh:
Hmm yeah, good catch. For some reason I thought we needed to escape parameters containing them with a : prefix, but apparently not!

horgh:
I force pushed an update.

src/fe-common/irc/fe-events.c
```diff
@@ -224,7 +224,7 @@ static void event_nick(IRC_SERVER_REC *server, const char *data,
 static void event_mode(IRC_SERVER_REC *server, const char *data,
 		       const char *nick, const char *addr)
 {
-	char *params, *channel, *mode;
+	char *params = NULL, *channel = NULL, *mode = NULL;
```
ailin-nemui: for the moment we try to stay c89 compatible so we cannot initialise them here

src/irc/core/modes.c
```diff
@@ -480,8 +480,8 @@ static void event_user_mode(IRC_SERVER_REC *server, const char *data)
 static void event_mode(IRC_SERVER_REC *server, const char *data,
 		       const char *nick)
 {
-	IRC_CHANNEL_REC *chanrec;
-	char *params, *channel, *mode;
+	IRC_CHANNEL_REC *chanrec = NULL;
+	char *params = NULL, *channel = NULL, *mode = NULL;
```
ailin-nemui: see above ..

src/fe-common/irc/fe-events.c
horgh: Ah okay. I've fixed that. Thanks for looking at this!

src/irc/core/irc.c
```diff
@@ -281,6 +284,31 @@ char *event_get_params(const char *data, int count, ...)
 	return duprec;
 }
 
+/* Given a string containing <params>, strip any colon prefixing <trailing>. */
+static void strip_params_colon(char *const params)
+{
+	if (!params) {
```
ailin-nemui: now we just need to get your code into the same "style" as irssi. instead of !var we have to use `var ==  NULL`

src/irc/core/irc.c
```diff
@@ -281,6 +284,31 @@ char *event_get_params(const char *data, int count, ...)
...
+
+	char *s = params;
+	while (*s != '\0') {
+		if (*s == ':') {
+			memmove(s, s+1, strlen(s+1)+1);
+			return;
+		}
+
+		s = strchr(s, ' ');
+		if (!s) {
```
ailin-nemui: .

src/irc/core/irc.c
```diff
@@ -281,6 +284,31 @@ char *event_get_params(const char *data, int count, ...)
...
 }
 
+/* Given a string containing <params>, strip any colon prefixing <trailing>. */
+static void strip_params_colon(char *const params)
+{
+	if (!params) {
+		return;
+	}
+
+	char *s = params;
```
ailin-nemui: and declar char *s on top, assign here

src/irc/core/irc.c
horgh: Ah yeah. I will update all these. Thanks! Having clang-format will be nice!

src/fe-common/irc/fe-events.c
dequis: FTR this is actually legal under c89 https://godbolt.org/g/9S3BWR

(but no need to change it back)

ailin-nemui:
@irssi/developers

Issue #745, opened 2017-09-08
/lastlog -clear breaks /clear


ailin-nemui:
if there are empty_linecount on the screen, and you try to `/lastlog -clear`, the screen is scrolled to bottom instead of remaining steady

reported by @koffafi

PR #768, opened 2017-10-09
wip

ailin-nemui:


rbisewski:
I've created a patch in my fork of irssi that seems to resolve issue #745, at least on tests I've ran with it on my docker image.

https://github.com/rbisewski/irssi/commit/11883133a03ef85c024da0b0aa860d15bdc657fc

This seems to work well for very large logs; though there is the limitation that if the user is in the middle of the outputted last log (i.e. between LASTLOG and END OF LASTLOG), it will still scroll them to bottom of the remaining output not cleared.

The reason for this is I think this is an improvement over leaving the user with the false impression of a blank screen and it completely solves the issue for larger logs. The alternative would require some additional logic and add complexity to the code, so in my opinion this is the elegant solution.

Let me know what you think, then consider running some tests on it to ensure there are no regressions due to this change, and afterwards merge it in if you are satisfied with this.

PR #769, opened 2017-10-09
Set host to an empty string on error

horgh:
While investigating #317, I noticed that it was possible we would access
an uninitialized buffer due to failing to check the return value of
net_ip2host(). This is done in several places. To make such uses safe,
set the host buffer to an empty string on error. It is possible callers
could be improved by handling the error in each spot, but this gives us
some safety.

### DIFF
--- a/src/core/network.c
+++ b/src/core/network.c
@@ -489,7 +489,16 @@ int net_gethostbyaddr(IPADDR *ip, char **name)
 
 int net_ip2host(IPADDR *ip, char *host)
 {
-	return inet_ntop(ip->family, &ip->ip, host, MAX_IP_LEN) ? 0 : -1;
+	if (inet_ntop(ip->family, &ip->ip, host, MAX_IP_LEN)) {
+		return 0;
+	}
+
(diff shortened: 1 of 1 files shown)


josephbisch:
LGTM

src/core/network.c
```diff
@@ -489,7 +489,16 @@ int net_gethostbyaddr(IPADDR *ip, char **name)
 
 int net_ip2host(IPADDR *ip, char *host)
 {
-	return inet_ntop(ip->family, &ip->ip, host, MAX_IP_LEN) ? 0 : -1;
+	if (inet_ntop(ip->family, &ip->ip, host, MAX_IP_LEN)) {
+		return 0;
+	}
+
+	// For callers that do not check our return value and pass in an
```
LemonBoy: The comment is quite out of place and is better suited to the commit message.
I'd just add `host[0] = '\0'` before the `return inet_ntop` and be done with this.

horgh:
Thanks for checking it out! I updated the change. A lot simpler!

PR #770, opened 2017-10-09
Remove a couple unused functions

horgh:
I noticed an unused function and then saw another. I see one was already marked deprecated, so maybe there's a reason we should keep it, but anyway.

### DIFF
--- a/src/core/network.c
+++ b/src/core/network.c
@@ -60,18 +60,6 @@ IPADDR ip4_any = {
 #endif
 };
 
-int net_ip_compare(IPADDR *ip1, IPADDR *ip2)
-{
-	if (ip1->family != ip2->family)
-		return 0;
-
-	if (ip1->family == AF_INET6)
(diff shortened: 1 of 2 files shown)


LemonBoy:
You should deprecate first, then remove, but not in a minor release.

ailin-nemui:
I think we can accept this into master..

ailin-nemui:
let's remove the deprecated one, the ip compare I don't really care.. could be useful but I don't see it used anywhere

dequis:
Oh right one of them is deprecated already. Looks like that was part of #422 included in 1.0.0. I'm ok with removing that.

net_ip_compare is small, doesn't hurt. But yeah, useless. Looks like timo sirainen copypasting his own code from dovecot.

horgh:
I reverted removing the one that wasn't already marked deprecated, and marked it deprecated.

PR #773, opened 2017-10-15
Correcting lastlog logic to prevent shifting down after a log clear

rbisewski:
As per my comments in this thread... https://github.com/irssi/irssi/pull/768

I think this might be a solution to the mentioned issue #745.

### DIFF
--- a/src/fe-text/textbuffer-view.c
+++ b/src/fe-text/textbuffer-view.c
@@ -1237,6 +1237,10 @@ void textbuffer_view_remove_line(TEXT_BUFFER_VIEW_REC *view, LINE_REC *line)
 	textbuffer_remove(view->buffer, line);
 }
 
+// TODO: this works better than the existing /lastlog logic, however
+//       this fails when the log is *much* longer than the screen size,
+//       so additional logic will be need to handle that; specifically
+//       consider simply reusing logic from the /clear function
 void textbuffer_view_remove_line_from_lastlog(TEXT_BUFFER_VIEW_REC *view, LINE_REC *line)
 {
 	unsigned char update_counter;
--- a/src/fe-text/textbuffer-view.c
+++ b/src/fe-text/textbuffer-view.c
@@ -21,6 +21,7 @@
 #define	G_LOG_DOMAIN "TextBufferView"
 
 #include "module.h"
+#include "levels.h"
 #include "textbuffer-view.h"
 #include "signals.h"
 #include "utf8.h"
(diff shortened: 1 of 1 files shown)


dequis:
[Chesterton's fence](https://en.wikipedia.org/wiki/G._K._Chesterton#Chesterton.27s_fence)

rbisewski:
I was under the impression the main error was the scrolling to the bottom of the screen after attempting to do a "/lastlog -clear" command. This does solve the use case when the user is viewing the "LASTLOG" section line.

Perhaps a more detailed description of the problem might be shed some light on this bug. Can you or some member of the team offer additional information?

ailin-nemui:
hi, thanks for your investigation. Unless you have an explanation for why removing that line would fix the issue, I'm afraid I cannot yet believe its simplicity.

Let us take a look at the issue again:
/echo abc
/clear
-> screen is empty
/lastlog abc
-> screen shows Lastlog
/lastlog -clear
-> screen should be *empty*, but instead it shows the previous lines

I applied your patch but did not observe any change in this regard. I believe this is what you noted yourself when you said it doesn't work if the user is in the "middle" of a lastlog.

In fact I don't understand in which case your patch *does* have an effect, care to elaborate?

rbisewski:
Thanks for elaborating the issue in more detail :)

Specifically the case I had in mind was:

a) if the user has a very large log (e.g. "/echo abc" several dozen times) 

b) scroll up a number of lines

c) /lastlog (which prints the log at the very bottom)

d) /lastlog -clear (which removes the log at the bottom)

e) the user's screen is not forced to the bottom

Perhaps not that common a use case.

For clarification... if the user does a "/lastlog -clear" did you want the screen to be completely empty, correct? Or should only the lines between "Lastlog" and "End of Lastlog" be cleared?

In any event, I'll take a closer look at the issue later and see if I can create a proper fix for this.

ailin-nemui:
I think it's not an easy fix (but that shouldn't stop you from proving me wrong!)

I do not see any difference if I do your step c) and d), with or without your change... maybe I'm stupid, quite possible

the idea is that /lastlog -clear should remote the log

if the screen "looked empty" (for example due to /clear), then after /lastlog and /lastlog -clear, it should look empty again

if the screen had 3 lines visible (for example /clear, /echo 1, /echo 2, /echo 3), then after /lastlog and /lastlog -clear, it should have 1-3 visible again

ailin-nemui:
by the way I think this is completely unrelated to **lastlog** itself, it's just an easy way to demonstrate the issue. The same issue probably exists with /scrollback levelclear -level XXX and also the trackbar script (if you /mark after /clear, then switch windows, and come back, the window is not empty anymore)

rbisewski:
Alright, I have what appears to be a solution to the use case mentioned by @ailin-nemui above.

https://github.com/rbisewski/irssi/commit/ba65b6308dbc9a2699e16e9ea2731892828460ba

Some basic testing seems to suggest it is fine, but additional testing could be worthwhile.

Finally, I can re-open request or make a new one, if you like.

rbisewski:
@ailin-nemui regarding your comment about the scrollback function, I think you probably bring up a good point. My current thinking is to try to get a functional piece of logic and document why in great detail once I have figured it out. That way, future development of this code ought to be easier :)

Anyway, I did some further testing and it is still a bit glitchy when the lastlog is quite large. So I'll look into that later and hopefully have a decent solution.

ailin-nemui:
we are closing this one for inactivity, but do not hesitate to reopen if you succeed in making progress!

PR #775, opened 2017-10-21
CAP 3.2 support

LemonBoy:
~~This is a prerequisite for the IRC v3.2 compliance.~~
Fuck this shit, I've accidentally implemented the whole thing.

This has been tested this with Freenode and Unrealircd and found no problem with the old 3.1 flow and Valgrind looks happy with this.
I've tested the multiline thing using netcat and the strings found on the ircv3 page and the `ADD`/`DEL` things are untested as of now (I'd like to talk with a real ircd instead of netcat...)
I'm not sure about the XS part as it kept loading the `Irssi.pm` module off `/usr` so feel free to give it a shot.

### DIFF
--- a/src/core/misc.c
+++ b/src/core/misc.c
@@ -218,6 +218,19 @@ GSList *gslist_remove_string (GSList *list, const char *str)
 	return list;
 }
 
+GSList *gslist_delete_string (GSList *list, const char *str, GDestroyNotify free_func)
+{
+	GSList *l;
+
+	l = g_slist_find_custom(list, str, (GCompareFunc) g_strcmp0);
+	if (l != NULL) {
+		free_func(l->data);
+		return g_slist_delete_link(list, l);
+	}
+
+	return list;
+}
+
 /* `list' contains pointer to structure with a char* to string. */
 char *gslistptr_to_string(GSList *list, int offset, const char *delimiter)
 {
--- a/src/irc/core/irc-cap.c
+++ b/src/irc/core/irc-cap.c
@@ -140,7 +140,7 @@ static void event_cap (IRC_SERVER_REC *server, char *args, char *nick, char *add
 	caps = g_strsplit(g_strchomp(list), " ", -1);
 	caps_length = g_strv_length(caps);
 
-	if (!g_strcmp0(evt, "LS")) {
+	if (!strcmp(evt, "LS")) {
 		if (server->cap_supported) {
 			g_hash_table_destroy(server->cap_supported);
 		}
@@ -204,7 +204,7 @@ static void event_cap (IRC_SERVER_REC *server, char *args, char *nick, char *add
 			}
 		}
 	}
-	else if (!g_strcmp0(evt, "ACK")) {
+	else if (!strcmp(evt, "ACK")) {
 		int got_sasl = FALSE;
 
 		/* Emit a signal for every ack'd cap */
@@ -216,7 +216,7 @@ static void event_cap (IRC_SERVER_REC *server, char *args, char *nick, char *add
 			else
 				server->cap_active = g_slist_prepend(server->cap_active, g_strdup(caps[i]));
 
-			if (!g_strcmp0(caps[i], "sasl"))
+			if (!strcmp(caps[i], "sasl"))
 				got_sasl = TRUE;
 
 			cap_emit_signal(server, "ack", caps[i]);
@@ -228,7 +228,7 @@ static void event_cap (IRC_SERVER_REC *server, char *args, char *nick, char *add
 		if (got_sasl == FALSE)
 			cap_finish_negotiation(server);
 	}
-	else if (!g_strcmp0(evt, "NAK")) {
+	else if (!strcmp(evt, "NAK")) {
 		g_warning("The server answered with a NAK to our CAP request, this should not happen");
 
 		/* A NAK'd request means that a required cap can't be enabled or disabled, don't update the
--- a/src/irc/core/irc-cap.c
+++ b/src/irc/core/irc-cap.c
@@ -148,8 +148,12 @@ static void event_cap (IRC_SERVER_REC *server, char *args, char *nick, char *add
 	caps_length = g_strv_length(caps);
 
 	if (!g_ascii_strcasecmp(evt, "LS")) {
-		/* Throw away everything and start from scratch */
-		g_hash_table_remove_all(server->cap_supported);
+		if (!server->cap_in_multiline) {
+			/* Throw away everything and start from scratch */
+			g_hash_table_remove_all(server->cap_supported);
+		}
+
+		server->cap_in_multiline = multiline;
 
 		/* Create a list of the supported caps */
 		for (i = 0; i < caps_length; i++) {
--- a/src/irc/core/irc-cap.c
+++ b/src/irc/core/irc-cap.c
@@ -147,7 +147,7 @@ static void event_cap (IRC_SERVER_REC *server, char *args, char *nick, char *add
 	caps = g_strsplit(g_strchomp(list), " ", -1);
 	caps_length = g_strv_length(caps);
 
-	if (!strcmp(evt, "LS")) {
+	if (!g_ascii_strcasecmp(evt, "LS")) {
 		/* Throw away everything and start from scratch */
 		g_hash_table_remove_all(server->cap_supported);
 
@@ -204,7 +204,7 @@ static void event_cap (IRC_SERVER_REC *server, char *args, char *nick, char *add
 			}
 		}
 	}
-	else if (!strcmp(evt, "ACK")) {
+	else if (!g_ascii_strcasecmp(evt, "ACK")) {
 		int got_sasl = FALSE;
 
 		/* Emit a signal for every ack'd cap */
@@ -228,7 +228,7 @@ static void event_cap (IRC_SERVER_REC *server, char *args, char *nick, char *add
 		if (got_sasl == FALSE)
 			cap_finish_negotiation(server);
 	}
-	else if (!strcmp(evt, "NAK")) {
+	else if (!g_ascii_strcasecmp(evt, "NAK")) {
 		g_warning("The server answered with a NAK to our CAP request, this should not happen");
 
 		/* A NAK'd request means that a required cap can't be enabled or disabled, don't update the
@@ -236,7 +236,7 @@ static void event_cap (IRC_SERVER_REC *server, char *args, char *nick, char *add
 		for (i = 0; i < caps_length; i++)
 			cap_emit_signal(server, "nak", caps[i]);
 	}
-	else if (!strcmp(evt, "NEW")) {
+	else if (!g_ascii_strcasecmp(evt, "NEW")) {
 		for (i = 0; i < caps_length; i++) {
 			char *key, *val;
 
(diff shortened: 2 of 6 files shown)


ailin-nemui:
is this finished already or still wip?

LemonBoy:
I guess it can be reviewed, yeah

src/irc/core/irc-cap.c
```diff
@@ -79,73 +79,144 @@ static void cap_emit_signal (IRC_SERVER_REC *server, char *cmd, char *args)
...
+				else
+					cap_finish_negotiation(server);
 
-			g_string_free(cmd, TRUE);
+				g_string_free(cmd, TRUE);
+			}
 		}
 	}
-	else if (!g_strcmp0(evt, "ACK")) {
+	else if (!strcmp(evt, "ACK")) {
```
dequis: Same as above

src/perl/irc/Irc.xs
```diff
@@ -34,10 +37,14 @@ static void perl_irc_server_fill_hash(HV *hv, IRC_SERVER_REC *server)
...
-		av_push(av, new_pv(tmp->data));
-	(void) hv_store(hv, "cap_supported", 13, newRV_noinc((SV*)av), 0);
+	hv_ = newHV();
+	g_hash_table_iter_init(&iter, server->cap_supported);
+	while (g_hash_table_iter_next(&iter, &key_, &val_)) {
+		char *key = (char *)key_;
+		char *val = (char *)val_;
+		hv_store(hv_, key, strlen(key), new_pv(val), 0);
+	}
+	(void) hv_store(hv, "cap_supported", 13, newRV_noinc((SV*)hv_), 0);
```
dequis: So this is a perl api change, right? Does perl handle this decently enough, falling back to keys only if it is accessed like an array instead of a hash? @ailin-nemui

src/irc/core/irc-cap.c
```diff
@@ -79,73 +79,144 @@ static void cap_emit_signal (IRC_SERVER_REC *server, char *cmd, char *args)
...
+
 	/* Strip the trailing whitespaces before splitting the string, some servers send responses with
 	 * superfluous whitespaces that g_strsplit the interprets as tokens */
 	caps = g_strsplit(g_strchomp(list), " ", -1);
 	caps_length = g_strv_length(caps);
 
-	if (!g_strcmp0(evt, "LS")) {
+	if (!strcmp(evt, "LS")) {
+		/* Throw away everything and start from scratch */
+		g_hash_table_remove_all(server->cap_supported);
```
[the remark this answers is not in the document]
LemonBoy: Good catch, I moved some code and forgot to re-test with multiline responses.
The aim was to make the parsing stateless but I think the best way to handle this case is to keep track of the completeness of the response in a(nother) flag in the `IRC_SERVER_CONNECT_REC`

src/perl/irc/Irc.xs
ailin-nemui: no that would not work..

ailin-nemui:
what's the status here @LemonBoy @dequis

LemonBoy:
It's missing another round of review, an idea about what/how to expose those new fields to the perl side and a use case :)

src/core/misc.c
```diff
@@ -218,6 +218,19 @@ GSList *gslist_remove_string (GSList *list, const char *str)
 	return list;
 }
 
+GSList *gslist_delete_string (GSList *list, const char *str, GDestroyNotify free_func)
```
ailin-nemui: you can remove the remove_string method, which is apparently flawed

src/irc/core/irc-cap.c
```diff
@@ -79,73 +79,148 @@ static void cap_emit_signal (IRC_SERVER_REC *server, char *cmd, char *args)
...
+	g_return_val_if_fail(name != NULL, FALSE);
+	g_return_val_if_fail(name[0] != '\0', FALSE);
+
+	const char *eq = strchr(name, '=');
+	/* KEY only value */
+	if (eq == NULL) {
+		*key = g_strdup(name);
+		*val = NULL;
+		return TRUE;
+	}
```
ailin-nemui: I also like this code style but until we get the fully automated code formatter working please adjust to `} else if {` on a single line

src/irc/core/irc-cap.c
```diff
@@ -79,73 +79,148 @@ static void cap_emit_signal (IRC_SERVER_REC *server, char *cmd, char *args)
...
+	/* Some values are in a KEY=VALUE form, parse them */
+	else if (eq[1] != '\0') {
+		*key = g_strndup(name, (gsize)(eq - name));
+		*val = g_strdup(eq + 1);
+		return TRUE;
+	}
+	/* If the string ends after the '=' consider the value
+	 * as invalid */
+	else {
```
ailin-nemui: does the spec not allow empty values?

src/irc/core/irc-servers.c
```diff
@@ -440,8 +440,10 @@ static void sig_disconnected(IRC_SERVER_REC *server)
 	gslist_free_full(server->cap_active, (GDestroyNotify) g_free);
 	server->cap_active = NULL;
 
-	gslist_free_full(server->cap_supported, (GDestroyNotify) g_free);
-	server->cap_supported = NULL;
+	if (server->cap_supported) {
```
ailin-nemui: indentation is off here. you can also try our new `git-clang-format`

src/perl/irc/Irc.xs
ailin-nemui: (but I guess we can break the API)

ailin-nemui:
would be nice if you can reply to all open comments @LemonBoy then we can merge this.

src/irc/core/irc-cap.c
LemonBoy: That's a good question, the spec states the format is `<name>[=<value>]` with no further indications.

src/perl/irc/Irc.xs
LemonBoy: My (wild) guess is that we have no users of this API heh
I'd avoid such a breaking change in a non-point release though, got a better idea that's backward compatible?

ailin-nemui: well, I can't think of any way other than adding another key, say cap_supported_hash or another method, since we initially put it it all in an array. but that seems silly imo

ailin-nemui:
(and fix the errors caused by -Werror=declaration-after-statement, see travis)

ailin-nemui:
@irssi/developers tagging this for merge

dequis:
Hm, haven't looked at this one in a while (silly rebases making it hard to tell which was the last commit i reviewed), I'll look again over the weekend.

src/core/misc.c
```diff
@@ -207,13 +207,15 @@ void gslist_free_full (GSList *list, GDestroyNotify free_func)
 	g_slist_free(list);
 }
 
-GSList *gslist_remove_string (GSList *list, const char *str)
+GSList *gslist_delete_string (GSList *list, const char *str, GDestroyNotify free_func)
```
dequis: `gslist_remove_string` was API and is still in misc.h

src/irc/core/irc-cap.c
```diff
@@ -79,73 +79,148 @@ static void cap_emit_signal (IRC_SERVER_REC *server, char *cmd, char *args)
...
+		*key = g_strdup(name);
+		*val = NULL;
+		return TRUE;
+	/* Some values are in a KEY=VALUE form, parse them */
+	} else if (eq[1] != '\0') {
+		*key = g_strndup(name, (gsize)(eq - name));
+		*val = g_strdup(eq + 1);
+		return TRUE;
+	/* If the string ends after the '=' consider the value
+	 * as invalid */
```
dequis: I dunno about this, a blank value could be meaningful. The spec doesn't consider this case but it doesn't say it should be rejected either.

src/irc/core/irc-cap.c
```diff
@@ -79,73 +79,148 @@ static void cap_emit_signal (IRC_SERVER_REC *server, char *cmd, char *args)
...
 			disable = (*caps[i] == '-');
 
 			if (disable)
-				server->cap_active = gslist_remove_string(server->cap_active, caps[i] + 1);
+				server->cap_active = gslist_delete_string(server->cap_active, caps[i] + 1, g_free);
 			else
 				server->cap_active = g_slist_prepend(server->cap_active, g_strdup(caps[i]));
 
-			if (!g_strcmp0(caps[i], "sasl"))
+			if (!strcmp(caps[i], "sasl"))
```
dequis: This change is just increasing the crashiness lol.

LemonBoy: Not quite as the right hand side is a constant string and the left hand side comes from a null-terminated vector of strings, I used the plain `strcmp` because of that.

src/irc/core/irc-cap.c
LemonBoy: My reasoning was that you either provide a value or you don't, the specification isn't very clear with it's `name[=value]` description and I'd suggest to raise this issue with the irc3x folks.

src/core/misc.c
LemonBoy: Indeed, but the old one was wrong and potentially leaky.
We could re-add it in a correct form using `g_slist_delete_link` instead or re-write it in terms of `gslist_delete_string` (with a `NULL` `free_func`, with minor modifications to `gslist_delete_string`).

src/irc/core/irc-cap.c
dequis: There's this: https://github.com/ircv3/ircv3-specifications/issues/334 - it's actually about message-tags, but people seem to agree that the solution should be the same for cap values too.

I like the idea of parsing `a=` as `key="a" value=""` (and `a` as `key="a" value=NULL`), it's fairly natural, so, just remove the special case imo

FWIW, dan- is working on a rewrite of the cap specs to make the whole thing less confusing, he's aware of this now.

src/irc/core/irc-cap.c
dequis: It just seemed pointless to bother changing it, but sure, it's harmless.

LemonBoy:
Done, I've addressed all the review comments. I think the whole PR should be squashed before merging it.

src/core/misc.c
dequis: lol this comment came back

ailin-nemui:
@LemonBoy you still need to remove the remove_string method

LemonBoy:
`remove_string` has been deprecated since @dequis noted that's part of the public API.

dequis:
Yeah in https://github.com/irssi/irssi/pull/775#discussion_r162847636 I basically suggested the opposite of what nei suggested in https://github.com/irssi/irssi/pull/775#pullrequestreview-87082343 (which i hadn't seen before) so adding that function back meant showing that comment again

ailin-nemui:
we need to get rid of broken functions

src/core/misc.c
ailin-nemui: I see, but only LemonBoy was using this api

dequis:
lol we forgot to send `CAP LS 302`

LemonBoy:
The plan was to bump the required version once we had some feature requiring to do so.

PR #776, opened 2017-10-22
Add tests for mode parsing

horgh:
This branches off of the changes in #766. I've moved the test program to a new tests subdirectory, switched to using GLib's testing framework, and changed the tests to be table driven.


ailin-nemui:
can you cherry pick bc059f2 only in this PR?

horgh:
I did that and edited it so tests would pass. By itself it would fail as it was testing the mode parsing PR. If we merge this first I can rebase that PR on it.

ailin-nemui:
it's ok if the tests fail as this means we have a bug....

horgh:
Well if we merged this with it failing, Travis would start failing everywhere! That seems unhelpful. I left TODO comments with what can be updated by the other PR.

PR #780, opened 2017-10-24
Add event_get_params to fe-fuzz

josephbisch:
Probably won't actually find anything. But, this should add a little extra coverage to the fuzzing.

### DIFF
--- a/src/fe-fuzz/irc/core/event-get-params.c
+++ b/src/fe-fuzz/irc/core/event-get-params.c
@@ -52,7 +52,6 @@ int LLVMFuzzerTestOneInput(const uint8_t* data, size_t size) {
 		return 0;
 	}
 	uint8_t count = *data;
-	/* malloc(size) instead of size+1, because we already used one byte of data */
 	gchar *copy = g_strndup((const gchar *)data+1, size-1);
 
 	char *output0;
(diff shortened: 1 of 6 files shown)


src/fe-fuzz/irc/core/event-get-params.c
```diff
@@ -0,0 +1,86 @@
...
+	module_register("core", "fe-fuzz");
+	return 0;
+}
+
+int LLVMFuzzerTestOneInput(const uint8_t* data, size_t size) {
+	if (size < 1) {
+		return 0;
+	}
+	uint8_t count = *data;
+	char *copy = (char *)malloc(sizeof(char)*(size-1+1));
```
dequis: Holy redundancy. Why the size-1+1?

sizeof(char) is 1 unless you're dealing with people who aren't fun at parties.

I wanted to say "No need to cast mallocs" but I'm not actually sure this is C and not C++. Is this C++?

Also, just use g_strndup?

josephbisch: The size-1+1 is because the first byte of data is already used, so we aren't going to use it again for the string. Then plus one for the terminator. Yeah, I could just do size, but I was trying to make the calculation clear.

I don't know why I'm casting the malloc. I copied that from the original fuzzer in fe-fuzz. Maybe at one point fe-fuzz just straight up used CXX for everything, but now it uses CC for compiling and then CXX for linking. So it does work without the cast.

ailin-nemui:
@josephbisch feel free to merge this when you think it's ready...

PR #783, opened 2017-10-29
use includes from irssi

ailin-nemui:
for discussion -- replace all includes with absolute ones; here on the IRC module only


ailin-nemui:
@irssi/developers we need to discuss if we want to do this. The goal would be to remove the -Is, and ease the transition/split into irssi and irssi-irc package which I am targeting for 1.2

ahf:
I have wanted something like this for a very long time, but with a slightly different directory structure:

I would lift all our "library" code into a `irssi.git/irssi/` directory and the `fe-text` code into `src/irssi/`. This way the includes would be `#include <irssi/core/foo.h>` without the `src`.

The problem with doing this is that some plug-ins needs headers in `src/fe-text/*` :-(

ailin-nemui:
I don't think we should break the old ../src/.. for no reason other than looking nicer. my suggestion is to wait until you fix and merge otr and then I will continue this in the fashion I have proposed (unless there are more votes against it or who want it structured like @ahf )

ahf:
Moving things around is not that much of a radical change and we might as well do it together with this while it is being looked at.

ailin-nemui:
we should first get it working without moving, then we can do the moving in a 2nd step, what do you think

ahf:
As long as it's done in a single release window that is fine with me.

PR #788, opened 2017-11-03
Keep a copy of the strings coming from the config

### LINKED ISSUE #787: /hilight generating garbage when passed with multiple -network parameters
EDIT: the bug turns out to be unrelated to the number of `-network` parameters passed, read comments below.

---

I haven't tested whether this happens only with `/hilight` and only with `-network`.

I wanted to set a hilight on two networks, and instead of passing the networks comma-separated in one `-network` parameter, I unwittingly passed two `-network` parameters, and irssi crashed.

_Uh oh_

I reopened irssi and ran the command again, and this time irssi didn't crash, but the echo of the command contained garbage in the `-network` parameter.

Found on irssi v1.0.5

Here I run the same command shown on the input bar a number of times.
2017-11-03-123839_swaygrab

LemonBoy:
Fixes #787, checked with valgrind.


dequis:
The commit message made this sound like a much bigger thing

PR #789, opened 2017-11-04
Change the default awaylog_file to ~/irclogs/away.log.

trasz:
The new location is more compatible with Capsicum sandboxing,
and also seems a bit more logical - it is, after all, a log file.

### DIFF
--- a/docs/capsicum.txt
+++ b/docs/capsicum.txt
@@ -7,7 +7,6 @@ or the libraries it depends on.
 To make Irssi enter capability mode on startup, add
 
 capsicum = "yes";
-awaylog_file = "~/irclogs/away.log";
 
 to your ~/.irssi/config and restart the client.  Alternatively you can
 enter it "by hand", using the "/capsicum enter" command.  From the security
--- a/docs/manual.txt
+++ b/docs/manual.txt
@@ -973,7 +973,7 @@ If <color> is a
 	yourself unaway, Irssi will display all the messages in the awaylog.
 
 	/SET awaylog_level <level> - Default is MSGS HILIGHT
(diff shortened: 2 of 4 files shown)


ailin-nemui:
opinions @dequis ?

dequis:
Idk, changing defaults isn't something to be done lightly. Users thay don't change it get the new value when upgrading from earlier versions and get a change of behavior, potentially breaking.

ailin-nemui:
unlike the real irclogs, the away.log is more of a transient file so I think that's part of the reason being inside the cache folder by default

trasz:
I agree this is not to be done lightly.  However, I think the potential for breaking people's setups is fairly small - precisely because it's transient.

On the other hand - it's not in the cache folder, it's in the ~/.irssi/, which I'd consider to be a configuration folder: it's where the config lives, as well as various scripts.  Personally I'd expect that directory to change relatively infrequently, as opposed to the ~/irclogs/, which change all the time.

ailin-nemui:
I am voting against the change. If irssi were following XDG, I would move the away.log into the cache dir. It doesn't feel right to be part of the to-be-archived logs folder

dkg:
I think if you're moving away.log, it should be moved to an ephemeral filesystem wherever possible, rather than putting it in the user's homedir.  See also #810 and #811 for improvements in actual ephemerality of away.log.

anarcat:
shouldn't capsicum sandboxing follow irssi configs, and not the other way around?

logs in `~/irclogs` have always been stranged to me in the first place as well...

ailin-nemui:
based on the recent discussions surrounding the awaylog file, I have an idea which could work: conditionally set the default based on existence of the away.log file. then we could "safely" move it. thoughts?

PR #790, opened 2017-11-15
reset colour at comma, like mIRC

### LINKED ISSUE #740: Color codes which end in commas both reset background and print comma as literal

ailin-nemui:
Fixes #742 and fixes #740

### DIFF
--- a/src/fe-common/core/formats.c
+++ b/src/fe-common/core/formats.c
@@ -1078,23 +1078,22 @@ static void get_mirc_color(const char **str, int *fg_ret, int *bg_ret)
 		bg = -1;
 	} else {
 		/* foreground color */
-		if (**str != ',') {
-			fg = **str-'0';
+		fg = **str-'0';
+		(*str)++;
+		if (i_isdigit(**str)) {
+			fg = fg*10 + (**str-'0');
 			(*str)++;
-			if (i_isdigit(**str)) {
-				fg = fg*10 + (**str-'0');
-				(*str)++;
-			}
 		}
+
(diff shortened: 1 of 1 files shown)


src/fe-common/core/formats.c
```diff
@@ -1072,7 +1072,8 @@ static void get_mirc_color(const char **str, int *fg_ret, int *bg_ret)
 	fg = fg_ret == NULL ? -1 : *fg_ret;
 	bg = bg_ret == NULL ? -1 : *bg_ret;
 
-	if (!i_isdigit(**str) && **str != ',') {
+	if (!i_isdigit(**str)) {
+		/* turn off color */
 		fg = -1;
 		bg = -1;
 	} else {
```
LemonBoy: The `if (**str != ',') {` check below can be removed because now `**str` is a digit.

ailin-nemui:
thanks for review, removed that check

PR #784, opened 2017-10-30
Turn the style guide into a clang-format file

LemonBoy:
Note that his has been written with clang-format 3.8 in mind since that's what Debian ships, see [here](http://releases.llvm.org/3.8.1/tools/docs/ClangFormatStyleOptions.html) for the docs.

I'm not too fond of the tabs + 100 columns combo since my laptop has a tiny screen, but that's not much of a problem.

So, I hereby declare the style-related bikeshed season open!

PR #791, opened 2017-11-18
Suggestion for code style fix.

fmmedeiros:
Adding brackets to make the relation between else and if statements explicitly. In a case in which indention changes, it is still clear that an specific else belongs to a specific if statement.


dequis:
Are you the same guy as #347?

fmmedeiros:
Hi @dequi, yes, that's me. Please let me this kind of issue makes sense for the project. Thanks

dequis:
We've decided we'll handle all style-related changes through the specification set in #784 instead

PR #795, opened 2017-11-29
check for declaration-after-statement on travis

ailin-nemui:
this will make variable declarations after statements illegal. most of the code follows this practice. in the past, this was required to be able to compile on BeOS iirc

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,7 +31,7 @@ before_install:
 
 install:
     - ./configure --with-proxy --with-bot --with-perl=module --prefix=$HOME/irssi-build
-    - make CFLAGS="-Wall -Werror"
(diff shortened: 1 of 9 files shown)


configure.ac
```diff
@@ -295,6 +295,9 @@ fi
 
 LIBS="$LIBS $GLIB_LIBS"
 
+GLIB_CFLAGS=" $GLIB_CFLAGS"
+GLIB_CFLAGS="${GLIB_CFLAGS// -I/ -isystem}"
```
dequis: wot

ailin-nemui: .... I don't think I intended to commit that.... If you're curious, it is a useful feature, if you want to compile your own code with -pedantic or more warnings, since -isystem will hide warnings in those directories ( as opposed to -I )

PR #802, opened 2018-01-04
Test for segfault on disconnect during signal processing

ailin-nemui:
Tests #796

PR #803, opened 2018-01-04
Postpone server cleanup until after unref

ailin-nemui:
Add a new signal, server destroyed, that is supposed to run the clean up
tasks of server disconnected. This is so that some structures will stay
around longer.

### DIFF
--- a/src/core/servers.c
+++ b/src/core/servers.c
@@ -460,8 +460,6 @@ static int server_remove_channels(SERVER_REC *server)
 
 void server_disconnect(SERVER_REC *server)
 {
-	int chans;
-
 	g_return_if_fail(IS_SERVER(server));
 
 	if (server->disconnected)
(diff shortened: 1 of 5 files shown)


ailin-nemui:
this PR needs to be checked for any adverse side effects that could be caused by adding the additional signal and moving the work

horgh:
If we didn't want to change/add signals, we could alternatively return if `mserver == NULL` in `flood.c` I think. Is it important to run `flood_newmsg()` if the server is disconnected?

ailin-nemui:
yes you're probably right 

lthough I think trying to improve the signals and refcounting is a good overall direction. 

(thanks to your test framework!) try removing https://github.com/irssi/irssi/pull/802/files#diff-3ef538dbbcaf339ceb86081bf85cd41fR118 the fatality increase here, I believe all hash table warnings are bugs. (but maybe they would be fixed by returning from flood_newmsg, I didn't check)

horgh:
Ah yeah. If I try returning early there instead of the fix in this branch, there are a bunch of hash_table_lookup assertion failures and a broken looking message in the status at the time of disconnect:

```
21:38 -!- Irssi: critical g_hash_table_lookup: assertion 'hash_table != NULL' failed
21:38 -!- Irssi: critical g_hash_table_lookup: assertion 'hash_table != NULL' failed
21:38 < :#test> t
21:38 -!- Irssi: critical g_hash_table_lookup: assertion 'hash_table != NULL' failed
21:38 -!- Irssi: critical g_hash_table_lookup: assertion 'hash_table != NULL' failed
```

So the fix here is better!

ailin-nemui:
@irssi/developers tagging this for merge unless you have a better suggestion or would prefer to discuss further

dequis:
I haven't looked at this one at all and it sounds fiddlier than average, give me some time.

ailin-nemui:
@dequis any updates?

ailin-nemui:
@irssi/developers @dequis if you don't want to check this any more I suggest we merge it

dequis:
I poked this for a while and failed to break it, so, yay? Not that i'm particularly inspired today.

Should this be mentioned in `docs/signals.txt`?

ailin-nemui:
I think it's better to leave undocumented for now

PR #805, opened 2018-01-05
fix /exec -o for blank lines

ailin-nemui:
since it is not allowed to send nothing, instead of spamming the status window
with error, send " " instead

Fixes FS#902

Patch by @dj1yfk


ailin-nemui:
@irssi/developers

PR #577, opened 2016-11-23
Minor corrections to the netsplit code.

LemonBoy:
Possibly solving #567

PR #812, opened 2018-01-10
revert netsplit print optimisation

### LINKED ISSUE #809: Netsplits don't show in all windows where users split
The fixes in #465 seem to have caused split users to only show as splitting in certain windows, whereas the expected behavior is to show them in every channel they split from. There appears no such split message at all in some windows. If you experience this same issue please share all the knowledge you have about the situation so this can be fixed.

ailin-nemui:
this reverts part of #465

unfortunately we need to further refine the initial patch

 - when filtering by channel, the whole split is cleaned up nevertheless
 - something similar happens for the netjoins
 - furthermore, we cannot wait only for PUBLIC msgs, j/p/q are equivalently relevant for temporal integrity

this is a temporary band aid until we can properly implement #465 and f.x #420. the basic idea presented therein seems sound

this is intended to fix #809

### DIFF
--- a/src/fe-common/irc/fe-netjoin.c
+++ b/src/fe-common/irc/fe-netjoin.c
@@ -253,15 +253,12 @@ static void sig_print_starting(TEXT_DEST_REC *dest)
 	if (!IS_IRC_SERVER(dest->server))
 		return;
 
-	if (!(dest->level & MSGLEVEL_PUBLIC))
-		return;
-
 	if (!server_ischannel(dest->server, dest->target))
 		return;
 
 	rec = netjoin_find_server(IRC_SERVER(dest->server));
 	if (rec != NULL && rec->netjoins != NULL)
-		print_netjoins(rec, dest->target);
+		print_netjoins(rec, NULL);
 }
 
 static int sig_check_netjoins(void)
--- a/src/fe-common/irc/fe-netsplit.c
+++ b/src/fe-common/irc/fe-netsplit.c
@@ -199,7 +199,7 @@ static void temp_split_chan_free(TEMP_SPLIT_CHAN_REC *rec)
 	g_free(rec);
 }
 
-static void print_splits(IRC_SERVER_REC *server, const char *channel)
+static void print_splits(IRC_SERVER_REC *server, const char *filter_channel)
 {
 	TEMP_SPLIT_REC temp;
 	GSList *servers;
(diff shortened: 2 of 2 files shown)


dequis:
Yeah sounds like the sanest thing to do. Is this going in 1.0.x and 1.1.0?

dequis:
Oh, uh, this needs review too as it's a partial revert so that could bring problems of its own too.

ailin-nemui:
I would include it in a potential 1.0.7 and ideally in 1.1.0

dequis:
So, relevant commits:

* #465 "Some small adjustments to the netsplit code"
    * ed06e43ec816eb4b9f1a736ecba6802a586887a8
    * 8f5e2005516023f5c96cf95ac5e9f917bda33172 "Avoid entering an endless loop while traversing the channel list"
    * c0f66c95ff00bdc22155a8ac53d887a9c1bbc90f ~~"how did we allow ourselves to get this commit message merged"~~
* #577 "Minor corrections to the netsplit code"
    * 7574bed26c98069945b3fbc3938819c10e65b045
    * 5f0e755a00c7cb0e85e167cd455bb3ce16b4d14b "Don't shadow the 'channel' variable when printing the netjoins."

Cherry picked the last two on top of the other three and made this branch for easy diffing of the whole thing:

https://github.com/dequis/irssi/compare/lemon-netsplit-base...dequis:cherry-picked-lemon-netsplit-stuff

Other commits touching fe-netjoin.c or fe-netsplit.c afterwards:

* 98ead50b4eea7cda4b357f1c4b19a741205c6fa4 "Prevent some potential null-pointer deferences" (GL!9)
* 7c09b72848f99886964266ff531b41c69fe138f5 "fe-netjoin: remove irc servers on "server disconnected" signal" (GL#7, GL!10)

Both look like crash fixes so probably not relevant.

dequis:
Things that aren't getting reverted from fe-netsplit:

* The old `sig_print_starting` didn't look at `dest`, and iterated over all servers.
* The `server_ischannel(dest->server, dest->target)`

Things that aren't getting reverted from fe-netjoin:

* The cleaner iteration style, with `g_slist_next` + `g_slist_delete_link`
* Same stuff as above

Mostly harmless stuff. Any particular reason to leave the server_ischannel in? Going back to iterating over all servers might be the safest option too, but that's probably too much and won't change things meaningfully. On the other hand leaving these things in gives us some bits of data, so,

I can't see this making things worse, at least.

ailin-nemui:
thanks for the review / research, maybe we should just completely revert? at least by checking the server 8 hope maybe we can have the trackbar<>netsplit printing fixed, but I couldn't verify that

ailin-nemui:
@irssi/developers

PR #814, opened 2018-01-16
Update Irssi website URLs

l2dy:
Avoid redirects and use https.

```
$ curl -I 'http://www.irssi.org'
HTTP/1.1 302 Moved Temporarily
[...]
Location: https://irssi.org/
```


dequis:
The scripts are synced from https://github.com/irssi/scripts.irssi.org - those local changes will be lost. Revert that part.

l2dy:
@dequis Done.

PR #821, opened 2018-01-19
Allow expandos in /hilight

vague666:


ailin-nemui:
it does sound like a useful feature. one downside I see is that it doesn't seem to support regex? Also I don't know about the overhead costs

vague666:
var expansion in regex would be another option to add, sure, but for (full) expando support I need to send in CHANNEL_REC to hilight_match_text too

ailin-nemui:
we already have /set hilight_nick_matches so I don't sure if we need *this* case ?

vague666:
Using /hilight gives more control and I see quite often questions about the possibility of using expandos in hilights

dequis:
One use case was this person called "day" who uses that nick in one network but other nicks in other networks

So the flexibility of `/set hilight_nick_matches` is desirable but they also wanted to have it limited to `/^day[>:,]/` because starting sentences with `/^day /` is apparently common enough to be annoying.

Regexp support is important for this to be useful for that use case.

The obvious issue is that $ is a regexp special char, so we need some escaping and a way to make this new behavior opt-in, along the lines of `/hilight -expandos`

vague666:
For regex we would have to compile the pattern at runtime, $N for example wouldn't work precompiled, no idea about the performance penalty

ailin-nemui:
I'm not against this but we have to implement the regex case, and that's going to be a bit tricky. it needs a cache that is invalidated based on signals (e.g  own nick change -> invalidate $N regex). also probably should support escaping of expandos inside regexen

vague666:
Can you expand on what you are saying? Why is the invalidation needed?

ailin-nemui:
we should just compile regex once, but for that we need to keep track of when they have to be recompiled because e.g. $N changed

vague666:
Time to revive this, I'm trying to come up with a good strategy for how the signal system should be built up, let's see where we end up :)

vague666:
I guess the question is if all expandos should be supported or just a finite set, I don't see much use for anything but $N, the way the code is now all expandos can be used

PR #827, opened 2018-01-23
add access to set hidden_level from perl…

vague666:
Requires a $view->redraw to update view

### DIFF
--- a/src/perl/textui/TextBufferView.xs
+++ b/src/perl/textui/TextBufferView.xs
@@ -17,6 +17,11 @@ CODE:
 	textbuffer_view_set_default_indent(view, default_indent, longword_noindent, NULL);
 
 void
+textbuffer_view_set_hidden_level(view, level)
+        Irssi::TextUI::TextBufferView view
+        int level
+
+void
 textbuffer_view_set_scroll(view, scroll)
 	Irssi::TextUI::TextBufferView view
 	int scroll


ailin-nemui:
we need a strategic decision, should we add a lot of setters? I am against it because they don't work like the hash access. and we have Irssi::command. maybe if we would implement hash setters instead, but that needs more code

vague666:
I'm the wrong person to answer that question :)

dequis:
Can't say much here myself, don't really understand the issue. What would this hash access / hash setter stuff look like?

ailin-nemui:
would look like $view->{hidden_level} = MSGLEVEL_...

ailin-nemui:
I will close this for now since I don't like adding more api functions that are better written as Irssi::command. but if someone else disagrees feel free to reopen/merge

vague666:
Time to discuss this too :)
I disagree, the problem is that running the commands with Irssi::command adds complexity to the scripts which can be simplified with api calls
Two examples:
- toggle_hidelevels.pl hooks to 'window hidelevel' command chain so that the script can know when users manually change window hidelevel, since the script toggles hidelevels by calling 'window hidelevel' from the script this has to be taken into account when hooking into the command chain
- away.pl calls 'away' to set/unset away status, it also hooks into the 'away' command chain and in many instances enough care hasn't been taken to take this into account

A way to access the lowlevel setters is the best(easiest) way imo, the hash setters code would be something to look at but XS evades me to a large degree, not enough interest in it to dive in

@irssi/developers

Issue #829, opened 2018-01-23
Add <Shift><Tab> completion


g-elfling:
Add <Shift><Tab> completion because pressing only `<Tab>` to complete the commands is allowed. `<Shift><Tab>` is not allowed, it doesn't do anything. For example in `/q<Tab>` cycles through commands `/query`, `/quit`, `/quote`, and `/Q`. So if one would type `/qu<Tab>` to complete to `/query` and then press `<Shift>` to cycle next to `/quit`. After that pressing `<Shift><Tab>` it would complete back to `/query`.

ailin-nemui:
in fa

PR #830, opened 2018-01-23
Add Shift-Tab completion to gui-readline

g-elfling:
Add <Shift><Tab> completion because pressing only `<Tab>` to complete the commands is allowed. `<Shift><Tab>` is not allowed, it doesn't do anything. For example in `/q<Tab>` cycles through commands `/query`, `/quit`, `/quote`, and `/Q`. So if one would type `/qu<Tab>` to complete to `/query` and then press `<Shift>` to cycle next to `/quit`. After that pressing `<Shift><Tab>` it would complete back to `/query`.

Fixes issue #829

### DIFF
--- a/src/fe-text/gui-readline.c
+++ b/src/fe-text/gui-readline.c
@@ -1279,9 +1279,8 @@ void gui_readline_init(void)
 
         /* line transmitting */
 	key_bind("send_line", "Execute the input line", "return", NULL, (SIGNAL_FUNC) key_send_line);
-	key_bind("word_completion_backward", "", NULL, NULL, (SIGNAL_FUNC) key_word_completion_backward);
-	key_bind("word_completion", "Complete the current word", "tab", NULL, (SIGNAL_FUNC) key_word_completion);
 	key_bind("word_completion_backward", "Completes the previous word", "shifttab", NULL, (SIGNAL_FUNC) key_word_completion_backward);
+	key_bind("word_completion", "Complete the current word", "tab", NULL, (SIGNAL_FUNC) key_word_completion);
 	key_bind("erase_completion", "Remove the completion added by word_completion", "meta-k", NULL, (SIGNAL_FUNC) key_erase_completion);
 	key_bind("check_replaces", "Check word replaces", NULL, NULL, (SIGNAL_FUNC) key_check_replaces);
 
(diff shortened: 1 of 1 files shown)


src/fe-text/gui-readline.c
```diff
@@ -1278,7 +1279,7 @@ void gui_readline_init(void)
 
         /* line transmitting */
 	key_bind("send_line", "Execute the input line", "return", NULL, (SIGNAL_FUNC) key_send_line);
-	key_bind("word_completion_backward", "", NULL, NULL, (SIGNAL_FUNC) key_word_completion_backward);
+	key_bind("word_completion_backward", "Completes the previous word", "shifttab", NULL, (SIGNAL_FUNC) key_word_completion_backward);
```
dequis: This is confusing, it was the first thing I read in this PR and what I interpreted was that, given `a b<Tab>` it completes based on `a` instead of `b`

ailin-nemui:
you can already do this without changing the source code with /bind meta2-Z key stab and /bind stab word_completion_backward
it's a good idea to make it default though

src/fe-text/gui-readline.c
```diff
@@ -1174,6 +1174,7 @@ void gui_readline_init(void)
 	key_bind("key", NULL, "^H", "backspace", (SIGNAL_FUNC) key_combo);
 	key_bind("key", NULL, "^?", "backspace", (SIGNAL_FUNC) key_combo);
 	key_bind("key", NULL, "^I", "tab", (SIGNAL_FUNC) key_combo);
+	key_bind("key", NULL, "meta2-Z", "shifttab", (SIGNAL_FUNC) key_combo);
```
ailin-nemui: rename this to `stab` to stay in line with the existing `chome` and `mup`

src/fe-text/gui-readline.c
ailin-nemui: Hard to come up with a good description.. maybe something like "Choose previous completion suggestion" ?

ailin-nemui:
..Makes me wonder if there is any way to undo a completion completely and revert to before I used Tab

src/fe-text/gui-readline.c
g-elfling: Ok done :P Dunno if "suggestion" word if redundant tho?

src/fe-text/gui-readline.c
g-elfling: ah yeah ofc, done

g-elfling:
Couldn't find that kinda feature, why not make it loool ;) It was weird that this `key_word_completion_backward` thing was built already but was not bound to shift-tab?

ailin-nemui:
lgtm @irssi/developers

PR #833, opened 2018-01-25
we probably should not try to replace replaces

ailin-nemui:
Fixes GL#23


josephbisch:
LGTM, except for possibly the thing about the commit mentioned

ailin-nemui:
@irssi/developers

PR #836, opened 2018-02-02
Fix resizing of windows when used incorrectly

ailin-nemui:


ailin-nemui:
@irssi/developers

dequis:
...going to keep that commit message?

PR #838, opened 2018-02-02
fix UB in file_checksum method

ailin-nemui:
remove UB on high chars

### DIFF
--- a/src/core/settings.c
+++ b/src/core/settings.c
@@ -665,7 +665,7 @@ void sig_term(int n)
    would be nice but would just take more space without much real benefit */
 static unsigned int file_checksum(const char *fname)
 {
-        char buf[512];
+        unsigned char buf[512];
         int f, ret, n;
 	unsigned int checksum = 0;
 
--- a/src/core/settings.c
+++ b/src/core/settings.c
@@ -29,7 +29,6 @@
 #include "settings.h"
 #include "default-config.h"
 
-#include <openssl/sha.h>
 #include <signal.h>
 
 #define SETTINGS_AUTOSAVE_TIMEOUT (1000*60*60) /* 1 hour */
@@ -48,7 +47,8 @@ static int timeout_tag;
 static int config_last_modifycounter;
 static time_t config_last_mtime;
 static long config_last_size;
-static unsigned char config_last_checksum[SHA_DIGEST_LENGTH];
+static guint8 *config_last_checksum;
+static gsize config_last_checksum_length;
 
 static SETTINGS_REC *settings_get(const char *key, SettingType type)
 {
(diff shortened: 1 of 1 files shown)


LemonBoy:
The commit message is atrocious but #836 takes the cake.

dequis:
c0f66c95ff00bdc22155a8ac53d887a9c1bbc90f never forget

josephbisch:
After applying this patch:

```
settings.c:678:25: runtime error: left shift of 138 by 24 places cannot be represented in type 'int'
```

ailin-nemui:
how about something real instead

LemonBoy:
~~I can't really wait to add more of OpenSSL in the mix~~

We can use [glib](https://developer.gnome.org/glib/stable/glib-Data-Checksums.html) instead since we bumped the minimum version to 2.28 some time ago.

dequis:
Yeah I thought the same thing but really I have no opinion on glib checksums vs openssl. The libcrypto parts of openssl are the safest and hashing is a relatively trivial problem.

`g_compute_checksum_for_data` would be shorter but heap allocated so to keep the same allocation style you would end up going through the `g_checksum_new` route which takes the same amount of code. Not that heap allocation hurts.

Also obligatory mention that sha1 has collisions even though the older checksum was worse lol.

ailin-nemui:
@LemonBoy want to rewrite it with glib? I don't care too much which of the libs we use

ailin-nemui:
@irssi/developers

dequis:
Hm, why did you choose not to use `g_compute_checksum_for_data`?

ailin-nemui:
for that we need to read the whole file into memory first

dequis:
Fair

src/core/settings.c
```diff
@@ -661,24 +662,44 @@ void sig_term(int n)
...
 	while ((ret = read(f, buf, sizeof(buf))) > 0) {
-		while (ret-- > 0)
-			checksum += buf[ret] << ((n++ & 3)*8);
+		g_checksum_update(c, buf, ret);
 	}
 	close(f);
-	return checksum;
+
+	checksum = g_new0(guint8, g_checksum_type_get_length(checksum_type));
+	g_checksum_get_digest(c, checksum, &checksum_length);
```
dequis: This crashes with valgrind (and gets good enough garbage in the stack to work decently the rest of the time)

```
==28250== Conditional jump or move depends on uninitialised value(s)
==28250==    at 0x56A2855: g_checksum_get_digest (in /usr/lib/libglib-2.0.so.0.5400.0)
==28250==    by 0x1B1AAE: file_checksum (settings.c:689)
==28250==    by 0x1B1BC1: irssi_config_save_state (settings.c:717)
==28250==    by 0x1B1C57: parse_configfile (settings.c:774)
==28250==    by 0x1B3310: init_configfile (settings.c:794)
==28250==    by 0x1B3310: settings_init (settings.c:901)
==28250==    by 0x19FE44: core_init (core.c:237)
==28250==    by 0x12BE0F: textui_init (irssi.c:147)
==28250==    by 0x12BE0F: main (irssi.c:310)

(irssi:28250): GLib-CRITICAL **: g_checksum_get_digest: assertion '*digest_len >= len' failed
zsh: trace trap (core dumped)  valgrind --log-file=leg ./src/fe-text/irssi
```

checksum_length is inout, so initialize it to `g_checksum_type_get_length(checksum_type)`

ailin-nemui: great

ailin-nemui:
I don't like inout parameters...

PR #843, opened 2018-02-08
restore compat with glib <2.40

ailin-nemui:
reported by @pstray


dequis:
Filed this https://bugzilla.gnome.org/show_bug.cgi?id=793300

PR #844, opened 2018-02-12
dcc get shouldn't fail when file attrs can't be changed

CatboxParadox:
Running irssi from within Termux on an un-rooted NVIDIA Shield TV.

DCC downloads never worked for me, no matter what target directory I'd choose... Would you be willing to accept this PR? I don't think the DCC download should fail just because the desired file attributes can't be set (according to strace fchmod fails with EPERM).

Also extending the hardlink workaround to EACCES, not just EPERM (happens when my target directory is an external drive w/ NTFS).

### DIFF
--- a/src/irc/dcc/dcc-get.c
+++ b/src/irc/dcc/dcc-get.c
@@ -237,8 +237,12 @@ void sig_dccget_connected(GET_DCC_REC *dcc)
 
 		if (temphandle == -1)
 			ret = -1;
-		else
-			ret = fchmod(temphandle, dcc_file_create_mode);
+		else {
+			if (fchmod(temphandle, dcc_file_create_mode) != 0)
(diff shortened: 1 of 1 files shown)


src/irc/dcc/dcc-get.c
```diff
@@ -247,7 +250,7 @@ void sig_dccget_connected(GET_DCC_REC *dcc)
 
 			if (ret == -1 &&
 			    /* Linux */
-			    (errno == EPERM ||
+			    (errno == EPERM || errno == EACCES ||
```
dequis: This probably belongs below `/* FUSE */` if it happens with ntfs-3g

ailin-nemui:
maybe it would be good to print a warning if chmod fails?

ailin-nemui:
looks ok to me..

PR #852, opened 2018-02-16
Document second parameter (seek position) of /cat command

dequis:


ailin-nemui:
I wonder if this should really be documented xD

dequis:
Hm. I didn't stop to think if I should.

ailin-nemui:
it's certainly an odd secret flag, but otherwise lgtm, feel free to merge

PR #857, opened 2018-03-08
Compare channels/networks fields case-insensitively

### LINKED ISSUE #856: /channel remove and case mismatch in buggy config
odd behaviour with buggy config:
```
chatnets = {
  Freenode = {
    type = "IRC";
  };
};

channels = (
  { name = "#python"; chatnet = "freenode"; autojoin = "yes"; },
);
```

```
/channel remove #python Freenode
/channel list
```
shows the channel as removed, but /save doesn't change it (since the network case is wrong)

reported by bifidotftw

LemonBoy:
Fixes #856

### DIFF
--- a/src/core/servers-setup.c
+++ b/src/core/servers-setup.c
@@ -478,8 +478,8 @@ static int compare_server_setup (CONFIG_NODE *node, SERVER_SETUP_REC *server)
 		return 0;
 	}
 
-	if (g_strcmp0(address, server->address) != 0 ||
-	    g_strcmp0(chatnet, server->chatnet) != 0 ||
+	if (g_ascii_strcasecmp(address, server->address) != 0 ||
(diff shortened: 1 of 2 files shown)


src/core/servers-setup.c
```diff
@@ -474,10 +474,15 @@ static int compare_server_setup (CONFIG_NODE *node, SERVER_SETUP_REC *server)
 	chatnet = config_node_get_str(node, "chatnet", NULL);
 	port = config_node_get_int(node, "port", 0);
 
+	if (address == NULL || chatnet == NULL) {
+		return 0;
+	}
+
 	if (g_strcmp0(address, server->address) != 0 ||
 	    g_strcmp0(chatnet, server->chatnet) != 0 ||
```
dequis: Did you forget these, or...?

LemonBoy: That's what you get when you manually select the hunks you need...

PR #862, opened 2018-03-17
fix a crash when trying to append to a NULL line

ailin-nemui:
reported by @vague666

### DIFF
--- a/src/fe-text/gui-printtext.c
+++ b/src/fe-text/gui-printtext.c
@@ -120,11 +120,18 @@ void gui_printtext_internal(int xpos, int ypos, const char *str)
 	next_xpos = next_ypos = -1;
 }
 
+static void view_add_eol(TEXT_BUFFER_VIEW_REC *view, LINE_REC **line);
+
 void gui_printtext_after_time(TEXT_DEST_REC *dest, LINE_REC *prev, const char *str, time_t time)
 {
 	GUI_WINDOW_REC *gui;
 
 	gui = WINDOW_GUI(dest->window);
+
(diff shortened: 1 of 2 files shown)


ailin-nemui:
@irssi/developers 

obscure textbuffer stuff only triggered from scripts

dequis:
What's the reproducer?

ailin-nemui:
```perl
$window->gui_printtext_after(undef, MSGLEVEL_CLIENTNOTICE, "xxx")
```
seen in similar fashion in `buf.pl`

vague666:
The fix helped but I am still having problems with this code:
http://vague.se/pastes/buf.pl
If I switch the print to gui_printtext_after I see the screen update with a bunch of text and then it's cleared, with print the text stays visible on screen, I'd like to know if any of you can reproduce the behavior

ailin-nemui:
did you try scolling up?

vague666:
Can't remember, I'll try it when I get home

vague666:
The view is actually scrolled up or whatcha-ma-call it
How the screen is after /upgrade:
https://vague.se/pastes/2018-03-21_19:53:03_1920x1080_scrot.png
after pgup:
https://vague.se/pastes/2018-03-21_19:53:41_1920x1080_scrot.png

vague666:
Not sure if that behavior is the desired one, I would expect =Upgrade= to be visible and not scrolled up

vague666:
Maybe I don't understand how gui_printtext_after works? If I think a bit about it, it looks like the text is printed from bottom up, is there a function that prints top to bottom?

PR #871, opened 2018-04-08
Make config_write more atomic to prevent truncation when out of space

### LINKED ISSUE #817: Disk full / write error on exit destroys config file? :-\
I encountered a disk full error in irssi, and shut down the machine to expand the VM's disk. When the machine came back and I started irssi, I discovered that the config file had been truncated to 0 bytes. I assume irssi tried to write it on exit when the machine shut down, but failed in a nasty way.

dequis:
This is modeled after glib's g_file_set_contents. It doesn't use that
function directly because the writing is done with GIOChannel
streaming-like writes and g_file_set_contents expects the whole thing to
be in-memory.

Main differences with g_file_set_contents:

- complete lack of win32 special casing (cygwin/WSL should work though)
- no fallocate() (linux only, but we don't know the size upfront, anyway)
- always calls fsync (glib skips it on btrfs or when not overwriting)

Other than that, it's the same old mkstemp + fsync + rename.

-------

Fixes #817

In https://github.com/irssi/irssi/issues/817#issuecomment-358341073 I mentioned I wasn't sure if i should keep copying g_file_set_contents(). Few months later, no one is going to refactor config writing to be in-memory, so finishing that WIP code is obviously better than nothing.

The WIP code used g_rename, g_unlink, but those require including gstdio.h and their implementation is all win32 portability code so mehhhh.

### DIFF
--- a/src/lib-config/write.c
+++ b/src/lib-config/write.c
@@ -301,30 +301,67 @@ int config_write(CONFIG_REC *rec, const char *fname, int create_mode)
 {
 	int ret;
 	int fd;
+	int save_errno;
+	char *tmp_name;
+	const char *dest_name;
 
 	g_return_val_if_fail(rec != NULL, -1);
         g_return_val_if_fail(fname != NULL || rec->fname != NULL, -1);
         g_return_val_if_fail(create_mode != -1 || rec->create_mode != -1, -1);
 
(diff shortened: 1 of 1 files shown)


ailin-nemui:
@irssi/developers

PR #873, opened 2018-04-16
record line info on empty lines

ailin-nemui:
Fix for GL#31


ailin-nemui:
we just need to check if this could have any odd unforeseen effects. I can't find them but maybe someone else has an idea where to look?

ailin-nemui:
@irssi/developers

ailin-nemui:
had to be reverted

PR #880, opened 2018-05-13
fix sending of server passwords starting with ":" or containing spaces

### LINKED ISSUE #362: Server passwords doesn't start with colon with spaced password
If the server password to be sent contains spaces, the argument has to have a colon prefixed to it so that the whole password is recognized as a single argument by the server, otherwise a password containing a space should be interpreted as having multiple arguments, which is nonsensical since the `PASS` command only takes one. irssi currently just sends `PASS <password>` instead.

jkhsjdhjs:
This PR fixes sending of server passwords according to RFC.
Fixes #362 

If the server passwords contains at least one space a colon will be added in front of it when logging in.
Same applies if the password begins with a colon.


ailin-nemui:
thanks for the fix, we've decided to ignore this for the time being

PR #883, opened 2018-06-05
Ignore hidden

vague666:
This appears to work but I'm a bit unsure about the bitwise operations in ignore_match_level


vague666:
I removed the three extra commits that somehow got included so we can open this up again :)

dequis:
Open a new one, the reopen button is disabled for this one ("The ignore_hidden branch was force-pushed or recreated.")

PR #884, opened 2018-06-06
Added HIDDEN level to ignores

vague666:


vague666:
The testing I did worked well but I might've missed something

src/core/ignore.h
```diff
@@ -29,6 +29,7 @@ int ignore_check(SERVER_REC *server, const char *nick, const char *host,
 enum {
 	IGNORE_FIND_PATTERN = 0x01, /* Match the pattern */
 	IGNORE_FIND_NOACT   = 0x02, /* Exclude the targets with NOACT level */
+	IGNORE_FIND_HIDDEN  = 0x03, /* Exclude the targets with HIDDEN level */
```
ailin-nemui: this should be 0x04 (bits go like 1 2 4 8 0x10 0x20 0x40...)

src/fe-common/core/fe-ignore.c
```diff
@@ -157,7 +157,8 @@ static void cmd_ignore(const char *data)
 		g_strsplit(chanarg, ",", -1);
 
 	rec = ignore_find_full(servertag, mask, patternarg, channels,
-			  IGNORE_FIND_PATTERN | ((level & MSGLEVEL_NO_ACT) ? IGNORE_FIND_NOACT : 0));
+			  IGNORE_FIND_PATTERN | ((level & MSGLEVEL_NO_ACT) ? IGNORE_FIND_NOACT : 0) |
```
ailin-nemui: this is getting too ugly, we should add a temp variable flags= IGNORE_FIND_PATTERN and ...if (level &...) flags |= ...;

vague666:
I would prefer making the #define ignore_match_level into a function too, but it didn't become too ugly when I added hidden to it

ailin-nemui:
I think what would help there is a `#define FLAG_MSGLEVELS ( MSGLEVEL_NO_ACT  | MSGLEVEL_HIDDEN )`

ailin-nemui:
@irssi/developers

dequis:
What does this do?

ailin-nemui:
I think it allows you to /ignore ..... hidden and then it will hide the messages instead of ignoring them completely

vague666:
Indeed

PR #886, opened 2018-06-14
remove = from format because people get confused by it

vague666:


ailin-nemui:
I like this but to be fair it does make the display a bit "odd" when filtering shorter settings names. however there aren't any alternatives I can think of

ailin-nemui:
@irssi/developers

PR #895, opened 2018-07-18
Keep errstr set to NULL if errno is not set

janikrabe:
Don't use errno if it is not set and show the default error message
instead. This prevents messages like "SSL handshake failed: Success"
from being shown.


janikrabe:
@dequis thanks. This happened after receiving an empty handshake from a misconfigured server. You can reproduce it by running `nc -lp 6697`, using `/connect -ssl localhost` and then killing the netcat process.

dequis:
Oh easier than I thought, so yeah this works. Thanks!

dequis:
@irssi/developers

PR #896, opened 2018-07-24
Don't ignore SIGWINCH when window size is unchanged

phillid:
Currently, the SIGWINCH handler assumes that the entire window doesn't
need to be redrawn. This assumption fails when session detachment tools
like abduco rely on SIGWINCH to ask a curses program to redraw itself
after a session is resumed.

~~This patch makes SIGWINCH trigger a full redraw instead of just a dirty
mark being applied.~~

The updated patch makes irssi stop ignoring a SIGWINCH if the reported terminal size hasn't changed.

### DIFF
--- a/src/fe-text/term.c
+++ b/src/fe-text/term.c
@@ -81,11 +81,9 @@ void term_resize_dirty(void)
 	if (!term_get_size(&width, &height))
 		width = height = -1;
 
-	if (height != term_height || width != term_width) {
-		term_resize(width, height);
-		mainwindows_resize(term_width, term_height);
-		term_resize_final(width, height);
-	}
+	term_resize(width, height);
+	mainwindows_resize(term_width, term_height);
+	term_resize_final(width, height);
 }
 
 #ifdef SIGWINCH


ailin-nemui:
you can run `dtach -r ctrl_l` or manually press Ctrl+L if you need this

phillid:
@ailin-nemui that is a workaround indeed. But surely turning SIGWINCH into a nop is expecting breakage to occur? People don't send SIGWINCH for no reason; they send it when a redraw is expected.

ailin-nemui:
no, regular terminals send winch to tell irssi that their WINdowsize CHanged and the dirty flag handles that case in an optimised manner (especially since a full redraw is not necessary in many cases)

phillid:
That leaves you wide open and exposed for when a full redraw is necessary. How many CPU ticks does this flawed optimisation save?

ailin-nemui:
a full redraw is never necessary unless you use abduco :)

ailin-nemui:
most importantly it makes the screen adapt to new window size much faster over slow network links

phillid:
That's true. More generally, anything that uses a nice simplistic approach like abduco where the terminal state is not kept between detachment and reattachment. I would have thought that this is a use case that irssi is prepared to support, considering the triviality of it. I am aware that slow network will exacerbate any increased redraw times. I notice that other IRC clients redraw on SIGWINCH without "optimisation".

phillid:
Hmm, looks like Travis itself broke rather than my code? I wonder if that can be re-run.

dequis:
Re-ran that job.

This version of the patch looks much more reasonable than the original one

ailin-nemui:
TODO: resize terminal interactively, check how many times winch is generated and what the size is at the point of TIOCGWINSZ. do this for at least gnome-terminal and original xterm

PR #905, opened 2018-08-10
Add tests for topic events

### LINKED ISSUE #903: "no topic set" regression
Received this on staff@irssi.org, haven't looked into it:

> Hello.
> 
> Upon joining the channel, it says:
> 
> 11:16:25 -!- Topic for ##austrian-economics: Welcome to the unofficial chat of the Ludwig von Mises Institute! Be warned: no censorship applies to this channel, and off-topic conversations may be present.
> 11:16:25 -!- Topic set by zolk3ri [~zolk3ri@gateway/tor-sasl/zolk3ri] [Thu Aug  9 11:01:33 2018]
> 
> but when I type /topic, it shows:
> 
> 11:55:08 -!- No topic set for ##austrian-economics
> 
> And it also doesn't show up on the top bar.
> 
> I am using Irssi v1.2-g0c94d44f.
> 
> Kind Regards,
> Zoltán Kéri

Related #897 #888

horgh:
As we called the same function for the topic info numeric, we ended up
losing the topic as that numeric does not include it. We now call a
different function to handle this case more carefully. Fixes #903.

Note if we want to be able to show who & unset a topic and when as we discussed on #903, this won't work.

Some manual tests:

- Test 1
  - Join channel with no topic. Look good.
  - Other user sets topic. Looks good.
  - Other user unsets topic. Looks good.

- Test 2
  - Join channel with no topic.
  - I set topic. Looks good.
  - I unset topic. Looks good.

- Test 3
  - Join channel with topic. Looks good.
  - Other user sets topic. Looks good.
  - Other user unsets topic. Looks good.

- Test 4
  - Join channel with topic. Looks good.
  - I set topic. Looks good.
  - I unset topic. Looks good.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -51,8 +51,23 @@ src/perl/*/Makefile.old
 src/fe-fuzz/crash-*
 src/fe-fuzz/oom-*
 
-tests/irc/core/test-irc
-tests/irc/core/test-irc.trs
+/core
+/tests/fe-common/core/test-formats
+/tests/fe-common/core/test-formats.log
+/tests/fe-common/core/test-formats.trs
+/tests/fe-common/core/test-suite.log
+/tests/irc/core/core
+/tests/irc/core/test-channel-events
+/tests/irc/core/test-channel-events.log
+/tests/irc/core/test-channel-events.trs
+/tests/irc/core/test-irc
+/tests/irc/core/test-irc.log
+/tests/irc/core/test-irc.trs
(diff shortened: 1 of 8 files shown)


horgh:
Yeah, that's way better! Thanks!

I also pulled back in the original PR changes as they're needed for this. Not sure if we want to rework the approach more, but anyway.

ailin-nemui:
@irssi/developers

PR #909, opened 2018-08-12
Fix `/save` replacing symlinks with regular files

### LINKED ISSUE #906: /save replaces symlinks with regular files
The `/save` command behaves differently on master (I tested 85df01d1338e759b7cac18414ff98a7dce98e2c4) than in 1.1.1: `/save` now replaces symbolic links in `~/.irssi/` with regular files. The new behavior is annoying for people that keep their dotfiles in a VCS repository and mostly use symlinks in `$HOME`.

8deb6182c692904733544568479035b31fe1007e may be the commit that changed the behavior.

meribold:
A side-effect of 8deb618 is that `/save` may replace configuration files that are symlinks with regular files.  Fix this by resolving all symlinks before renaming the temporary file.

This fixes #906.

### DIFF
--- a/src/lib-config/write.c
+++ b/src/lib-config/write.c
@@ -304,6 +304,7 @@ int config_write(CONFIG_REC *rec, const char *fname, int create_mode)
 	int save_errno;
 	char *tmp_name;
 	const char *dest_name;
+	char *real_dest = NULL;
 
 	g_return_val_if_fail(rec != NULL, -1);
         g_return_val_if_fail(fname != NULL || rec->fname != NULL, -1);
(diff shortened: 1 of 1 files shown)


ailin-nemui:
rename doesn't work if symlink point to another FS, I didn't check your fix yet but did you consider this?

meribold:
I didn't. I can change the code to base the path of the temporary file on the *expanded* path of the configuration file. I assume that would take care of this issue.

meribold:
I added another commit that should prevent renames across file systems.

ailin-nemui:
@dequis

ailin-nemui:
@irssi/developers

dequis:
Turns out this breaks when the file didn't already exist

```
Error: Couldn't save configuration file: No such file or directory
18:50 -!- Irssi: Error saving theme to /tmp/foobar/default.theme: No such file or directory
```

PR #897, opened 2018-07-25
Flag topic as unset if it is zero length

horgh:
This is to fix #888.

PR #911, opened 2018-08-13
Revert "Flag topic as unset if it is zero length"

ailin-nemui:
Reverts irssi/irssi#897

Hotfix for #903


ailin-nemui:
I will revert this quickly and then we will fix this again & more better

PR #913, opened 2018-08-15
Add server-fuzz to fe-fuzz

josephbisch:
Based on test-796.c by @horgh in the tests directory.

Will make corresponding oss-fuzz PR.

### DIFF
--- a/src/fe-fuzz/Makefile.am
+++ b/src/fe-fuzz/Makefile.am
@@ -1,6 +1,6 @@
 SUBDIRS = irc fe-common
 
-bin_PROGRAMS = irssi-fuzz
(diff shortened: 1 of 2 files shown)


Dor1s:
@ailin-nemui would you be interested in landing https://github.com/irssi/irssi/pull/921 and then enabling continuous fuzzing on OSS-Fuzz?

ailin-nemui:
@Dor1s thanks for your interest, unfortunately I have no experience with the great oss fuzz project and our best man here has gone back to the real life. #921 is probably not the correct thing to do, we would have to find out what's the real iss causing the log messages

Issue #907, opened 2018-08-11
HIDDEN level related crash


dequis:
Reported by yitz on irc (2018-08-11 03:48 UTC), happened twice, has coredumps and sent them to me.

Irssi 1.1.1-2 on arch linux, no debug symbols.

Stack traces posted: http://ix.io/1jRw http://ix.io/1jRx http://ix.io/1jRy (last one is the same but with gdb)

The shorter one:

```
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x0000555dfa3cb297 in ?? ()
[Current thread is 1 (Thread 0x7ff2de438740 (LWP 13322))]
(gdb) bt
#0  0x0000555dfa3cb297 in ?? ()
#1  0x0000555dfa3cb6c5 in textbuffer_view_scroll ()
#2  0x0000555dfa3bc74a in gui_window_scroll ()
#3  0x0000555dfa43d9e2 in ?? ()
#4  0x0000555dfa43df97 in signal_emit ()
#5  0x0000555dfa3f3d19 in ?? ()
#6  0x0000555dfa3f5832 in key_pressed ()
#7  0x0000555dfa3b9269 in ?? ()
#8  0x0000555dfa43d9e2 in ?? ()
#9  0x0000555dfa43df97 in signal_emit ()
#10 0x0000555dfa3ba7da in ?? ()
#11 0x00007ff2df3d9cb3 in ?? () from /usr/lib/libglib-2.0.so.0
#12 0x00007ff2df3d91d6 in g_main_context_dispatch () from /usr/lib/libglib-2.0.so.0
#13 0x00007ff2df3d95b1 in ?? () from /usr/lib/libglib-2.0.so.0
#14 0x00007ff2df3d963e in g_main_context_iteration () from /usr/lib/libglib-2.0.so.0
#15 0x0000555dfa3b5c42 in main ()
(gdb)
```

Started happening after switching to the libpurple variant of bitlbee, and three days after upgrading from 1.1.1-1 to 1.1.1-2, which is a perl 5.28 rebuild with no other changes.

```
00:49 <+yitz> Started playing with libpurple yesterday. Thu 2018-08-09 15:34:29 PDT  &  Fri 2018-08-10 20:41:51 PDT
00:50 <+yitz> So ... as of so far, I got a coredump every day I've been using it
```

BItlbee itself appears to not have crashed.

```
[...]

PR #914, opened 2018-08-20
properly record line info on empty lines

ailin-nemui:
another attempt at fixing #873 GL#31 #907 

it's basically the same as before with a special case for trackbar.


ailin-nemui:
anything which prints an empty string was previously ignored. now, it starts a new line. this might cause issues -- already identified the trackbar one last time, but it's unknown if there are other special cases which need consideration

dequis:
>it's unknown if there are other special cases which need consideration

This makes the `g_slice_new0` solution look more attractive tbh

ailin-nemui:
but that doesn't record the proper time and level. we can do both, this one to fix a missing thing and new0 to avoid uninitialised memory

ailin-nemui:
@irssi/developers

PR #919, opened 2018-08-23
Use-after-frees

ailin-nemui:
coypu AT sdf DOT org wrote:

> Hi folks!
> 
> I tried clang-static-analyzer on irssi 1.1.1, it seems like it
> finds some things. Here's a diff, but there might be more that you would
> want to check, or choose to work differently.
> 
> (in special-vars.c, ret is commands->data sometime)
> 
> I hope it's not too much trouble if reported as a confidential bug.
> Thanks.

### DIFF
--- a/src/core/modules.c
+++ b/src/core/modules.c
@@ -289,7 +289,7 @@ void modules_deinit(void)
 
 	while (list != NULL) {
 		module_uniq_destroy(list->data);
-		gconstpointer tmp = list->data;
+		void *tmp = list->data;
 		list = g_slist_remove(list, list->data);
 		g_free(tmp);
 	}
--- a/src/core/modules.c
+++ b/src/core/modules.c
@@ -289,8 +289,9 @@ void modules_deinit(void)
 
 	while (list != NULL) {
 		module_uniq_destroy(list->data);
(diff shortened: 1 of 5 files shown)


src/core/modules.c
```diff
@@ -289,8 +289,9 @@ void modules_deinit(void)
 
 	while (list != NULL) {
 		module_uniq_destroy(list->data);
-		g_free(list->data);
+		gconstpointer tmp = list->data;
 		list = g_slist_remove(list, list->data);
+		g_free(tmp);
```
LemonBoy: `g_free` doesn't like the `gconstpointer`

ailin-nemui: it's a bug in the C language

LemonBoy: Not the only one and not the worst one but still :)

dequis:
Turns out the dcc part of the patch changed a use after free from "referencing the address of an object which just got deallocated without touching its contents" to "reading an address out of an object that just got deallocated" making it crash on exist. Cool. That's fixed now.

ailin-nemui:
cool

ailin-nemui:
@irssi/developers

PR #931, opened 2018-09-04
Disconnect SASL properly in case the sasl module got unloaded from server

### LINKED ISSUE #629: sasl_disconnect_on_failure not effective when server ignores CAP
cf #618

ailin-nemui:
stops from getting on the network when sasl is unavailable

fixes #629

### DIFF
--- a/src/fe-common/irc/fe-sasl.c
+++ b/src/fe-common/irc/fe-sasl.c
@@ -40,36 +40,14 @@ static void sig_sasl_failure(IRC_SERVER_REC *server, const char *reason)
 	printformat(server, NULL, MSGLEVEL_CRAP, IRCTXT_SASL_ERROR, reason);
 }
 
-static void sig_cap_end(IRC_SERVER_REC *server)
-{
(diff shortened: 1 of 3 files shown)


ailin-nemui:
by the way the disconnect code was in the wrong place (frontend instead of core)

LemonBoy:
So this server pulls the `sasl` cap during the negotiation? What a wild world we live in.
The patch is fine, just a couple of nits:

- `g_hash_table_lookup_extended` can be replaced with `g_hash_table_lookup` since we only care about the key being there
- `cap_supported` cannot be null if we get there

ailin-nemui:
thanks! we can also get there when the server didn't have any CAP

ailin-nemui:
also it seems that we store value NULL so that would create a problem with the non-extended hash table lookup

LemonBoy:
> we can also get there when the server didn't have any CAP

I forgot about ill-behaving servers :)

> also it seems that we store value NULL so that would create a problem with the non-extended hash table lookup

Yup, I also forgot about the missing values!

PR #933, opened 2018-09-08
Hide all levels in window hiddenlevel from act bar

dwfreed:
Also don't always hide HIDDEN level from activity bar.


dwfreed:
Suggested way to fix this? gui-windows.h is in fe-text, but this code is in fe-common; arguably window activity is an fe-text thing, though

ailin-nemui:
could  you instead /ignore * joins parts quits hidden no_act? In fact I'm not sure anymore if we should even have hidden added to the no_act levels at all

dwfreed:
Having to maintain both an ignore and a hiddenlevel sucks, especially when you want to remove something from the hiddenlevel for a while

PR #938, opened 2018-09-17
Don't trigger activity on hidden lines by default

ailin-nemui:
attempt to fix #933

### DIFF
--- a/docs/signals.txt
+++ b/docs/signals.txt
@@ -286,6 +286,7 @@ themes.c:
 
 window-activity.c:
  "window hilight", WINDOW_REC
+ "window hilight check", TEXT_DEST_REC, char *msg, int *data_level, int *should_ignore
  "window dehilight", WINDOW_REC
  "window activity", WINDOW_REC, int old_level
  "window item hilight", WI_ITEM_REC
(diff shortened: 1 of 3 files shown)


dwfreed:
activity_hide_level shouldn't have to include HIDDEN, because that ends up being counterintuitive.  It would suggest that whenever a message has HIDDEN level, it would not trigger activity, rather than being based on the window's hidelevel setting.  The hidelevel activity hiding should always happen, and if a user wants to put HIDDEN in their activity_hide_level, then that will always not trigger activity if the message has HIDDEN level, even if window hidelevel would not hide it (eg because it's set to none).

ailin-nemui:
then we don't have a way to trigger activity for hidden things, which is bad imo. for example you may want to hide the joins by default but still see in your Act bar that someone joined

dwfreed:
That's also counterintuitive; people aren't going to want to see activity for a channel and not see anything change in the channel as a result (because of their window hidelevel).  If they cared about the activity, they'd generally care to see it too, so they'd remove that level from their window hidelevel.

ailin-nemui:
@irssi/developers

PR #941, opened 2018-09-17
Add window_default_hidelevel setting

dwfreed:
Allows specifying the hidelevel that new windows get by default.


ailin-nemui:
@irssi/developers

PR #945, opened 2018-09-23
makes '/window item goto' search on all servers

### LINKED ISSUE #904: /window item goto only searches on active server
When there are channels from different servers in the current window `/window item goto #channel` only works for channels on the active network. The reason is that [window-commands.c#L543](https://github.com/irssi/irssi/blob/master/src/fe-common/core/window-commands.c#L543) looks for items on the active server only.

It might help to change that line so server says NULL instead. It might need to handle when items with the same name from several networks, or enable users to define the network at the end or with -network like for /msg

bcattaneo:
Fixes #904 partially, still need to do something about this:
> It needs to handle when items with the same name from several networks, or should enable users to define the network at the end or with -network like for /msg

### DIFF
--- a/src/fe-common/core/window-commands.c
+++ b/src/fe-common/core/window-commands.c
@@ -541,6 +541,8 @@ static void cmd_window_item_goto(const char *data, SERVER_REC *server)
 		item = tmp == NULL ? NULL : tmp->data;
 	} else {
 		item = window_item_find_window(active_win, server, target);
+		if (item == NULL)
+			item = window_item_find_window(active_win, NULL, target);
 	}
 
         if (item != NULL)


ailin-nemui:
my only concern is that it's more difficult to implement the proper behaviour, which should  be /win item goto #chan , /win item goto #chan , /win item goto #chan , ... should probably cycle between all items . but I guess we could also use this patch as it's an improvement over the current situation. @irssi/developers  ?

ailin-nemui:
I would prefer the following solution: starting with the currently active item, search for the ->next item which matches your goto channel. this may require a new function in window-items.c. If you want you can try to implement that

bcattaneo:
@ailin-nemui sounds good! let me work on it

ailin-nemui:
Closing for inactivity

PR #948, opened 2018-09-28
invalidate startline and bottom_startline when hidden

ailin-nemui:


ailin-nemui:
@irssi/developers

ailin-nemui:
may be worth to ponder if there is something I overlooked

PR #951, opened 2018-09-29
fe-text: clear hidelevel in layout if default

dwfreed:
This is required, otherwise setting the hidelevel to the default will
not be written in the layout and so won't persist.


ailin-nemui:
@irssi/developers

PR #953, opened 2018-09-30
Use ##__VA_ARGS__ extension to allow for argument-less ...

dwfreed:
This also makes the macros work when no arguments are given.  This is technically a GCC extension, but both GCC and clang support it in standards-compliant mode without complaint.


ailin-nemui:
@irssi/developers

PR #954, opened 2018-09-30
 Add macros to check for correct signal_emit params

dwfreed:
This allows the compiler to throw an error if the params argument doesn't match the number of params provided after.  Also fixes the issues the macros detected.

### DIFF
--- a/src/core/commands.c
+++ b/src/core/commands.c
@@ -305,7 +305,7 @@ void command_runsub(const char *cmd, const char *data,
 
 	if (*data == '\0') {
                 /* no subcommand given - list the subcommands */
-		signal_emit("list subcommands", 2, cmd);
+		signal_emit("list subcommands", 1, cmd);
 		return;
 	}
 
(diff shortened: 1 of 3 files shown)


ailin-nemui:
will this cause any issues on solaris builds?

ahf:
Neat.

dequis:
Hm, turning a public symbol into a macro is at least ABI breaking, not sure if it counts as API breaking too.

dwfreed:
It's only API breaking if you're dumb enough to try to use the symbol in any other way besides calling it.  But yes, it is ABI breaking.  About to go to sleep, will bump the ABI number later (or somebody else can push that commit to my branch)

ailin-nemui:
we might be able to solve this better by using a signal registry and typed emissions

dwfreed:
The current type system we have, for things like IS_SERVER, is a not-insignificant chunk of CPU time in my playback tests, which is why I've talked about getting rid of it on IRC.  Also, that is a runtime check, whereas this is a compile-time check.

PR #955, opened 2018-10-01
fix and document window width on screen width dependency

ailin-nemui:
previously, the rsplits would be created in an incorrect way and the text would spill over for widths < 20

### DIFF
--- a/src/fe-text/mainwindows.h
+++ b/src/fe-text/mainwindows.h
@@ -5,7 +5,7 @@
 #include "term.h"
 
 #define WINDOW_MIN_SIZE 2
-#define NEW_WINDOW_WIDTH 10
+#define NEW_WINDOW_WIDTH 20 /* must be >= MIN_SCREEN_WIDTH defined in term.c */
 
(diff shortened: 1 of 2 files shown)


ailin-nemui:
@irssi/developers

src/fe-text/mainwindows.h
```diff
@@ -5,7 +5,7 @@
 #include "term.h"
 
 #define WINDOW_MIN_SIZE 2
-#define NEW_WINDOW_WIDTH 10
+#define NEW_WINDOW_WIDTH 20 /* must be >= MIN_SCREEN_WIDTH defined in term.c */
```
LemonBoy: What about using `G_STATIC_ASSERT` [1] to enforce this invariant?

[1] https://developer.gnome.org/glib/stable/glib-Miscellaneous-Macros.html#G-STATIC-ASSERT:CAPS

ailin-nemui: nice idea but won't work unless we make MIN_SCREEN_WIDTH public

PR #956, opened 2018-10-02
Remove cubes from irssi.conf

ailin-nemui:
This is long and ugly and better done externally (script)


ailin-nemui:
@irssi/developers

PR #957, opened 2018-10-04
Improve rawlog performance

ailin-nemui:


dequis:
Fancy! When does this bottleneck / how to test perf difference?

ailin-nemui:
@dwfreed complained about this when replaying his enormous znc with largish rawlog_lines

src/core/rawlog.c
```diff
@@ -61,12 +62,8 @@ void rawlog_destroy(RAWLOG_REC *rawlog)
 /* NOTE! str must be dynamically allocated and must not be freed after! */
 static void rawlog_add(RAWLOG_REC *rawlog, char *str)
 {
-	if (rawlog->nlines < rawlog_lines || rawlog_lines <= 2)
-		rawlog->nlines++;
-	else {
-		void *tmp = rawlog->lines->data;
-		rawlog->lines = g_slist_remove(rawlog->lines,
-					       rawlog->lines->data);
+	if (rawlog->lines->length >= rawlog_lines && rawlog_lines > 0) {
```
dwfreed: Maybe make this a while?  So rawlog_lines can be reduced without needing to reconnect to have it apply to existing rawlogs.

dwfreed:
Other than that and the backwards use of the queue (which, since it's a doubly-linked list, doesn't affect performance, just prevents the use of g_queue_foreach giving the right order of lines), looks good to me

ailin-nemui:
while loop jay nay? @dequis

ailin-nemui:
@dwfreed do you want to push a commit to implement your suggestions?

vague666:
Just running the patch seems to work ok and rawlog is printed out in the correct order :)
I don't have znc to test playback performance though

ailin-nemui:
thanks, vague

PR #959, opened 2018-10-05
Route notices intended for channels into the channel

ailin-nemui:
This works for

    -ChanServ(ChanServ@services.)- [#irssi] Welcome to #irssi.

or

    -bI(~xxx@unaffiliated/account)- [#irssi] your message is restrained

etc.

### DIFF
--- a/src/fe-common/irc/fe-irc-messages.c
+++ b/src/fe-common/irc/fe-irc-messages.c
@@ -201,12 +201,15 @@ static void sig_message_irc_action(IRC_SERVER_REC *server, const char *msg,
 
 static char *notice_channel_context(SERVER_REC *server, const char *msg)
 {
+	if (!settings_get_bool("notice_channel_context"))
+		return NULL;
+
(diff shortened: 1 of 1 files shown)


ailin-nemui:
good idea

dwfreed:
1. In my case, I have a permanent query window with ChanServ (and other services), intentionally so that all ChanServ messages go there; this bypasses that
2. As an oper, I don't want random users to be able to flood the channel window of their choice just by forming their message in a special way

dequis:
+1 for the setting, I thought about it but failed to come up with use cases where it's undesirable.

Regarding the latter, that's probably worth thinking about a bit more, maybe check what restrictions are set by other clients that implement this behavior. Channel membership of the source nick?

ailin-nemui:
we talked about it on IRC but ChanServ, a prime offender, is often not on channel

ailin-nemui:
@irssi/developers

PR #971, opened 2018-11-09
fix paste_join_multiline

### LINKED ISSUE #970: paste_join_multiline not working?

ailin-nemui:
fixes #970

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -732,6 +732,7 @@ scripts/examples/Makefile
 tests/Makefile
 tests/fe-common/Makefile
 tests/fe-common/core/Makefile
+tests/fe-text/Makefile
 tests/irc/Makefile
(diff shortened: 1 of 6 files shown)


vague666:
Seems to work better but I'm not sure I agree with the assumption that only indented lines should be squached to the earlier unindented line
I'd prefer having all \n replaced with , regardless of indentation(which can/should be removed)
Maybe make the separation char a setting, paste_replace_newline_with

vague666:
Maybe make a lone \n on a line the squashing separator, ie
line1
line2
line3

line4
line5

turns into
line1, line2, line3
line4, line5

ailin-nemui:
that could be discussed in a feature request, this is to restore the code to working conditions

PR #973, opened 2018-11-12
expose wcwidth and related to perl

ailin-nemui:
easier for scripts to match Irssi's width calculations


ailin-nemui:
@irssi/developers

PR #975, opened 2018-11-19
Fix irssi build on IBM i and AIX

NattyNarwhal:


ailin-nemui:
this is highly asymmetric. for the time being maybe you can get by with -Dcap_init=irc_cap_init  in your cflags? You're probably the single only person who cares about and has access to AIX.

cap_ may be overloaded but we should change all API functions and will have to bump the abi constant

NattyNarwhal:
Setting a preprocessor flag would also refine the name for the system definition in headers too.

The change to the automake flags in configure.ac don't affect `make dist` release tarballs. It does squelch some warnings in autogen though, because of [this](https://stackoverflow.com/questions/21609580/autotools-build-fails-due-to-subdir-objects-option-in-am-init-automake). It seems this could have deeper implications though, so I can revert it.

ailin-nemui:
consistently rename all public cap functions, increase ABI number in common.h and we can ship this

NattyNarwhal:
Changes made. I also reverted my changes to the configure script because an RPM build environment would set the magic variables for 64-bit mode automatically, and a user can do it themse4lves if building outside of RPM anyways. The subdir-objects change has been reverted too, no changes to function. Based on what upstream says, you'll probably want to toggle that bit, but it's not my decision to make.

ailin-nemui:
@irssi/developers

src/irc/core/irc-cap.h
```diff
@@ -1,9 +1,10 @@
 #ifndef __IRC_CAP_H
 #define __IRC_CAP_H
 
-void cap_init(void);
-void cap_deinit(void);
-int cap_toggle (IRC_SERVER_REC *server, char *cap, int enable);
-void cap_finish_negotiation (IRC_SERVER_REC *server);
+/* this is prefixed as to not conflict with an AIX/i function in their stdlib */
```
dequis: I think we can drop this comment

PR #978, opened 2018-11-25
Fix inconsistent escaping from expand_escapes

### LINKED ISSUE #974: Inconsistent escaping/unescaping of quotes
Cc: @osm, @ailin-nemui

When `expand_escapes` is enabled, completions are escaped using `escape_string()`. (This was implemented in #709.)
`escape_string()`, among others, escapes `'` to `\'` and `"` to `\"`.
However, `\'` and `\"` are *not* expanded by the `expand_escapes()` function. So you end up sending spurious backslash characters.

ailin-nemui:
This reverts commit 62fd3ac1804d5d45bc2302ceac9ee9c7ef372e1a.

We need the fine-grained escape because expand_escapes is different
from quoted command argument expansion. The completion should only
escape the backslash.

Fixes #974


src/core/misc.c
```diff
@@ -741,15 +741,16 @@ int expand_escape(const char **data)
 	}
 }
 
-/* Escape all '"', "'" and '\' chars with '\' */
-char *escape_string(const char *str)
+/* Escape all the characters in `what' with a backslash */
+char *escape_string(const char *str, const char *what)
```
dequis: This looks API-breaking (unless you're 200% certain no modules use this)

ailin-nemui: yes, I resurrected this from https://github.com/irssi/irssi/pull/658/commits/027acffb4208d7e6ba8e229cbf6c3dae6f5dabaf

should we add another escape function instead?

ailin-nemui:
@dequis ping

src/core/misc.c
dequis: Sure, can't imagine any other way.

dequis:
So hmmm there's escape_string() which escapes quotes and backslashes, which sucks as a name but we can't change. And now we have escape_string_escape() which only escapes backslashes and the name is similarly ambiguous. `escape_string_backslashes`? If we're going for single-parameter functions that don't have a "what" parameter that is.

ailin-nemui:
ok...

PR #979, opened 2018-11-25
add keep_word setting to completion

ailin-nemui:
This setting can be used to put the word you are completing into the
completion list itself


ailin-nemui:
@irssi/developers

PR #985, opened 2018-12-07
automake warns that subdir-objects will be enabled in 2.0

ailin-nemui:
this was mentioned in #975


ailin-nemui:
@irssi/developers 
I didn't notice any issues with this sooo

PR #988, opened 2019-01-09
show output of git describe as PACKAGE_VERSION

ailin-nemui:
also fixes GNUism in sed call

reported by Zed`


dwfreed:
Note that this changes PACKAGE_VERSION from `1.2-gc7ce0f32` to `1.2-dev-320-gc7ce0f32`

ailin-nemui:
on bsd it would incorrectly show 1.2-1.2-dev-320-gc7ce0f32 due to the lack of `\+` support. I think showing it like this would be fine and even more insightful

ailin-nemui:
@irssi/developers

PR #1000, opened 2019-01-30
Store window bounds type

ailin-nemui:
extension of #999 to additionally take a note of the bounds' types if they were queries


ailin-nemui:
@irssi/developers

Issue #997, opened 2019-01-24
some crash


ailin-nemui:
CRITICAL (recursed) **: strarray_find: assertion 'item != NULL' failed

tried to make windows with sticky servers, may or may not be related

ailin-nemui:
```
(gdb) bt
#18 0x00007f2132ba60df in g_log (log_domain=log_domain@entry=0x0, log_level=log_level@entry=G_LOG_LEVEL_CRITICAL, format=format@entry=0x7f2132bf4320 "%s: assertion '%s' failed") at gmessages.c:1413
#19 0x00007f2132ba6909 in g_return_if_fail_warning (log_domain=log_domain@entry=0x0, pretty_function=pretty_function@entry=0x4b8158 <__func__.12656> "strarray_find", expression=expression@entry=0x4a5af3 "item != NULL") at gmessages.c:2762
#20 0x000000000048d137 in strarray_find (array=array@entry=0x2dc3e90, item=0x0) at ../../../src/core/misc.c:156
#21 0x000000000044866c in strarray_find_dest (array=0x2dc3e90, dest=dest@entry=0x7ffe30d50560) at ../../../../src/fe-common/core/fe-common-core.c:485
#22 0x000000000045e6b8 in sig_hilight_text (dest=0x7ffe30d50560, msg=0x2ed2880 "\004\071/-\004g!\004\071/-\004g index Help topics available to users:") at ../../../../src/fe-common/core/window-activity.c:91
#23 0x000000000049a912 in signal_emit_real (rec=rec@entry=0x24d6d60, params=params@entry=3, va=va@entry=0x7ffe30d50298, first_hook=<optimized out>) at ../../../src/core/signals.c:242
#24 0x000000000049aef8 in signal_emit_id (signal_id=<optimized out>, params=params@entry=3) at ../../../src/core/signals.c:304
#25 0x00000000004599e0 in print_line (dest=dest@entry=0x7ffe30d50560, text=text@entry=0x2eb0b40 "index Help topics available to users:") at ../../../../src/fe-common/core/printtext.c:182
#26 0x000000000045a280 in printformat_module_dest_charargs (module=module@entry=0x4a7b7f "fe-common/irc", dest=dest@entry=0x7ffe30d50560, formatnum=formatnum@entry=115, arglist=arglist@entry=0x7ffe30d503f0) at ../../../../src/fe-common/core/printtext.c:80
[...]

PR #1003, opened 2019-02-02
correct fix for #997

ailin-nemui:


ailin-nemui:
is this better @irssi/developers

PR #1004, opened 2019-02-04
fix gui_input_get_extent

ailin-nemui:
it was causing a free of data with [transfer=none]

reported by @GinjaNinja32


ailin-nemui:
@irssi/developers

PR #1013, opened 2019-02-12
do not stop autolog_ignore_targets from matching itemless targets

### LINKED ISSUE #1012: autolog_ignore_targets cannot ignore non-windowed targets
there may be an issue here

https://github.com/irssi/irssi/blob/63cc937e67214b1f793bc794a7d6fd9721ae2085/src/fe-common/core/fe-common-core.c#L488-L491

reported by @blabber on IRC

ailin-nemui:
fixes #1012


blabber:
Works for me :+1:

PR #1014, opened 2019-02-13
fix test on Big Endian 64bit, due to pointer size mismatch

ailin-nemui:


ailin-nemui:
the code was wrong (casting a small size ptr to a larger one)

ailin-nemui:
reported via debian build infrastructure / @rhonda

ailin-nemui:
@irssi/developers

dwfreed:
The fix makes sense, but why are you making a block here? I'd just put the variable declaration at the top of the function with the others.

ailin-nemui:
it's just a test.. but you wouldn't want any access to the temporary buf_len, could lead to wrong usage

PR #1016, opened 2019-02-15
Fix finding OpenSSL when 1.1 with deprecated APIs disabled

neheb:
SSL_library_init is a deprecated function. OPENSSL_init_ssl is not in 1.0.2.
SSL_CTX_new is in both.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -319,7 +319,7 @@ PKG_CHECK_MODULES([OPENSSL], [openssl], [
 	CFLAGS="$CFLAGS $OPENSSL_CFLAGS"
 	LIBS="$LIBS $OPENSSL_LIBS"
 ], [
-	AC_CHECK_LIB([ssl], [SSL_library_init], [
+	AC_CHECK_LIB([ssl], [SSL_CTX_new], [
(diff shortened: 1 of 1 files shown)


ailin-nemui:
@neheb any reason pkg-config doesn't work on your system?

neheb:
No idea. I just assumed as a different project needed a similar fix. irssi compiles with and without deprecated APIs.

ailin-nemui:
the logic is like this: 

```
IF { Openssl can be found with pkg-config } THEN {
    use pkg-config values
} ELSE {
    try to find old ssl with SSL_library_init (because there was no pkg-config at that time)
}
```

so your patch should not be needed if pkg-config works. that's what I am trying to understand

neheb:
Hrm so the patch is for the fall back path.

dequis:
So if I'm understanding this right, this was never actually broken, but a different project used a similar pattern without the pkg-config step, and this is applying the same fix even though the relevant path is never reached.

ailin-nemui:
@neheb any clarification here?

neheb:
I'll just close this

Issue #940, opened 2018-09-17
C module packaging effort


ailin-nemui:
wip testground: https://build.opensuse.org/project/show/home:ailin_nemui:testtest

todo:

- [x] :heavy_check_mark: dcc-send-limiter
- [x] :heavy_check_mark: item-sendqueue
- [x] :heavy_check_mark: python2
- [x] :heavy_check_mark: tcl
- [x] :heavy_check_mark: theme-indent
- [x] :heavy_check_mark: icb
- [x] :heavy_check_mark: xmpp
- [ ] silc
- [ ] quassel
- [x] :heavy_check_mark: robustirc
- [x] :heavy_check_mark: fish
- [x] :heavy_check_mark: ignore_oc
- [ ] xirssi

PR #1017, opened 2019-02-18
Remove unused functions

horgh:
I noticed a bug in one of these functions but then saw that it wasn't actually even used. So I figure it is better to delete!

These are public so perhaps we don't want to merge yet.

**Edited by Nei**
Functions to be removed:
- NET_CALLBACK
- NET_HOST_CALLBACK
- RESOLVED_NAME_REC
- net_gethostbyaddr_nonblock
- net_connect_nonblock
- [ SIMPLE_THREAD_REC, simple_init, simple_readpipe ]

### DIFF
--- a/src/core/net-nonblock.c
+++ b/src/core/net-nonblock.c
@@ -63,7 +63,7 @@ int net_gethostbyname_nonblock(const char *addr, GIOChannel *pipe,
 	/* child */
 	srand(time(NULL));
 
-        memset(&rec, 0, sizeof(rec));
+	memset(&rec, 0, sizeof(rec));
 	rec.error = net_gethostbyname(addr, &rec.ip4, &rec.ip6);
 	if (rec.error == 0) {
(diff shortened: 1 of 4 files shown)


horgh:
I checked all modules I could find and none use these functions. (I checked all in https://github.com/irssi/irssi/issues/940 and all listed on irssi.org except psyc as I couldn't find that one's code).

ailin-nemui:
sounds good to me, we should bump ABI constant before merge

ailin-nemui:
@irssi/developers

dequis:
Can we at least go through deprecation in one release? And maybe stub out the contents of the functions, but don't break API just yet

ailin-nemui:
my opinion is that we should skip deprecation
- irssi is not so popular
- public modules do not use these functions

ailin-nemui:
@irssi/developers please vote on the first post

PR #1020, opened 2019-02-23
allow rawlog-buffer to shrink again after lowering "rawlog_lines"

teschi:
It seems that when setting the size of the rawlog-buffer (using "rawlog_lines") to a value lower the number of lines already in the buffer, the buffer will never shrink again.
This change allows multiple lines to be popped from the queue.

### DIFF
--- a/src/core/rawlog.c
+++ b/src/core/rawlog.c
@@ -62,7 +62,7 @@ void rawlog_destroy(RAWLOG_REC *rawlog)
 /* NOTE! str must be dynamically allocated and must not be freed after! */
 static void rawlog_add(RAWLOG_REC *rawlog, char *str)
 {
-	if (rawlog->lines->length >= rawlog_lines && rawlog_lines > 0) {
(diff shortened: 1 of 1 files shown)


ailin-nemui:
we had briefly discussed this is #957

teschi:
Sorry for replying late (I completely missed the notification).

My use-case was recording on IRC level into files, and rotating the raw-logs from inside irssi.
I had issues that the new logfile ends up having duplicates of the buffer, so i basically wanted to set the buffer to a single line.
I did play around for a bit (mainly in the config files), but couldn't get it smaller.
(There might be ways to do this without that code change, but as I already modified the log-format for my purpose, I just also added the ability to shrink)

I primarily raised the pull request because the behavior of: only ever removing one item after adding one and therefore never being able to shrink the buffer, "felt" a bit wrong. And removing items in a loop as long as the buffer is larger than it's maximum seemed a bit more straight-forward to me.

I don't have a strong opinion about this, as the change is kinda cosmetic, therefore I don't have any issue if it get's rejected.

PR #1022, opened 2019-03-13
Implement a method to delete messages older than given time in scroll…

heikkiorsila:
Implement a method to delete messages older than given time in scrollback history.

The maximum age of messages is controlled by /set scrollback_max_age variable.
E.g. /set scrollback_max_age 7days

Rationale: I created this patch to let my client forget messages in a private chat that I don't want to keep around. In practice, "day changed" messages force an effective event to delete old messages when no new messages have been sent to a chat.


ailin-nemui:
@irssi/developers

ailin-nemui:
@heikkiorsila you can consider contributing a description for https://github.com/irssi/irssi.github.io/blob/master/documentation/settings.markdown

PR #1025, opened 2019-03-22
Add setting to choose between window name and visible name in actlist

vague666:


src/fe-text/statusbar-items.c
```diff
@@ -123,7 +123,9 @@ static char *get_activity_list(MAIN_WINDOW_REC *window, int normal, int hilight)
 						 window->hilight_color,
 						 window->refnum);
 		if (add_name && window->active != NULL)
-			g_string_append_printf(format, ":%s", window->active->visible_name);
+			g_string_append_printf(format, ":%s",
+				settings_get_bool("actlist_prefer_window_name") &&
```
ailin-nemui: please cache the setting value before the loop

src/fe-text/statusbar-items.c
```diff
@@ -76,7 +76,7 @@ static char *get_activity_list(MAIN_WINDOW_REC *window, int normal, int hilight)
 	GString *format;
 	GList *tmp;
         char *ret, *name, *value;
-        int is_det;
+        int is_det, pref_name;
 	int add_name = settings_get_bool("actlist_names");
```
ailin-nemui: here is outside the loop :-)

PR #1026, opened 2019-03-22
add access to set hidden_level from perl

vague666:
revisit!
requires $view->redraw to update view


ailin-nemui:
lgtm. bump ABI in common.h and see if there are any other textbuffer_view_ methods you want exposed?

ailin-nemui:
@irssi/developers

dequis:
Yeah sure why not

Issue #1029, opened 2019-04-01
Improve SSL Documentation


blackthornedk:
Since #207 was done, the documentation on https://irssi.org/documentation/help/server/ was only partially updated, and https://irssi.org/documentation/settings/ doesn't mention ssl_* or tls_* at all.

Also it was not clear to me, that tls_verify is not enabled by default. What does that mean? Were no certificate checks done at all on my connections, before now?

ailin-nemui:
> Also it was not clear to me, that tls_verify is not enabled by default. What does that mean? Were no certificate checks done at all on my connections, before now?

indeed

by the way you can get involved and submit a PR with updated docs :-)

blackthornedk:
> by the way you can get involved and submit a PR with updated docs :-)

I already sent in https://github.com/irssi/irssi/pull/1030 but I don' t know where the website docs are located.

blackthornedk:
Updated with https://github.com/irssi/irssi.github.io/pull/63
But I have no idea how to fix https://irssi.org/documentation/settings/
I'm a bit confused about the servers / server sections of the last file.

ailin-nemui:
Cool! Please have a look at the comment on top of https://github.com/irssi/irssi.github.io/blob/ef6f263ff09eb663bf563c81dbe0bdface8edb9b/documentation/help/server.markdown

The change has to be submitted to https://github.com/irssi/irssi/blob/master/src/fe-common/irc/fe-irc-server.c instead!

What's wrong about the settings docs? I think there are no TLS related settings other than https://irssi.org/documentation/settings/#tls_verbose_connect

blackthornedk:
> Cool! Please have a look at the comment on top of https://github.com/irssi/irssi.github.io/blob/ef6f263ff09eb663bf563c81dbe0bdface8edb9b/documentation/help/server.markdown
> 
> The change has to be submitted to https://github.com/irssi/irssi/blob/master/src/fe-common/irc/fe-irc-server.c instead!

OK, I updated #1030 instead. 
So no changes are needed to the other repository, for the website to be updated?

> What's wrong about the settings docs? I think there are no TLS related settings other than https://irssi.org/documentation/settings/#tls_verbose_connect

I think I misunderstood the purpose of 'settings'. I read them as a definition file for the .irssi/config file, and not as general purpose settings. As I read over them again, I suspect it to be a definition of the various settings used by the /SET command?

ailin-nemui:
Nice!

[...]

PR #1030, opened 2019-04-01
Add tls_* options to manual

blackthornedk:
Add tls_* options to the SERVER section of the manual.
Part of solution to #1029


ailin-nemui:
@irssi/developers

PR #1036, opened 2019-04-06
Check return value of BIO_read() in set_pubkey_info()

### LINKED ISSUE #943: Missing Error Check in function BIO_read()
Hi,
Function BIO_read() returns either the amount of data successfully read (if the return value is positive) or that no data was successfully read if the result is 0 or -1. If the return value is -2 then the operation is not implemented in the specific BIO type. However, function BIO_read() in  irssi/src/core/network-openssl.c  didn't check the return value (non-nagetive) , before passing the return value as an argument to the array named buffer . See the following code:

line: 650,651
line: 658,659
screenshot from 2018-09-19 11-54-24

Chi Li, Zuxing Gu, Jiecheng Wu

berkerpeksag:
Fixes #943

I tried to copy what Python does in the ssl module:
 https://github.com/python/cpython/blob/2ea8099523581cf2ecc060831a53debb57ff98ee/Modules/_ssl.c#L4733-L4738

Please let me know if you want me to update `NEWS` file.

### DIFF
--- a/src/core/network-openssl.c
+++ b/src/core/network-openssl.c
@@ -601,18 +601,19 @@ static void set_cipher_info(TLS_REC *tls, SSL *ssl)
 	tls_rec_set_cipher_size(tls, SSL_get_cipher_bits(ssl, NULL));
 }
 
-static void set_pubkey_info(TLS_REC *tls, X509 *cert, unsigned char *cert_fingerprint, size_t cert_fingerprint_size, unsigned char *public_key_fingerprint, size_t public_key_fingerprint_size)
+static gboolean set_pubkey_info(TLS_REC *tls, X509 *cert, unsigned char *cert_fingerprint, size_t cert_fingerprint_size, unsigned char *public_key_fingerprint, size_t public_key_fingerprint_size)
 {
+	gboolean ret = TRUE;
 	EVP_PKEY *pubkey = NULL;
 	char *cert_fingerprint_hex = NULL;
 	char *public_key_fingerprint_hex = NULL;
 
 	BIO *bio = NULL;
 	char buffer[128];
-	size_t length;
+	ssize_t length;
 
-	g_return_if_fail(tls != NULL);
-	g_return_if_fail(cert != NULL);
+	g_return_val_if_fail(tls != NULL, FALSE);
+	g_return_val_if_fail(cert != NULL, FALSE);
 
 	pubkey = X509_get_pubkey(cert);
 
@@ -648,6 +649,11 @@ static void set_pubkey_info(TLS_REC *tls, X509 *cert, unsigned char *cert_finger
 	bio = BIO_new(BIO_s_mem());
 	ASN1_TIME_print(bio, X509_get_notBefore(cert));
 	length = BIO_read(bio, buffer, sizeof(buffer));
+	if (length < 0) {
+		ret = FALSE;
+		BIO_free(bio);
+		goto done;
+	}
 	buffer[length] = '\0';
 	BIO_free(bio);
 	tls_rec_set_not_before(tls, buffer);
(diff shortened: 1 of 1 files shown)


ailin-nemui:
@ahf

src/core/network-openssl.c
```diff
@@ -648,6 +648,10 @@ static void set_pubkey_info(TLS_REC *tls, X509 *cert, unsigned char *cert_finger
 	bio = BIO_new(BIO_s_mem());
 	ASN1_TIME_print(bio, X509_get_notBefore(cert));
 	length = BIO_read(bio, buffer, sizeof(buffer));
+	if (length < 0) {
+		g_free(buffer);
+		return FALSE;
```
dequis: This leaks cert_fingerprint_hex, public_key_fingerprint_hex, bio, pubkey, doing "goto error" is encouraged here. Also buffer is stack-allocated, doesn't need to be freed.

Same thing for the return FALSE below.

berkerpeksag: Good points, I've just addressed your comments. Let me know if you want me to manually squash these commits.

Thank you for the review!

src/core/network-openssl.c
```diff
@@ -601,18 +601,19 @@ static void set_cipher_info(TLS_REC *tls, SSL *ssl)
 	tls_rec_set_cipher_size(tls, SSL_get_cipher_bits(ssl, NULL));
 }
 
-static void set_pubkey_info(TLS_REC *tls, X509 *cert, unsigned char *cert_fingerprint, size_t cert_fingerprint_size, unsigned char *public_key_fingerprint, size_t public_key_fingerprint_size)
+static gboolean set_pubkey_info(TLS_REC *tls, X509 *cert, unsigned char *cert_fingerprint, size_t cert_fingerprint_size, unsigned char *public_key_fingerprint, size_t public_key_fingerprint_size)
 {
+	int ret = 1;
```
dequis: Might as well set it as gboolean and TRUE (which is the exact same type and value in practice, but yeah)

src/core/network-openssl.c
```diff
@@ -648,6 +649,10 @@ static void set_pubkey_info(TLS_REC *tls, X509 *cert, unsigned char *cert_finger
 	bio = BIO_new(BIO_s_mem());
 	ASN1_TIME_print(bio, X509_get_notBefore(cert));
 	length = BIO_read(bio, buffer, sizeof(buffer));
+	if (length < 0) {
+		ret = 0;
```
dequis: ...and FALSE here

src/core/network-openssl.c
```diff
@@ -656,13 +661,20 @@ static void set_pubkey_info(TLS_REC *tls, X509 *cert, unsigned char *cert_finger
...
 	buffer[length] = '\0';
 	BIO_free(bio);
 	tls_rec_set_not_after(tls, buffer);
 
+done:
 	g_free(cert_fingerprint_hex);
 	g_free(public_key_fingerprint_hex);
 	EVP_PKEY_free(pubkey);
+
+	return ret ? TRUE : FALSE;
```
dequis: ...and just "return ret" here

(the ternary makes sense in the other function that may return -1 or 0, but this one is straight gboolean)

src/core/network-openssl.c
```diff
@@ -648,6 +649,10 @@ static void set_pubkey_info(TLS_REC *tls, X509 *cert, unsigned char *cert_finger
 	bio = BIO_new(BIO_s_mem());
 	ASN1_TIME_print(bio, X509_get_notBefore(cert));
 	length = BIO_read(bio, buffer, sizeof(buffer));
+	if (length < 0) {
+		ret = 0;
+		goto done;
```
dequis: This leaks bio, same with the other goto.

berkerpeksag: Which way would you prefer to address this? Add `BIO_free(bio);` calls inside the if branches or move `BIO_free(bio);` inside the goto?

src/core/network-openssl.c
berkerpeksag: Done in https://github.com/irssi/irssi/pull/1036/commits/a6ade0bc96e32a14cc864cbe0e143e043fc8583a.

src/core/network-openssl.c
dequis: Either would work, but considering the scope of bio (it refers to two different allocations depending on where you are in the function) it might make more sense to have it inside the ifs.

berkerpeksag: Done, thank you again!

berkerpeksag:
Do you need anything else from me? Or are we waiting for another pair of eyes?

ailin-nemui:
irssi project is short on maintainers/reviewers. thanks for your patience

berkerpeksag:
@ailin-nemui thank you for the invite!

berkerpeksag:
Just a friendly ping. Please let me know if you need anything from me in order to get this merged. Happy holidays!

PR #1037, opened 2019-04-06
Move NULL check for cert above pubkey

### LINKED ISSUE #944: Wrong Error Check position in function SSL_get_peer_certificate()
Hi,
Function SSL_get_peer_certificate() returns a pointer to an X509 certificate on success or NULL on error. However, the function SSL_get_peer_certificate() in irssi/src/core/network-openssl.c didn't check the return value is NULL or not, before passing the return value as an argument to the function X509_get_X509_PUBKEY().  See the following details:

screenshot from 2018-09-19 16-08-57

=================================================================================

X509_get_X509_PUBKEY() doesn't check the argument which is the return value of function SSL_get_peer_certificate() is NULL or not before using it.

Ref: https://github.com/openssl/openssl/blob/master/crypto/x509/x509_set.c
screenshot from 2018-09-19 16-13-47

=================================================================================

Anyhow, the safest behavior  is checking the return value of function SSL_get_peer_certificate() first , then passing the return value as an argument to function X509_get_X509_PUBKEY(). See the following details:

cert = SSL_get_peer_certificate(chan->ssl);
if (cert == NULL) {
	g_warning("TLS server supplied no certificate");
	ret = 0;
	goto done;
}
pubkey = X509_get_X509_PUBKEY(cert);

=================================================================================

Chi Li, Zuxing Gu, Jiecheng Wu

berkerpeksag:
This patch was originally proposed by Chi Li.

Fixes #944

### DIFF
--- a/src/core/network-openssl.c
+++ b/src/core/network-openssl.c
@@ -854,14 +854,13 @@ int irssi_ssl_handshake(GIOChannel *handle)
 	}
 
 	cert = SSL_get_peer_certificate(chan->ssl);
-	pubkey = X509_get_X509_PUBKEY(cert);
-
 	if (cert == NULL) {
 		g_warning("TLS server supplied no certificate");
 		ret = 0;
 		goto done;
 	}
 
+	pubkey = X509_get_X509_PUBKEY(cert);
 	if (pubkey == NULL) {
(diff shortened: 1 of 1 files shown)


ailin-nemui:
@irssi/developers

PR #1043, opened 2019-05-13
fix realpath on old solaris

### LINKED ISSUE #1042: /SAVE fails in Solaris 10 "Invalid argument"
For some reason Solaris 10 does not accept NULL as a parameter.  I remembered that this was an issue while ago with GNU screen also

ERRORS
     The realpath() function will fail if:

     EINVAL          Either the file_name or resolved_name  argument is a null pointer.

src/lib-config/write.c

	/* expand all symlinks; else we may replace a symlink with a regular file */
	dest_name = realpath(base_name, NULL);

It was fixed with this

	char realbuf[PATH_MAX];
	dest_name = realpath(base_name, realbuf);

After this I'm able to use /SAVE command again

ailin-nemui:
POSIX.1-2001 did not implement realpath(..., NULL) yet.
Fixes #1042

### DIFF
--- a/src/lib-config/write.c
+++ b/src/lib-config/write.c
@@ -315,6 +315,15 @@ int config_write(CONFIG_REC *rec, const char *fname, int create_mode)
 	/* expand all symlinks; else we may replace a symlink with a regular file */
 	dest_name = realpath(base_name, NULL);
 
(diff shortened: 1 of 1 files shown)


ailin-nemui:
@irssi/developers

PR #1046, opened 2019-05-22
Implement social media style @nick completion.

Tronic:
Autocomplete "@nick " instead of traditional IRC style "nick: "
enabled by /set completion_char @. Use nick<tab> as usual.


vague666:
There's already script to accomplish this so no need to bloat irssi
https://github.com/irssi/scripts.irssi.org/blob/master/scripts/slack_complete.pl
https://github.com/irssi/scripts.irssi.org/blob/master/scripts/complete_at.pl

PR #1048, opened 2019-05-28
Make it possible to list specific section with /set

vague666:


dwfreed:
Perhaps support partial section names?  Like how /set supports partial key names for settings.

Issue #671, opened 2017-03-11
Remove support for non-UTF8 terminals


ailin-nemui:
We are planning to remove 8-bit and Chinese support from Irssi.

Interaction with legacy IRC channels would still be provided through /recode, as it is currently.

However, Irssi would stop working on non-UTF-8 terminals (or at least appear heavily glitched)

joneskoo:
I think that using only Unicode sounds like a great idea in 2017. 🎉 

For the record (while I wouldn't use it), shouldn't it still be possible to use _screen_ to translate to a legacy encoding, even if irssi only implements UTF-8? For the one or two people who have some legacy terminal.

horgh:
It seems like a good idea. The world seems to be moving to UTF-8 by default/only. Plus, if this lets the codebase be simplified, that's great! Deleted code is debugged code!

joneskoo:
I just thought of another reason than simplification why this is a great
idea.

If irssi rejects to start in a wrong encoding environment, that makes it
more user friendly. There are less things to set up and you need to do the
right thing (allows any characters, no corruption) before it works. It will
prevent messing up rendering due to encoding mismatch. A simple error maybe
with a link to irssi website? (Can update without a release.)

If there's no alternate configuration, it will be hard for various "stack
overflow" answers to instruct using legacy encoding leading to hard to
understand encoding errors later.

timur-enikeev:
If it's harder to support many encodings, it is correct decision.

saahriktu:
Very grievous. I prefer to use irssi with the singlebyte locale (KOI8-R) which perfectly fits to my framebuffer terminal. Singlebyte encodings have their own advantages such as low memory usage, fast processing and strong classic API (fgetc(), fputc(), printf(),... , "str + N" to move the pointer by N chars,... ).

dequis:
@saahriktu Nice, glad you've found this issue. Can you provide some more details about your setup?

1. What distro?
2. How do you set up a KOI8-R framebuffer terminal? Is it console-setup stuff or some third party software?
3. Are there any IRC networks that primarily use KOI8-R?
4. Do you use the "recode" feature?
5. Please post the output of the 'locale' command

>Singlebyte encodings have their own advantages such as low memory usage, fast processing and strong classic API (fgetc(), fputc(), printf(),... , "str + N" to move the pointer by N chars,... ).

[...]

PR #1057, opened 2019-06-17
restore 8bit support in input entry

### LINKED ISSUE #1018: Input line is echoing utf8 chars on non-utf8 terminal
I have a legacy Linux installation that still uses ISO-8859-1 charset. Commit 9072e988194a36186bd98db844b8285e3f37069a causes irssi to erroneously echo utf8 chars to the input line (start irssi, press 'ä' key, irssi outputs 'Ã¤' on the input line, cursor is placed between the two chars) instead of ISO-8859-1. Everything else seems to work fine, with no encoding problems when sending/receiving text from/to IRC channels. Only input line is affected.

ailin-nemui:
This should fix #1018

Regression of #764

Based on patch by @saahriktu

### DIFF
--- a/src/fe-text/gui-entry.c
+++ b/src/fe-text/gui-entry.c
@@ -379,11 +379,19 @@ static void gui_entry_draw_from(GUI_ENTRY_REC *entry, int pos)
 		if (new_xpos > end_xpos)
 			break;
 
-		if (entry->hidden)
+		if (entry->hidden) {
                         g_string_append_c(str, ' ');
-		else if (unichar_isprint(c))
-			g_string_append_unichar(str, c);
-		else {
+		} else if (unichar_isprint(c)) {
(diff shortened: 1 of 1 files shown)


ailin-nemui:
at least until #671 is finished

dequis:
Kinda weird to check `unichar_isprint` against a character that might not be unicode after all, but i suspect this isn't the only place where that is done. Would be nice to know how to test this.

ailin-nemui:
I agree and I haven't tested this, only followed similar code

ailin-nemui:
@irssi/developers

suiited:
mute the thread

Sent with [ProtonMail](https://protonmail.com) Secure Email.

‐‐‐‐‐‐‐ Original Message ‐‐‐‐‐‐‐

Issue #730, opened 2017-07-15
-tls_capath and -tls_cafile imply -tls_verify, break -notls_verify and -notls


dequis:
`-notls` covers the `-tls_verify` implication, but neither of them cover this `-tls_capath`/`-tls_cafile` implication.

Also, there's no way to get rid of `-tls_capath`/`-tls_cafile` without editing the config because the empty string is equivalent to not changing it.

An extension of #117, not quite a regression but an edge case not covered by the new feature.

Example:

`/server add -tls_verify example.net`

```
  {
    address = "example.net";
    port = "6667";
    use_tls = "yes";
    tls_verify = "yes";
  }
```

`/server modify -notls example.net`

```
  {
    address = "example.net";
    port = "6667";
    use_tls = "no";
    tls_verify = "no";
  }
```

So far so good.

Now add capath.

`/server modify -tls_capath /foo/bar example.net`

```
  {
    address = "example.net";
    port = "6667";
    use_tls = "yes";
    tls_verify = "yes";
    tls_capath = "/foo/bar";
  }
```

The following don't do anything:

`/server modify -tls_capath '' example.net`

`/server modify -notls_verify example.net`

`/server modify -notls example.net`

PR #1060, opened 2019-06-29
disable tls when using -notls

vague666:


vague666:
hopefully a #730 fix

vague666:
https://github.com/irssi/irssi/pull/1060/files#diff-820acea5c23ba424526175f91ccd42acL153 needs a bit of thought I reckon, not sure how to handle it

ailin-nemui:
I think we should keep these entries and just not use them

ailin-nemui:
we may need to check why they cannot be unset(?)

vague666:
Not sure what you mean

ailin-nemui:
> -tls_capath/-tls_cafile implication.

my requirement: do not unset tls_capath (like you do in this patch), instead, do not enable TLS unless tls or tls_verify is given. That way, you can "keep" the tls_capath setting and only enable/disable the tls for a server, and  do not have to repeat the tls_capath setting when you enable tls again

> Also, there's no way to get rid of -tls_capath/-tls_cafile without editing the config because the empty string is equivalent to not changing it.

this might need to be fixed, so that -tls_capath '' does remove the capath

vague666:
Ok, I've added -tls_capath '' et al so tls options can be removed with empty string
Where should I look to check for use_tls to disable tls options if use_tls is FALSE?
So far I've added a check to https://github.com/irssi/irssi/blob/master/src/core/servers-setup.c#L170
but all the different code paths makes me uncertain in case I am missing something

ailin-nemui:
@vague666 can you check if this is fixed in latest master?

PR #1063, opened 2019-07-03
reapply a theme to previous formats

ailin-nemui:


ailin-nemui:
todos:
- automate it
- smart match on /format changes
- use the new feature to implement server_time
- what to do about expandos? now they are lost/reevaluated
- fix/hide it from buf.pl
- tighten up code when removing lines (e.g from perl)

cool:
- change nick width on the fly
- test themes

vague666:
Great work! That's a shitload of code you've produced :)

ailin-nemui:
to test, run /scrollback redraw

ailin-nemui:
## discussion from irc
- reimplement glib code, do not require newer glib
- design doesn't work, wastes 2x the memory (instead: need to avoid storing rendered lines)

dequis:
I think a design that only stores a limited, tiny amount of lines also wouldn't need reimplementing glib stuff, as a bunch of strdup wouldn't hurt in that case

ailin-nemui:
superseded by #1079

PR #1072, opened 2019-07-18
Add "automatically generated by configure" comments to source

dequis:
Since running ./configure now changes git-tracked source files to have
this line, might as well have that git tracked so it doesn't create
local changes all the time.


ailin-nemui:
good point, instead I need to rip out this part of the code from configure.ac :)

ailin-nemui:
otherwise we will desync

dequis:
The part I don't quite get is when this change happened, there's not much going on configure.ac

PR #1076, opened 2019-07-27
Free g_strdup() allocated memory with g_free()

l2dy:
Memory allocated with g_malloc() must be freed with g_free(). The
allocators may use different memory pools.

See: https://developer.gnome.org/glib/stable/glib-Memory-Allocation.html


ailin-nemui:
thanks

PR #1078, opened 2019-07-27
fix cut off text with theme-indent and /set indent_always off

ailin-nemui:


dequis:
How do I repro this issue?

ailin-nemui:
add
```
indent_default = "...something...";
```
to your theme `abstracts`

```
/load theme_indent
/set indent_always off
```

check that wrapped long words (e.g. URLs ) get overwritten by the indentation (without this fix)

What I hope the fix does (: disable the callback and the indent function for line continuations and long words

ailin-nemui:
@irssi/developers

PR #1079, opened 2019-08-07
reapply a theme to previous formats

ailin-nemui:
- [x] fix /set hide_colors 
- [x] fix line IDs on perl side
- [x] reimplement the colour stripping in line2text


ailin-nemui:
another attempt at #1063

ailin-nemui:
- [x] reimplement glib code, do not require newer glib

- [x] design doesn't waste twice the memory, uses cache and original formats

- completely removed the old textbuffer representation ( https://github.com/shabble/irssi-docs/wiki/Notes-256-Colour#textbuffer-encoding )
  - this may have some drawbacks:
    - texts are no longer NUL-safe
    - additional escaping routines would need to be introduced if text is supposed to contain formatting codes
  - benefits
    - one less layer of indirection
    - NUL-terminated strings, no weird issues with cut off UTF8 (although we fixed that)
    - simpler code

- the textbuffer-formats is an extra module, so if we unhook the signals it should go back to the "old way" of storing pre-rendered text

#### Todos

- [ ] differentiate between deleted formats and empty lines
- [ ] fix all the bugs

----

- we should check if the removal of the optimised textbuffer has any visible performance implications

vague666:
first-off: autotools compiled successfully
then, I have a lot of blinking elements after applying the patch
https://vague.se/pastes/2019-08-07_11:41:58_1920x1080_scrot.png
https://vague.se/pastes/2019-08-07_11:42:26_1920x1080_scrot.png
I'll test reapplying themes next

vague666:
`/scrollback redraw` seems to work well, too bad it's not done automatically yet :)

ailin-nemui:
-> under what circumstances would we "want" an automated scrollback redraw?

vague666:
`/set theme` and `/format ....` imo, and `/window theme`
basically in all instances :) 
Maybe control it with a setting

ailin-nemui:
@vague666 now with less blink

dequis:
You kinda got unrelated stuff in this branch. Also, nice.

dequis:
...I guess that's one way to make the changes not unrelated

Seriously though this PR was already massive enough that it's very hard to give it a proper review, adding a bunch of new features on top doesn't help :(

Feel free to have a kitchen sink branch to experiment with, but don't make that this branch. I'd rather aim to get the original feature reviewed and merged.

ailin-nemui:
this now depends on #1089 

the code is mostly split into 3 parts;
-  ripping out the old textbuffer implementation and storing the line texts and formats
-  the expando caching in special-vars
-  the perl code

ailin-nemui:
you should be able to review each commit individually, but  it's still a big hairy lump

for the view_update_line_cache in textbuffer-view.c, it would make sense to review with white-space ignored (?w=1 )

ailin-nemui:
@irssi/developers please test

vague666:
There's some drawing issues
tested with 1079.diff and  Irssi v1.2-dev-522-g991e24fb-dirty and themes https://irssi-import.github.io/themes/elite.theme and https://bitbucket.org/realnot/dirssi/src/master/
type | elite | dirssi
-----|------|-----
patched | elite.theme patched | dirssi.theme patched
unpatched | elite.theme unpatched | dirssi.theme unpatched

Also, I'd still like the scrollback to be redrawn automatically if possible when theme is applied

ailin-nemui:
thank you very much for the detailed report. btw the auto redraw will come, just not in this PR!

ailin-nemui:
I think I fixed dirssi, still need to check elite

ailin-nemui:
now I think that should fix elite

vague666:
Seems to work as expected now with the themes I've tested :)

ailin-nemui:
@irssi/developers last chance to review....

PR #1080, opened 2019-08-07
WIP for saving only channel-window layout

vague666:
Need some opinions on how to go forward with this


dequis:
What does "saving only channel-window layout" mean?

vague666:
save only layout of items that are channels if setting is true (realised now channel-window was bad wording by me :) )

ailin-nemui:
the code you wrote, looks like it wouldn't save windows if there were no items? are you saying your code doesn't work?

PR #1083, opened 2019-08-09
migrate workflow to yml

ailin-nemui:


ailin-nemui:
heh. It doesn't work (?) so maybe we need to wait for github somehow. let's visit this again later

ailin-nemui:
superseded by #1103

PR #1086, opened 2019-08-13
increase disconnect buffer and get rid of some magic 510 values

ailin-nemui:
this is also in preparation for a configurable IRC line length


ailin-nemui:
with tags support, the disconnect buffer wouldn't possibly fit "one complete line until the \r\n", so it's safer to increase this a bit

PR #1088, opened 2019-08-13
add a new "print noformat" signal to be emitted on printtext

ailin-nemui:
this will be used together with "print format" in the reformattable
textbuffer implementation


ailin-nemui:
@irssi/developers

PR #1089, opened 2019-08-13
implement reference counted strings

ailin-nemui:
- on glib >=2.58, use the implementation provided by glib
- otherwise, a hash table will contain the strings

this data type is used to save space in the upcoming reformattable
textbuffer implementation


src/fe-text/textbuffer-commands.c
```diff
@@ -358,6 +359,13 @@ static void cmd_scrollback_status(void)
 	printtext(NULL, NULL, MSGLEVEL_CLIENTCRAP,
 		  "Total: %d lines, %dkB of data",
 		  total_lines, (int)(total_mem / 1024));
+	{
+		char *tmp = i_refstr_table_size_dbg();
```
dequis: So this function seems to be more than a debug feature but part of a user facing feature, can you rename the _dbg part? (no opinion on new names)

(shame about the feature not existing when using glib's implementation, which isn't too different, just thread safe. which... we don't really care about? oh no. what if we just use our own)

ailin-nemui: in my eyes the scrollback memory feature is sort of a debug thing. I will think of a better function name :D

PR #1090, opened 2019-08-13
incoming irc message tags support

ailin-nemui:
this is just the very bare bones to accept and ignore incoming
tags. further patches can build upon it

original implementation by @dequis #576

### DIFF
--- a/src/irc/core/irc.c
+++ b/src/irc/core/irc.c
@@ -32,6 +32,7 @@
 char *current_server_event;
 static int signal_default_event;
 static int signal_server_event;
+static int signal_server_event_tags;
 static int signal_server_incoming;
 
 #ifdef BLOCKING_SOCKETS
@@ -158,6 +159,11 @@ static char *split_nicks(const char *cmd, char **pre, char **nicks, char **post,
 
 	*pre = g_strdup(cmd);
 	*post = *nicks = NULL;
+
+	if (**pre == '@') { /* message-tags */
+		arg++;
+	}
+
 	for (p = *pre; *p != '\0'; p++) {
 		if (*p != ' ')
 			continue;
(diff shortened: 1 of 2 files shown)


dequis:
Reminds me of #576 :D

ailin-nemui:
oops

ailin-nemui:
in #576, I wrote

> I would like to see actual use case implementation before proceeding here...

now, I'm posting the same patch again but I'm already working on use cases. Separate patch for easier review

dequis:
Yeah no worries, I didn't really have the time to do the whole thing (and still don't)

src/irc/core/irc.c
```diff
@@ -382,14 +406,14 @@ static char *irc_parse_prefix(char *line, char **nick, char **address)
...
+	char *nick, *address, *tags;
 
 	g_return_if_fail(server != NULL);
 	g_return_if_fail(line != NULL);
 
-	line = irc_parse_prefix(line, &nick, &address);
-	if (*line != '\0')
-		signal_emit_id(signal_server_event, 4, server, line, nick, address);
+	line = irc_parse_prefix(line, &nick, &address, &tags);
+	if (*line != '\0' || tags != NULL)
```
dequis: So wait, this is handling tags and no line like...

    @tags \r\n

I don't think that's a valid irc message (the command is always required) or something that makes sense to have special cases for.

src/irc/core/irc.c
```diff
@@ -158,6 +159,11 @@ static char *split_nicks(const char *cmd, char **pre, char **nicks, char **post,
 
 	*pre = g_strdup(cmd);
 	*post = *nicks = NULL;
+
+	if (**pre == '@') { /* message-tags */
+		arg++;
```
dequis: This one took me a while to understand, could use a longer comment

src/irc/core/irc.c
```diff
@@ -441,12 +465,14 @@ static void irc_init_server(IRC_SERVER_REC *server)
 void irc_irc_init(void)
 {
 	signal_add("server event", (SIGNAL_FUNC) irc_server_event);
+	signal_add("server event tags", (SIGNAL_FUNC) irc_server_event_tags);
```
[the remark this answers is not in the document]
ailin-nemui: Good catch

src/irc/core/irc.c
ailin-nemui: wouldn't it be better to preserve and make accessible such an invalid line nevertheless?

src/irc/core/irc.c
ailin-nemui: I added a comment

dequis: Sounds good

src/irc/core/irc.c
dequis: Yeah, I don't know. I guess it doesn't really make other things more complicated and it's harmless to do so.

PR #1091, opened 2019-08-13
enable irc cap ls 302

ailin-nemui:
this is needed to learn about the values of capabilities, otherwise
they are not sent for backwards compatibility reasons

the values are required to support for example the maximum line length
value (upcoming patch)

also change the strings to #defines for easier handling


dequis:
>also change the strings to #defines for easier handling

Easier how? Locality is underrated imo

ailin-nemui:
we may need to refer to those caps in further hash table lookups

PR #1092, opened 2019-08-13
sending of tag messages and customisable maximum irc message length

ailin-nemui:


src/irc/core/irc-servers.c
```diff
@@ -475,6 +478,14 @@ static void sig_server_quit(IRC_SERVER_REC *server, const char *msg)
 	g_free(recoded);
 }
 
+static void cap_maxline(IRC_SERVER_REC *server)
+{
+	int maxline = atoi(g_hash_table_lookup(server->cap_supported, CAP_MAXLINE));
```
dequis: `atoi` is a rather shitty function especially when given bad inputs. We have `parse_uint` in misc.c which wraps strtoul to have more reasonable error handling, go with that imo?

src/irc/core/irc.c
```diff
@@ -49,55 +49,87 @@ static void strip_params_colon(char *const);
...
+
 	if (server->cmdcount == 0)
 		irc_servers_start_cmd_timeout();
 	server->cmdcount++;
 
 	if (!raw) {
+		const char *tmp = cmd;
+
+		server_supports_tag = server->cap_supported != NULL &&
+			g_hash_table_lookup_extended(server->cap_supported, CAP_MESSAGE_TAGS, NULL, NULL);
```
dequis: You can have a message-tags section without the message tags cap being enabled, for example if you request the server-time, batch or account-tag caps. I think this entire check could just go, and we should handle the tags section unconditionally.

dequis: Wait shit this is when sending. This is why i should use the 'start a review' thing to hide my shame when I realize things seconds after sending a comment.

My comment might still apply though. Need to keep reading.

dequis: Yeah nevermind this entire thing is okay actually. It's enforcing the message-tags specific lengths for the tag section only, and doing tag-specific truncation which isn't a thing without the cap. Nevermind.

dequis:
I like what I see!

So this adds CAP_MESSAGE_TAGS to the .h file but doesn't request it, is that intended?

ailin-nemui:
it is, we use it to ensure if a server understands /expects tags as input, but we don't necessarily need to request tags going in our direction

that said we're probably going to request it in a future PR (soon)

dequis:
I guess it's easily testable with `/quote cap req message-tags`

ailin-nemui:
should be!

src/irc/core/irc-servers.c
ailin-nemui: cool

src/irc/core/irc.c
ailin-nemui: :-)

ailin-nemui:
thanks for review!

src/irc/core/irc-servers.c
```diff
@@ -476,6 +479,16 @@ static void sig_server_quit(IRC_SERVER_REC *server, const char *msg)
 	g_free(recoded);
 }
 
+static void cap_maxline(IRC_SERVER_REC *server)
+{
+	int maxline = 0;
+	if (parse_uint(g_hash_table_lookup(server->cap_supported, CAP_MAXLINE),
+		       NULL, 10, &maxline) &&
+	    maxline >= MAX_IRC_MESSAGE_LEN + 2 /* 2 bytes for CR+LF */) {
```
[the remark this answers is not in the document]
ailin-nemui: better?

ailin-nemui:
@dequis can this be merged?

ailin-nemui:
@irssi/developers

ailin-nemui:
nice!

PR #1094, opened 2019-08-13
show invites of other users when provided

ailin-nemui:


src/fe-common/irc/fe-events.c
```diff
@@ -252,13 +252,18 @@ static void event_pong(IRC_SERVER_REC *server, const char *data, const char *nic
...
-	char *params, *channel;
+	char *params, *invited, *channel;
 
 	g_return_if_fail(data != NULL);
 
-	params = event_get_params(data, 2, NULL, &channel);
-	signal_emit("message invite", 4,
-		    server, get_visible_target(server, channel), nick, addr);
+	params = event_get_params(data, 2, &invited, &channel);
+	if (server->nick_comp_func(invited, server->nick) == 0) {
```
dequis: Out-of-scope but this function seems very underused for own-nick comparisons in other parts of the codebase, those seem to prefer `g_ascii_strcasecmp` which is going to be accurate than this one.

src/fe-common/core/fe-messages.c
```diff
@@ -538,6 +538,20 @@ static void sig_message_invite(SERVER_REC *server, const char *channel,
...
 
+static void sig_message_invite_other(SERVER_REC *server, const char *channel,
+				     const char *invited, const char *nick, const char *address)
+{
+	char *str;
+	int level = MSGLEVEL_INVITES;
+
+	ignore_check_plus(server, nick, address, channel, invited, &level, FALSE);
+
+	str = show_lowascii(channel);
```
dequis: TIL about this function! Seems like the point is to prevent people from introducing color codes in channel names to trick others into joining channels they don't actually trust, which makes sense for incoming invites for you.

For invites to other people in channels you already are part of, it doesn't make as much sense.

But harmless, so I don't mind if you keep it this way.

PR #1095, opened 2019-08-13
fix cap queue order

ailin-nemui:
as capabilities are prepended to the queue, it needs to be reversed once before sending it out


ailin-nemui:
@irssi/developers

PR #1097, opened 2019-08-13
use extended-join capability

ailin-nemui:


src/core/nick-rec.h
```diff
@@ -8,6 +8,7 @@ time_t last_check; /* last time gone was checked */
 char *nick;
 char *host;
 char *realname;
+char *account;
```
dequis: At what point are we ABI bumping? It'd be annoying and conflicting to do it in the middle of all these PRs, but.

ailin-nemui: yes...

PR #1099, opened 2019-08-14
support cap away-notify

ailin-nemui:


src/fe-common/irc/fe-irc-messages.c
```diff
@@ -308,6 +342,7 @@ void fe_irc_messages_init(void)
         signal_add_last("message irc notice", (SIGNAL_FUNC) sig_message_irc_notice);
         signal_add_last("message irc own_ctcp", (SIGNAL_FUNC) sig_message_own_ctcp);
         signal_add_last("message irc ctcp", (SIGNAL_FUNC) sig_message_irc_ctcp);
+        signal_add_last("message irc away", (SIGNAL_FUNC) sig_message_irc_away);
```
dequis: Is this the right location for this sort of event? All the others are channel-targeted and this one broadcasts, more like joins or quits, which aren't in this file.

ailin-nemui: it's a bit of a question. I felt this event belongs more to IRC than the generic code (where the joins and quits are)

but I think it could be argued either way. I can move it if you think that's better (the message would lose the `irc' bit)

ailin-nemui:
I added some commits to move the code to fe-common/core, we can either use these changes or throw them away

src/fe-common/irc/fe-irc-messages.c
dequis: Yeah nevermind, didn't consider that all the other events are not irc specific, but this one certainly is. That grouping makes more sense than "channel-targeted"

(not that i'm a big fan of allowing irssi to support non-irc protocols, but, whatever, let's keep that design decision)

dequis: Wait nevermind you changed it since I originally wrote this comment. Oops. I.. also don't mind this other version. Like you said, can be argued either way.

dequis: This last version has it in the same file as the spread_server_message_to_windows function, which I'd say is the strongest argument so far.

PR #1100, opened 2019-08-15
refactor quit message into a separate function

ailin-nemui:


dequis:
Context: This refactor is currently a commit in a bunch of other PRs implementing ircv3 caps (#1096, #1098, #1099) and I asked to have this split to its own PR to simplify the review process

ailin-nemui:
I "moved the part handler up" so the patch is less confusing

PR #1102, opened 2019-08-16
fix wrong version in version-date when building git with merges of old commits on top

ailin-nemui:


ailin-nemui:
@irssi/developers

utils/irssi-version.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/sh
 
-DATE=`GIT_DIR=$1/.git git log -1 --pretty=format:%ai HEAD 2>/dev/null`
+DATE=`GIT_DIR=$1/.git git log -1 --pretty=format:%ci HEAD 2>/dev/null`
```
berkerpeksag: This LGTM but we can use the following commands to get date and time without doing any parsing on our end:

```sh
$ git log -1 --pretty=format:%cd --date=format:%Y-%m-%d
2019-07-17
$ git log -1 --pretty=format:%cd --date=format:%H:%M:%S
23:28:15
```

ailin-nemui: Nice. however the time mustn't start with 0 so I'm not sure if we can safely skip the awk

ailin-nemui: @dwfreed commented to just turn all those constants into strings

ailin-nemui:
I'm merging this now let's keep the other improvements in mind

PR #1108, opened 2019-08-19
implement server-time

ailin-nemui:


ailin-nemui:
this is currently layered on top of some other prerequisite PRs

ailin-nemui:
to try it, /set show_server_time on

ailin-nemui:
@irssi/developers

src/core/servers.c
```diff
@@ -660,6 +660,14 @@ void server_change_nick(SERVER_REC *server, const char *nick)
 	signal_emit("server nick changed", 1, server);
 }
 
+void server_stash_meta(SERVER_REC *server, const char *meta_key, const char *meta_value)
+{
+	server->current_incoming_meta =
+	    g_slist_prepend(server->current_incoming_meta, g_strdup(meta_key));
+	server->current_incoming_meta =
+	    g_slist_prepend(server->current_incoming_meta, g_strdup(meta_value));
```
dequis, line 668: Why not a hash table? It's not like linked lists are particularly efficient.

src/irc/core/irc.c
```diff
@@ -383,9 +383,45 @@ static void irc_server_event(IRC_SERVER_REC *server, const char *line,
...
+		return NULL;
+	}
+
+	if ((end = strchr(val, ';')) != NULL) {
+		return g_strndup(val, end - val);
+	} else {
+		return g_strdup(val);
+	}
+}
+
```
dequis, line 415: I'd much prefer something that goes tag by tag, and puts every key-value pair it finds into a hash table. searching for ";key=" sounds like it will fail in a bunch of edge cases, and as we add support for more tags, this will just awkwardly re-scan the tag string several times.

src/fe-text/textbuffer-formats.c
```diff
@@ -14,7 +14,23 @@
...
 
 TEXT_BUFFER_REC *color_buf;
-int scrollback_format;
+gboolean scrollback_format;
+gboolean show_server_time;
+
+#if GLIB_CHECK_VERSION(2, 56, 0)
+/* nothing */
+#else
+/* compatibility code for old GLib */
```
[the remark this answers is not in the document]
ailin-nemui: shall I move it to misc? or make a new compat file?

src/fe-text/textbuffer-formats.h
```diff
@@ -3,6 +3,11 @@
 
 #include <irssi/src/fe-text/textbuffer.h>
 
+typedef struct _TEXT_BUFFER_META_REC {
+	gint64 server_time;
+	GSList *next;
```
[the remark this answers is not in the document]
ailin-nemui: the meta will consist of pre-parsed meta-data like server_time (to be extended later, but keep in mind that this storage space is required for *each* line), and arbitrary generic meta-data in a list.

src/core/servers.c
ailin-nemui: they are space efficient and easy to push though. do you think a hash table has so much advantages that you want me to reimplement this as hash instead?

src/irc/core/irc.c
ailin-nemui: afaik the IRCv3 tag spec was written in such a way that there are no edge cases. the advantage of re-scanning is that we can just keep the current TAG-string parameter emitting code, otherwise we have to do something much more complicated.

src/core/servers.c
ailin-nemui: note that this stash in particular is only pushed to, then it is iterated over and turned into a TEXT_BUFFER_META_REC

src/fe-text/textbuffer-formats.c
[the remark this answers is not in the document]
ailin-nemui: if I move it to misc, do we need to rename it to i_date_time_new_from_iso8601  and let the existing impl be a @define for newer glib?

src/irc/core/irc.c
ailin-nemui: in any case, we (ideally) need a function that can be accessed from Perl that turns the existing tags string from irc_server_event_tags into the tag value

src/core/servers.c
dequis: >they are space efficient

I figured I'd test that claim, wrote [this test program][1] which puts a bunch of key/value pairs in gslist or ghashtable with random 8 char strings (so not testing for allocator efficiency here, which linked lists rely a lot more on)

Ran the thing with massif and uploaded [screenshots of massif-visualizer here][2]. You can view most of those side-by-side (except the 100 ones where the y axis is 90kb in one and 80kb in the other, annoying.)

In the graphs, the memory usage of linked lists increases linearly in a straight line, and hash tables are similarly linear but much more steppy, sometimes more memory efficient, sometimes less. The difference doesn't look significant. Area under curve looks more or less the same by eyeballing.

The peak memory usage:

|pairs|list|hash|
|------|----|----|
|10|22.2 KiB|22.3 KiB|
|100|26.5 KiB|26.0 KiB|
|1000|70.0 KiB|79.2 KiB|
|10000|506.8 KiB|514.6 KiB|

Can go either way. Mostly depends on which side of the curve you land on. Not really significant.

Also, I read this rant a while ago and it really stuck with me:

https://rust-unofficial.github.io/too-many-lists/#an-obligatory-public-service-announcement

It addreses a bunch of common points in favor of linked lists, but mostly from the perspective of using vectors instead (in our case, GArray). Except we're storing key-value data, and we already have a nice data structure for that.

>easy to push

That'd be `g_hash_table_insert(hash_table, meta_key, meta_value)` vs the four lines this comment thread highlighted. 


[1]: https://dump.dequis.org/indexed/massifs-hash-vs-list/asd.c
[2]: https://dump.dequis.org/indexed/massifs-hash-vs-list/graphs/

src/fe-text/textbuffer-formats.h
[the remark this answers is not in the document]
dequis: Does that mean that `next` will be always null for now?

src/fe-text/textbuffer-formats.c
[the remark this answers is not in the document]
dequis: YYyyeah i think that means we don't really have a better place, exposing this stuff through the irssi headers sounds like a bad time. Keeping it here sounds like the lesser evil.

src/irc/core/irc.c
dequis: >afaik the IRCv3 tag spec was written in such a way that there are no edge cases

Yeah, fair, not many. But value-less tags (`@key1=value1;key2;key3`) are definitely a thing.

src/fe-text/textbuffer-formats.h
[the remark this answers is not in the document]
ailin-nemui: yes

src/core/servers.c
ailin-nemui: interesting. you have to create the hash table first though

ailin-nemui:
I made a hash table implementation

ailin-nemui:
@dequis do you want to review this again? Now we have a tags parser and everything is hashes

ailin-nemui:
@irssi/developers

dequis:
Okay, okay, give me a few days.

ailin-nemui:
@dequis how many days do you want :)?

dequis:
Uhm. A few more?

ailin-nemui:
@dequis ping

dequis:
Damn time flies huh

src/core/refstrings.h
```diff
@@ -6,6 +6,8 @@
 #if GLIB_CHECK_VERSION(2, 58, 0)
 
 #define i_refstr_init() /* nothing */
+/* callback implementation */
+void i_refstr_release(char *str);
 #define i_refstr_release(str) ((str) == NULL ? NULL : g_ref_string_release(str))
```
dequis, line 11: This sounds potentially bad? Having the name defined as both a function and a macro and needing an undef in refstrings.c

dequis:
I like this a lot!

I tried really hard to find something to complain about to justify the time it took me to get around to this, and luckily I got something in the end. Phew.

But yeah all good from architectural point of view. I'm a bit rusty on detecting memory bugs visually, we should just fuzz this stuff.

src/core/refstrings.h
ailin-nemui: just make all of them functions then?

ailin-nemui:
thanks so much for your time and looking through it, much appreciated

src/core/refstrings.h
dequis: Yeah, good idea, sounds more consistent from api/abi perspective too

dequis (approved):
:+1: from the code review perspective, didn't do extra testing myself but we can do that after the merge

PR #1111, opened 2019-08-22
add various new message events to signals.txt

ailin-nemui:
these are from the merged ircv3 patches


ailin-nemui:
@irssi/developers

PR #1120, opened 2019-08-30
free params

ailin-nemui:


ailin-nemui:
let's see if oss-fuzz can find this missing free as well @josephbisch

PR #1122, opened 2019-09-04
add /identify command

isundil:
A a new command /identify [password]
The aim of this command is to avoid storing the Nickserv password in a config file and to avoid typing directly the command
It can be customized in chatnet using identifycmd (default is PRIVMS NickServ IDENTIFY %s)
The password is requested like /oper opername (similar to noecho but print spaces)


dequis:
Uhm, as a general guideline for any open source project, please discuss any non-trivial changes or additions in the issue tracker before spending any significant amount of time working on them.

isundil:
I understand your opinion and we can discuss about the enhancement here.
I wouldn't have done this if the enhancement was more time consuming.

ailin-nemui:
could it be more appropriate to simply use an /alias identify ^msg nickserv identify and combine this with a script such as http://anti.teamidiot.de/static/nei/*/Code/Irssi/conceal.pl to hide the password from sight ?

(most people would probably prefer to store their passwords and  networks such as freenode require identifying on connect for best user experience like permit to join channels and hide the user host)

isundil:
As must of the IRC servers have an identify service I find it relevant to have it built-in.
As I also use others IRC client I let nickserv handle my autojoin channels

dequis:
Well, from an ircv3 perspective, identifying through nickserv is a legacy method, and has several downsides compared to SASL (mainly race conditions). I wouldn't encourage its use.

isundil:
Well, I didn't known about the SASL authentication. Maybe it's better to make it an alias then

ailin-nemui:
thank you for taking your time to provide this PR. Unfortunately, at this time, we do not consider this a core feature that should be provided inside Irssi. If you would like to make this available as a script we can discuss a PR that enhances the scripting API. There are also some other ideas about passwords which you could work on if you want (please discuss further to avoid any disappointments)

PR #1123, opened 2019-09-04
enable self-message support

ailin-nemui:


ailin-nemui:
without this, znc doesn't display your other client PMs for instance

ailin-nemui:
(note that this spec has been removed from IRCv3 but they have no usable replacement)

ailin-nemui:
there is no way to know which messages are echoes and which are from other connected clients :-)

dequis:
Yeah, so if you disable local echo entirely and display every incoming message, you don't need to know that.

But it gets perceived as higher latency when sending messages, can be weird for people used to it being instant (or can be seen as a feature confirming messages got to the server!). It'd be nice as an off-by-default setting.

PR #1124, opened 2019-09-04
fix crash in join due to incorrect free

ailin-nemui:


ailin-nemui:
reported by @josephbisch

PR #1125, opened 2019-09-04
add new meson build option -Dfuzzer-link-language=cpp for oss-fuzz

ailin-nemui:


ailin-nemui:
oss-fuzz needs to be linked with clang++ according to https://google.github.io/oss-fuzz/getting-started/new-project-guide/, this patch allows us to do this with Meson >= 0.51, and the flag -Dfuzzer-link-language=cpp

dequis:
If it works it's fine by me I guess, just one question about the "UI" side of this - does it ever make sense to pass `-Dfuzzer-link-language=c`?

Ignoring that it's the default (because, I assume, the fuzzer is not being built most of the time), that configuration is always broken when building the fuzzer, isn't it?

...now I'm not sure how it ever worked in the first place.

ailin-nemui:
the oss-fuzz docs say we have to link with clang++

but in fact you can build the fuzzer locally, and the libfuzzer included in clang has no such requirement at all; it will build just fine with -Dfuzzer-link-language=c (the default)

PR #1127, opened 2019-09-15
Fix memory leak in cap code

josephbisch:
Malformed CAPs fail to free params before returning from event_cap.

This makes oss-fuzz sad.


ailin-nemui:
#1120

Issue #354, opened 2015-11-14
STATUSMSG interferes with hilights


ilbelkyr:
Some brief testing:

```
1114 104804 <+Nymeria> normal message test
1114 104838 <+Nymeria> hilight ilbelkyr normal message test
1114 104902 <+Nymeria:+#apastebin> statusmsg test
1114 104929 <+Nymeria:+#apastebin> should hilight ilbelkyr but doesn't! statusmsg test
```

The 2nd and 4th of these messages should be hilighted, but only the 2nd is; the others are only considered normal non-hilight activity. (According to `printlevels.pl`, all of these are `PUBLICS`; only the 2nd is also `HILIGHTS`.)

Relevant hilight entry: `1114 103840    2 ilbelkyr  -priority 50`

Possibly relevant `set hilight` output:

```
1114 103845 [lookandfeel]
1114 103845 activity_hilight_level = MSGS DCCMSGS
1114 103845 hilight_act_color = %M
1114 103845 hilight_color = %Y
1114 103845 hilight_level = PUBLIC DCCMSGS NOTICES MODES WALLOPS MSGS TOPICS
1114 103845 hilight_nick_matches = OFF
1114 103845 hilight_nick_matches_everywhere = OFF
```

Notices behave as expected:

```
1114 110858 |NOTICES| -Nymeria:#apastebin- notice
1114 110905 |NOTICES HILIGHTS| -Nymeria:#apastebin- notice ilbelkyr
1114 110925 |NOTICES| -Nymeria:+#apastebin- statusnotice
1114 110928 |NOTICES HILIGHTS| -Nymeria:+#apastebin- statusnotice ilbelkyr
```

vague666:
If you try
1114 103845 hilight_nick_matches = ON
1114 103845 hilight_nick_matches_everywhere = ON
does it work better? (remove the hilight for the test)

ilbelkyr:
Doesn't work for me with those, either. :(

dequis:
>10:16 < vague> dx, any idea about this, https://github.com/irssi/irssi/issues/354 ?
>10:19 < dx> all this stuff https://github.com/irssi/irssi/blob/ac04fa7/src/fe-common/core/fe-messages.c#L185-L215
>10:19 < dx> ...is missing from https://github.com/irssi/irssi/blob/ac04fa7/src/fe-common/irc/fe-irc-messages.c#L70-L96

ilbelkyr:
[...]

PR #1129, opened 2019-09-29
#354 Fix hilight behavior for STATUSMSG

dcollinsn:
This patch allows irc_op_public messages to properly trigger
hilights when the message mentions the current nick or one of our
hilights. This is done by copying the required code from
sig_message_public. This is important because Freenode has begun
using this message type for messages that can only be seen by ops
due to the +z channel mode, and ops will want to be notified of
watchwords even in that type of message.

To test, make two connections to Freenode, join a new channel. The
first client to join that channel will be an op. To establish a
baseline, use the non-opped client to attempt to "ping" the opped
client by addressing it by name and using terms in /hilight. Then,
set channel mode to +mz and use the non-opped client to send the
messages again. Without this patch, no message will "ping" the opped
client with +mz set. With this patch, "pings" should operate
normally, causing a bell, hilighting the window number, and so on.

What I don't know is whether there is any other code from
sig_message_public that should be copied over too. In particular,
the lines related to "ignore_check_plus", "emphasis", and
"printnick", I don't know if they are needed here. I also don't know
if there are any other message types that these changes should be
    applied to.

### DIFF
--- a/src/fe-common/irc/fe-irc-messages.c
+++ b/src/fe-common/irc/fe-irc-messages.c
@@ -67,13 +67,14 @@ static void sig_message_own_public(SERVER_REC *server, const char *msg,
 
 }
 
-/* received msg to all ops in channel */
+/* received msg to all ops in channel.
+   TODO: this code is a duplication of sig_message_public */
 static void sig_message_irc_op_public(SERVER_REC *server, const char *msg,
 				      const char *nick, const char *address,
 				      const char *target)
 {
 	CHANNEL_REC *chanrec;
-	char *nickmode, *optarget, *prefix, *color;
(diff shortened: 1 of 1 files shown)


ailin-nemui:
Thanks for your contribution! I suspect the ignore_check_plus does indeed need to go there as well, we have to check if it is possible to /ignore and /ignore HIDDEN/NO_ACT these messages properly

ailin-nemui:
@dcollinsn can you test this additional commit

ailin-nemui:
@irssi/developers

dcollinsn:
@ailin-nemui Yes, tested and works fine for me!

PR #1132, opened 2019-10-05
Remove an unused function and deprecate another

horgh:


ailin-nemui:
I can understand the duality of providing conversion functions. @irssi/developers  ?

PR #1134, opened 2019-10-12
Use correct TLS settings when reconnecting

### LINKED ISSUE #1027: TLS flag not reset during connection attempts
I have a TLS and a non-TLS server in the same rotation:

    eu.irc6.net          6667  IRCnet     tls
    irc.in.tum.de        6667  IRCnet     

When I use `/connect IRCnet` while not connected to a v6 capable net, I get a connection error trying to reach the v6 server, and then a TLS protocol error from the other server, even though that doesn't have a `tls` flag. If I interrupt with `/rmreconns` and give a new `/connect IRCnet` command, it works if the first server it tries is the non-TLS server, so it seems that the `tls` flag is stuck between connection attempts.

This is irssi 1.0.7 as found in Debian stable and 1.2.0 as found in Debian stable-backports.

horgh:
This is to fix #1027.

### DIFF
--- a/src/core/servers-reconnect.c
+++ b/src/core/servers-reconnect.c
@@ -151,10 +151,6 @@ static int server_reconnect_timeout(void)
 
 static void sserver_connect(SERVER_SETUP_REC *rec, SERVER_CONNECT_REC *conn)
 {
-        conn->family = rec->family;
-	conn->address = g_strdup(rec->address);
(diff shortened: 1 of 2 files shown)


ailin-nemui:
Nice!

ailin-nemui:
@irssi/developers

PR #1135, opened 2019-10-13
Remove resolve_reverse_lookup setting

horgh:
This basically just reverts 5b60b859d45feaa5504a196d77a5a4a318130493 which added it.

### DIFF
--- a/src/core/net-nonblock.c
+++ b/src/core/net-nonblock.c
@@ -27,13 +27,11 @@
 
 /* nonblocking gethostbyname(), ip (IPADDR) + error (int, 0 = not error) is
    written to pipe when found PID of the resolver child is returned */
-int net_gethostbyname_nonblock(const char *addr, GIOChannel *pipe,
-			       int reverse_lookup)
+int net_gethostbyname_nonblock(const char *addr, GIOChannel *pipe)
 {
 	RESOLVED_IP_REC rec;
 	const char *errorstr;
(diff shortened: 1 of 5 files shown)


ailin-nemui:
@irssi/developers

src/core/net-nonblock.c
```diff
@@ -27,13 +27,11 @@
 
 /* nonblocking gethostbyname(), ip (IPADDR) + error (int, 0 = not error) is
    written to pipe when found PID of the resolver child is returned */
-int net_gethostbyname_nonblock(const char *addr, GIOChannel *pipe,
-			       int reverse_lookup)
+int net_gethostbyname_nonblock(const char *addr, GIOChannel *pipe)
```
dequis: This looks API breaking, keep the param

ailin-nemui: maybe better to break the API since the feature is gone?

dequis: I don't think this feature is as important as the ability to do DNS resolutions at all. Callers won't really care if reverse lookup is done or not.

horgh: Good catch, thank you! It's there again.

dequis:
Thanks! Looks good. 

ABI version bump maybe?

dequis:
(RESOLVED_IP_REC becoming shorter)

PR #1137, opened 2019-10-17
handle levels more logically

vague666:
I've only tested this for hidelevel but I reckon it should work the same for everything else
IMO `/window hidelevel JOINS PARTS` should set the absolute value of hidelevel
and `+JOINS +PARTS` and `-JOINS -PARTS` should be used to add/delete levels

### DIFF
--- a/src/core/levels.c
+++ b/src/core/levels.c
@@ -173,19 +173,24 @@ int combine_level(int dest, const char *src)
 
 	g_return_val_if_fail(src != NULL, dest);
 
-	list = g_strsplit(src, " ", -1);
-	for (item = list; *item != NULL; item++) {
-		itemname = *item + (**item == '+' || **item == '-' ? 1 : 0);
-		itemlevel = level_get(itemname);
-
-		if (g_strcmp0(itemname, "NONE") == 0)
-                        dest = 0;
-		else if (**item == '-')
-			dest &= ~(itemlevel);
-		else
-			dest |= itemlevel;
+	if (index(src, '+') == NULL && index(src, '-') == NULL) {
+		dest = level2bits(src, NULL);
 	}
-	g_strfreev(list);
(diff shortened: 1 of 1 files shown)


ailin-nemui:
while I don't necessarily disagree in principle, this would be a huge change in behaviour (currently JOINS and +JOINS has the same meaning)

that's why I'm hesitant about making such a change

vague666:
I agree it does change things around but is it such a big change really? Usage wise
combine_level isn't used a lot in core, no idea about modules out there in the world
```
vague@irssi:~/irssi$ grep -RI combine_level src/*
src/core/levels.c:int combine_level(int dest, const char *src)
src/core/levels.h:int combine_level(int dest, const char *src);
src/fe-common/core/fe-ignore.c: rec->level = combine_level(rec->level, levels);
src/fe-common/core/window-commands.c:   window_set_level(active_win, combine_level(active_win->level, data));
src/fe-text/textbuffer-commands.c:      combine_level(gui->view->hidden_level, data));
src/perl/common/Core.xs:combine_level(level, str)
```
and only scripts in the repo using it are
```
sbclearmatch.pl
sb_search.pl
```

vague666:
Or maybe we should rewrite https://github.com/irssi/irssi/blob/master/src/fe-text/textbuffer-commands.c#L96 somehow

ailin-nemui:
fwiw you can already set absolute levels using `/window hidelevel NONE joins parts quits`

vague666:
It's not very logical

ailin-nemui:
I can't think of a good solution ...... I don't want to break backward compat

ailin-nemui:
@irssi/developers any other opinions?

ailin-nemui:
let's chat if you can think of any ideas and see if we can agree

PR #1138, opened 2019-10-19
Improve clarity of nick in use message

mquin:
If irssi's preferred nick is in use irssi will issue a whois command and report some information
on the current user of the nick. As the "is owned by" wording can be confusing to users of networks with nickname registration, I propose rephrasing this to "is in use by".


ailin-nemui:
@irssi/developers

PR #1141, opened 2019-10-31
Fix some glib deprecation warnings

dequis:
Most of these have been deprecated since forever (2.2), but they didn't
raise warnings. Now they do, and the warnings are not the most verbose
warnings you could ask for, but, they point in the right direction.

This doesn't handle the GTimeVal deprecation warnings. Those seem
trickier since they cover API, will look into those right after this.


ailin-nemui:
Nice!

PR #1146, opened 2019-12-10
Reconnect retry IPv4 on IPv6 failure 

### LINKED ISSUE #641: Does not fall back to to IPv4 address if IPv6 address is unreachable
This is [FS#579](https://github.com/irssi-import/bugs.irssi.org/issues/579) carried over from #26:

    resolve_prefer_ipv6 = ON
    Irssi: Looking up irc.oftc.net
    Irssi: Connecting to irc.oftc.net [2605:bc80:3010:101:0:2:0:6667] port 6697
    Irssi: Unable to connect server irc.oftc.net port 6697 [Network is unreachable]
    resolve_prefer_ipv6 = OFF
    Irssi: Looking up irc.oftc.net
    Irssi: Connecting to irc.oftc.net [2600:3c03::f03c:91ff:fe96:35cd] port 6697
    Irssi: Unable to connect server irc.oftc.net port 6697 [Network is unreachable]

The `resolve_prefer_ipv6` setting — also subject of #190 — is not really relevant here. The problem is that for chat servers that advertise `AAAA` records in DNS, unless Irssi can establish an IPv6 connection to the server, the connection fails. The only way to get it to work is to use `/connect -4`.

The correct behaviour with `getaddrinfo()` (which respects the lookup order defined in `/etc/gai.conf`) would be for Irssi to fall back to using the IPv4 address, if such is advertised (which it is in this case).

slingamn:
Track the address family of the last failed connection attempt (either immediately or during TLS handshake), then disprefer that address family during reconnection.

A note on testing ipv6 connectivity failure on Linux: `disable_ipv6` will not simulate it correctly, because irssi calls `getaddrinfo(3)` with `AI_ADDRCONFIG`, which suppresses resolution of ipv6 addresses if no local ipv6 address is available.

The correct way to simulate this is with `disable_ra`. That way, the link-local address will remain, and `getaddrinfo(3)` will return addresses corresponding to AAAA records, but attempts to connect to them will fail with "Network is unreachable".

I tested a few different cases (disable_ra for cutting off ipv6, an iptables rule dropping outgoing packets for cutting off ipv4), but this probably needs more testing --- let me know if you have suggestions.

### DIFF
--- a/src/core/server-connect-rec.h
+++ b/src/core/server-connect-rec.h
@@ -11,6 +11,7 @@ int proxy_port;
 char *proxy_string, *proxy_string_after, *proxy_password;
 
 unsigned short family; /* 0 = don't care, AF_INET or AF_INET6 */
+unsigned short chosen_family; /* family actually chosen during name resolution */
 char *tag; /* try to keep this tag when connected to server */
 char *address;
 int port;
@@ -43,5 +44,6 @@ unsigned int unix_socket:1; /* Connect using named unix socket */
 unsigned int use_tls:1; /* this connection uses TLS */
 unsigned int tls_verify:1;
 unsigned int no_connect:1; /* don't connect() at all, it's done by plugin */
+unsigned short last_failed_family; /* #641: if we failed to connect to ipv6, try ipv4 and vice versa */
 char *channels;
 char *away_reason;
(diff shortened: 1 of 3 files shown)


slingamn:
This is really a band-aid because it imposes the standard reconnection timeout (default 5 minutes?) on any attempt to fail over from ipv6 to ipv4 or vice versa. Ideally in the future we'd do one of the following:

1. A suggestion from nei in #irssi: resolve all the addresses, then do local round-robin over them (sorting them in the order implied by `resolve_prefer_ipv6`) until a connection succeeds.
2. A full Happy Eyeballs implementation

ailin-nemui:
@irssi/developers

ailin-nemui:
nb this will require an abi bump due to added field

ailin-nemui:
@irssi/developers

PR #1151, opened 2020-01-15
Add a "server outgoing" signal mirroring "server incoming"

### LINKED ISSUE #1148: Add a "server outgoing" signal mirroring "server incoming"
For all incoming messages, irssi first emits a "server incoming" signal. This allows for very low-level access to the incoming IRC traffic. I was looking for an equivalent signal for full control over outgoing messages and came up empty; as far as I can tell, one would have to individually bind to each relevant signal that could trigger a message to be sent. I'd like to propose adding a "server outgoing" signal. It would be triggered once for each line just before it is actually sent to the server and have two arguments, the `SERVER_REC` and the line to be sent. Stopping the signal would also stop the line from being sent.

JustAnotherArchivist:
Fixes #1148

### DIFF
--- a/docs/signals.txt
+++ b/docs/signals.txt
@@ -128,6 +128,7 @@ irc-nicklist.c:
  "server nick changed", SERVER_REC
 
 irc-servers.c:
+ "server outgoing", SERVER_REC, char *str
  "event connected", SERVER_REC
 
 irc-cap.c
--- a/src/irc/core/irc-servers.c
+++ b/src/irc/core/irc-servers.c
@@ -603,7 +603,7 @@ static int server_cmd_timeout(IRC_SERVER_REC *server, gint64 now)
 
 	/* send command */
 	len = strlen(cmd);
-	irc_server_send_data(server, cmd, len);
+	signal_emit("server outgoing", 3, server, cmd, len);
 
 	/* add to rawlog without [CR+]LF */
         if (len > 2 && cmd[len-2] == '\r')
@@ -1060,6 +1060,7 @@ void irc_servers_init(void)
 	cmd_tag = -1;
 
 	signal_add_first("server connected", (SIGNAL_FUNC) sig_connected);
+	signal_add("server outgoing", (SIGNAL_FUNC) irc_server_send_data);
 	signal_add_last("server destroyed", (SIGNAL_FUNC) sig_destroyed);
 	signal_add_last("server quit", (SIGNAL_FUNC) sig_server_quit);
 	signal_add("server cap ack " CAP_MAXLINE, (SIGNAL_FUNC) cap_maxline);
@@ -1088,6 +1089,7 @@ void irc_servers_deinit(void)
 		g_source_remove(cmd_tag);
 
 	signal_remove("server connected", (SIGNAL_FUNC) sig_connected);
+	signal_remove("server outgoing", (SIGNAL_FUNC) irc_server_send_data);
 	signal_remove("server destroyed", (SIGNAL_FUNC) sig_destroyed);
         signal_remove("server quit", (SIGNAL_FUNC) sig_server_quit);
 	signal_remove("server cap ack " CAP_MAXLINE, (SIGNAL_FUNC) cap_maxline);
--- a/src/irc/core/irc-servers.h
+++ b/src/irc/core/irc-servers.h
@@ -160,7 +160,6 @@ void irc_server_send_action(IRC_SERVER_REC *server, const char *target,
 char **irc_server_split_action(IRC_SERVER_REC *server, const char *target,
 			       const char *data);
 void irc_server_send_away(IRC_SERVER_REC *server, const char *reason);
-void irc_server_send_data(IRC_SERVER_REC *server, const char *data, int len);
 void irc_server_init_isupport(IRC_SERVER_REC *server);
 
 void irc_servers_start_cmd_timeout(void);
--- a/src/irc/core/irc.c
+++ b/src/irc/core/irc.c
@@ -117,7 +117,7 @@ void irc_send_cmd_full(IRC_SERVER_REC *server, const char *cmd,
 	}
 
 	if (send_now) {
-                irc_server_send_data(server, str->str, str->len);
+                signal_emit("server outgoing", 3, server, str->str, str->len);
 		g_string_free(str, TRUE);
 	} else {
 


ailin-nemui:
@irssi/developers any opinions here?

dequis:
Not really, I guess it's alright?

ailin-nemui:
@JustAnotherArchivist how do you feel about this instead

```
diff --git a/src/irc/core/irc-servers.c b/src/irc/core/irc-servers.c
index 9779ab58..67ce601a 100644
--- a/src/irc/core/irc-servers.c
+++ b/src/irc/core/irc-servers.c
@@ -553,6 +553,8 @@ void irc_server_send_away(IRC_SERVER_REC *server, const char *reason)
 
 void irc_server_send_data(IRC_SERVER_REC *server, const char *data, int len)
 {
+	signal_emit("server outgoing", 3, server, data, len);
+
 	if (net_sendbuffer_send(server->handle, data, len) == -1) {
 		/* something bad happened */
 		server->connection_lost = TRUE;
```

dequis:
I think you want to be able to cancel the signal or modify the args, and your version doesn't allow that, while the original does

ailin-nemui:
If author requires the ability to modify the data before it goes into socket write. (then I would prefer to rename the signal to server send data or something}

dequis:
Doesn't "server incoming" work the same way? You can modify what happened on the wire before it continues normal processing by irssi. This is just mirroring that.

docs/signals.txt
```diff
@@ -128,6 +128,7 @@ irc-nicklist.c:
  "server nick changed", SERVER_REC
 
 irc-servers.c:
+ "server outgoing", SERVER_REC, char *str
```
dwfreed: This needs a length parameter:

```suggestion
 "server outgoing", SERVER_REC, char *str, int len
```

ailin-nemui: ... eww ...

true, but we should avoid char*+len style signal apis, easy to break memory

dwfreed: Then we'd need to change the C/Perl boundary to handle a new type that'll do the length param for us.  Perl does have an SvPV(SV *, STRLEN) macro that'll put the length of the string into the second parameter (STRLEN is a macro that usually expands to size_t).

dequis: Why does this one need it and "server incoming" doesn't? I don't see any other signal passing explicit length arguments after strings

dwfreed: Because server incoming is reading from the socket, and that what was read is NULL-terminated by the function doing the reading, so length doesn't matter.  Writing to a socket requires the length to write, which is taken very early and passed around.  Another option could be to change core to do an strlen at the last possible moment, since protocols irssi is likely to ever handle will never allow NULLs, and then this function and corresponding signal don't need a length parameter.

dequis: Of the two uses of this signal, both present in this PR, one does strlen() explicitly to acquire the length right before calling it, the other takes `gstring->str` which is [guaranteed to be null-terminated][1].

I don't think this function is special, no other code is really prepared to handle strings that aren't null terminated. We should just require the argument to be null terminated for users of the signal, which in practice is already a requirement for every char *

[1]: https://developer.gnome.org/glib/stable/glib-Strings.html

dequis: ..now I realize that you're right in that it needs a length parameter here in signals.txt, because the way it's used in the other parts of the diff passes that length parameter, only so that it can match the existing signature of `irc_server_send_data`. It's all about skipping one strlen for the gstring case, welp.

src/irc/core/irc-servers.h
```diff
@@ -160,7 +160,6 @@ void irc_server_send_action(IRC_SERVER_REC *server, const char *target,
 char **irc_server_split_action(IRC_SERVER_REC *server, const char *target,
 			       const char *data);
 void irc_server_send_away(IRC_SERVER_REC *server, const char *reason);
-void irc_server_send_data(IRC_SERVER_REC *server, const char *data, int len);
```
dequis: This sounds like removing API

JustAnotherArchivist:
Yes, the intention is to be able to block or modify the message before sending for full flexibility. Is there a good reason not to allow that, given that "server incoming" is implemented that way?

src/irc/core/irc-servers.h
JustAnotherArchivist: I removed that because it was listed under "INTERNAL" (not visible in the diff as it's a couple lines further up). My understanding was that it was listed there only because it had to be called from `src/irc/core/irc.c`.

docs/signals.txt
ailin-nemui: then get rid of the len and good to merge ?

JustAnotherArchivist: Hmm, good point. I didn't think much about this and simply did it the same way as "server incoming" is defined.

@dequis While `gstring->str` is guaranteed to be null-terminated, it can still contain nuls when using `g_string_append_len`.

As I understand it (see the other comment), `irc_server_send_data` is internal API, so changing its signature should not be an issue. For IRC specifically, we could probably get away with dropping the length parameter and only passing the `*char` since IRC does not permit NULs anywhere. So I guess the question is, should "server incoming/outgoing" be generic enough to handle any protocol or IRC-specific (as incoming is currently and outgoing is implemented here)?

JustAnotherArchivist: So, what's the solution here? Remove the length parameter and run `strlen` inside `irc_server_send_data`?

ailin-nemui: @dwfreed any thoughts?

src/irc/core/irc-servers.c
```diff
@@ -603,7 +603,7 @@ static int server_cmd_timeout(IRC_SERVER_REC *server, gint64 now)
 
 	/* send command */
 	len = strlen(cmd);
-	irc_server_send_data(server, cmd, len);
+	signal_emit("server outgoing", 3, server, cmd, len);
 
 	/* add to rawlog without [CR+]LF */
```
ailin-nemui, line 608: just looking at this patch after @JustAnotherArchivist pinged me on IRC, I notice that the wrong command will be logged to rawlog as far as I can tell

JustAnotherArchivist: Oh, yes. Good catch, thank you! There could be signal handlers both before and after `irc_server_send_data` that might modify `cmd`, so this issue exists on both the direct sends (where the rawlog is currently written before the signal) and the ones from the queue (where it's after). I think I'll just move the`rawlog_output` call to `irc_server_send_data`.

Should a command be in the rawlog if `net_sendbuffer_send` fails?

JustAnotherArchivist: While finally looking into this, I realised that it's even trickier. There's also the `server_redirect_command` business. I don't fully understand what it does yet, but it complicates things. It isn't possible to simply use `server->cmdqueue->next->data` in `irc_server_send_data` because signal handlers might change the queue inbetween (e.g. with calls to `Server::send_raw_now` in Perl), so the beginning of the queue might no longer correspond to the message being sent. The easiest way around it would be to add the redirect argument to the signal, but as I don't understand what the redirects do in the first place, I'm not sure if that would actually be a good idea.

ailin-nemui: we had a brief talk about it on IRC and in summary we found that the current Irssi architecture does not seem to allow us to pass some variables along a signal emission unless they are part of the signal itself (or as peksy C globals). And we still don't know how to solve this issue here

ailin-nemui:
a different variant of this has been merged into ailin-nemui/irssi#43 (using GString)

PR #1155, opened 2020-02-09
rewrite how /bind searches for bindings when listing them

vague666:


ailin-nemui:
what use case is this?

vague666:
When searching it searches from the beginning, so you need to know what you are looking for to find it, would be better to search the same way /set txt finds settings with txt in them

ailin-nemui:
mhm, I guess that's fine. can you change it to strstr instead of g_strrstr though ;p

@irssi/developers  ?

ailin-nemui:
@irssi/developers

PR #1158, opened 2020-03-09
-time was missing in /ignore output and made it human-readable

vague666:


ailin-nemui:
@irssi/developers ?

dequis:
Weird commit subject phrasing but ok. It sounded like "-time was missing in the output, and that was the reason it was readable. Now it's not. Get rid of it again"

PR #1159, opened 2020-03-12
Make sure witem exists in cmd_cat when using -window

vague666:


ailin-nemui:
@vague666 can you test my change?

vague666:
Works for me, although I'd prefer to print to an itemless window if that's possible, like if item == NULL then Irssi::active_win <-- whatever c code that calls to find a target

ailin-nemui:
then you can maybe change the code to use 
```
if (-window flag is given) {
    printtext_window(active_win, ...);
} else {
    printtext(NULL, NULL, ...);
}
```
?

ailin-nemui:
@irssi/developers

PR #1161, opened 2020-03-18
unignore_time can not be NULL, check for 0 instead

vague666:


dequis:
Damn we're good at code review

PR #1165, opened 2020-03-31
fix warning in cmd_ping

ailin-nemui:


dequis:
Oh is this the sort of thing where 64bit platforms are happy but 32bit complain? 

I think G_GINT64_FORMAT exists for portability here

```
"%s PING %" G_GINT64_FORMAT " %" G_GINT64_FORMAT
```

Looks like ass but yeah.

ailin-nemui:
or maybe we need to cast something to (long). and there is another warning caused by a corresponding sscanf. and scanning to long long is not portable. advice of real C programmers (i.e. @dequis ) appreciated

dequis:
Oh that sscanf is annoying. So the old version pre-GTimeVal-deprecation-fixes was

```
if (sscanf(data, "%ld %ld", &tv2.tv_sec, &tv2.tv_usec) < 1) {
```

where tv2 is a GTimeVal

```
struct _GTimeVal
{
  glong tv_sec;
  glong tv_usec;
} GLIB_DEPRECATED_TYPE_IN_2_62_FOR(GDateTime);
```

So one way is to have the sscanf put stuff into `long`-sized temp vars instead of operating directly on the gint64... but i guess we're being pushed to do this for y2038 reasons.

Can we uhhhhh `event_get_params(data, 2, &sec, &usec)` and `g_ascii_strtoull()` each?

ailin-nemui:
@dequis do you think this would work?

dequis:
Before 2038, i guess, yeah

dequis:
wait what

ailin-nemui:
this still needs to be fixed... but it should be y2k38 safe

PR #1166, opened 2020-04-02
fix perl module build on openbsd

ailin-nemui:
unfortunately, some mangling is needed to create the correct linker and compiler invocations

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -406,6 +406,12 @@ if want_capsicum
   endif
 endif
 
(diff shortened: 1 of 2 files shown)


ailin-nemui:
this fixes perl errors because symbols in libm cannot be found and creates the "same" module as libtool

ailin-nemui:
originally reported by letty

ailin-nemui:
@irssi/developers

PR #1167, opened 2020-04-02
fix crash on /quit when unloading modules

ailin-nemui:
this fixes a crash on /quit when the module unloaded is trying to reference symbols from already-unloaded modules, by reversing the lists.

### DIFF
--- a/src/core/modules.c
+++ b/src/core/modules.c
@@ -214,7 +214,7 @@ MODULE_FILE_REC *module_register_full(const char *name, const char *submodule,
 		module = g_new0(MODULE_REC, 1);
 		module->name = g_strdup(name);
 
-                modules = g_slist_append(modules, module);
+		modules = g_slist_prepend(modules, module);
 	}
 
 	file = module_file_find(module, submodule);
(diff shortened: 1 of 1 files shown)


ailin-nemui:
originally reported by letty

ailin-nemui:
as a side effect, this will reverse the output of /load. not sure if that's anything to worry about

ailin-nemui:
@irssi/developers

dequis:
Can you post some specifics? Seems like a rather arbitrary fix without context.

ailin-nemui:
if we /load perl, then Irssi will append "core", then "fe", to module->files
https://github.com/irssi/irssi/blob/00a7efc69f8cfc14fcfdf123421015e99ac07926/src/core/modules.c#L229

on quit, Irssi will unload all modules by iterating over module->files:
https://github.com/irssi/irssi/blob/8b7bf1f04eefea51fdeb9d151ef0be2c5956c0ea/src/core/modules-load.c#L439

when the fe-perl is unloaded, it will call perl_core_print_script_error:
https://github.com/irssi/irssi/blob/00a7efc69f8cfc14fcfdf123421015e99ac07926/src/perl/perl-fe.c#L292

at that time, due to the order of the list (first "core" is unloaded, then "fe"), this function already ceased to exist:
https://github.com/irssi/irssi/blob/00a7efc69f8cfc14fcfdf123421015e99ac07926/src/perl/perl-core.c#L385

that leads to a segfault on openbsd

ailin-nemui:
by reverting the list, a module that loads "core" and "fe" will put "fe", then "core" in the list. now, the unloading will be "fe", then "core", thus avoiding the crash

dequis:
Thanks!

Issue #828, opened 2018-01-23
irssi fails to close when launched embedded in a terminal


ghost:
`uxterm -e irssi`

![output2](https://user-images.githubusercontent.com/11262717/35299003-3ae59766-0084-11e8-9181-837734e15987.gif)

dequis:
So this is urxvt is sending SIGHUP (and nothing else) and irssi reloads config on SIGHUP. I don't know what to say here. Maybe urxvt should be trying a bit harder?

hhiru:
I'm sorry, I misspelled. I meant uxterm

dequis:
Maybe uxterm should be trying a bit harder?

ailin-nemui:
works fine in my uxterm, I just tried. maybe it's urxvt after all? otherwise it seems to be specific to your system and not related to irssi. HUP to irssi -> reload config, that works as intended. but your "x" should not send HUP (probably)

ailin-nemui:
`uxterm -e irssi` indeed "x" sends sighup. however, from irssi perspective it is working as intended, hup -> reload config.

ailin-nemui:
maybe we should change that or offer a setting. can't seem to reliably rewire signals from scripts

ailin-nemui:
proposed solution: /set quit_on_hup ON/OFF

Pinguin1234:
I am working on it in PR #1169 .

Pinguin1234:
I think this could be closed now.

ailin-nemui:
thanks @Pinguin1234

PR #1169, opened 2020-04-03
Added option to quit in SIGHUP

Pinguin1234:
Fixes Issue #828

### DIFF
--- a/src/core/core.c
+++ b/src/core/core.c
@@ -88,6 +88,11 @@ static void sig_reload_config(int signo)
         reload_config = TRUE;
 }
 
+static void sig_quit_on_hup(int signo)
+{
+	signal_emit("gui exit", 0);	
+}
+
 static void read_settings(void)
 {
 	static int signals[] = {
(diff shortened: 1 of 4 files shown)


ailin-nemui:
nice!

ailin-nemui:
@irssi/developers

src/core/core.c
```diff
@@ -88,6 +88,11 @@ static void sig_reload_config(int signo)
         reload_config = TRUE;
 }
 
+static void sig_quit_on_hup(int signo)
+{
+	signal_emit("gui exit", 0);
```
dequis, line 93: This isn't [async-signal safe][1]. The code can safely inside signal handlers is very limited, and there's stuff inside signal_emit and the gui exit handlers that interacts with the memory allocator, which holds a lock. So if a SIGHUP arrives at the precise moment some other part of the code is doing memory allocation stuff anywhere else, that'll deadlock irssi.

You have to do something similar to the handler above, which just changes the value of the global `reload_config` to TRUE, and nothing else, just letting some code outside the signal handler do the actual config reload.

Things you could do here:

1. Use the `quitting` global variable, defined at `src/fe-text/irssi.c`. This will quit cleanly, but it means other signal handlers of `gui exit` won't run, like autosave.
2. Drop this separate handler, and repurpose/rename the `reload_config` flag to a generic `sighup_received`, and adapt the current `reload_config` handler implemented in `src/fe-text/irssi.c` so that it either reloads the config or quits depending on what the setting says.

[1]: https://wiki.sei.cmu.edu/confluence/display/c/SIG30-C.+Call+only+asynchronous-safe+functions+within+signal+handlers

Pinguin1234: Thanks for the info! I actually didn't know that, as I never used signals before.

ailin-nemui:
you think this is good now @dequis ?

dequis:
Yeah, definitely okay from the signal side. Is `sighup_received` an ABI change? At least I can't imagine that change breaking anything, it's internal enough.

ailin-nemui:
@Pinguin1234 please indent your code with TAB

and increase the ABI_VERSION in common.h

Pinguin1234:
Should be done now.

ailin-nemui <notifications@github.com> schrieb am Mo., 6. Apr. 2020, 16:03:

PR #1172, opened 2020-04-12
run clang-format as a pull request hook

ailin-nemui:


dequis:
Sup?

ailin-nemui:
not sure how to best go about it. you can check the current state here :  https://github.com/irssi/irssi/pull/1079#partial-pull-merging

what do you think, is failing the  build like that good (enough)?

dequis:
Oh that makes more sense. Yeah, I'd just tweak the output so that the "cat diff" and "cmp" steps are the same (`&&` them together) so if it fails it shows the diff right in the failing step.

Also goddamn the scope of that PR was already scary before

ailin-nemui:
@dequis I think it's ready

ailin-nemui:
let's see how much self inflicted pain that will be!

PR #1179, opened 2020-04-21
Check for abi compatibility using github actions

ailin-nemui:


ailin-nemui:
wip

dequis:
Noice

ailin-nemui:
this will use abigail's abidiff so it's probably not bullet proof all but at least should stop us from big mistakes

ailin-nemui:
seems to be doing it's thing

https://github.com/irssi/irssi/actions/runs/84183959

@irssi/developers

PR #1183, opened 2020-04-28
manually handle NUL unicode in g_utf8_get_next_char_validated

### LINKED ISSUE #1180: ctrl+space freezes irssi on GLib >2.62

ailin-nemui:
A change in GLib 2.63 broke some assumptions in Irssi that the null-byte NUL  / U+0000 is a valid Unicode character. This would occur when the user types Ctrl+Space. As a result, the input loop never manages to process the NUL-byte (and any other user input that follows, ever).

This patch adds a manual check that properly advances the input loop if GLib returns -2 (incomplete character) despite the length being positive and a NUL is in first position.

Fixes #1180 
https://gitlab.gnome.org/GNOME/glib/-/merge_requests/967
https://gitlab.gnome.org/GNOME/glib/-/issues/2093

### DIFF
--- a/src/fe-text/term-terminfo.c
+++ b/src/fe-text/term-terminfo.c
@@ -672,7 +672,11 @@ void term_stop(void)
 
 static int input_utf8(const unsigned char *buffer, int size, unichar *result)
 {
-	unichar c = g_utf8_get_char_validated((char *)buffer, size);
+	unichar c = g_utf8_get_char_validated((char *) buffer, size);
+
+	/* GLib >= 2.63 do not accept Unicode NUL anymore */
+	if (c == (unichar) -2 && *buffer == 0 && size > 0)
+		c = 0;
 
 	switch (c) {
 	case (unichar)-1:


dequis:
Much neater message, thanks!

Sorry for being annoying about this but: Can you also `--amend` that message to the commit itself? I suspect we might end up seeing that commit turned as a patch file as part of distro packages for a while.

ailin-nemui:
updated

eqhmcow:
sorry to bother; but any reason this wasn't merged? or was is merged in a different commit to master?

Thank you

LaserEyess:
What's the status on this? Some IMEs use ctrl+space to switch between layouts, which crashes irssi every time.

felixfontein:
@ailin-nemui is there a chance that this could get merged (and a new irssi version released soon)? This bug is really annoying, since I have to kill irssi every time it happens :)

ailin-nemui:
hi @felixfontein even if there were an Irssi release there is nothing guaranteeing you that distributions would ship the update

ailin-nemui:
@irssi/developers

felixfontein:
@ailin-nemui true, but it definitely increases the chances! Thanks a lot :)

LaserEyess:
Just want to say I have been using a patched irssi for over a week now and this does fix the issue, in case you were waiting on user testing. I also disagree that there is 'no guarantee' that distros will ship this update. For distros affected by newer glibs and irssi, the most likely will if this fix is approved, either through backporting the fix or by updating irssi in their repos.

ailin-nemui:
@LaserEyess they can already do that now, though. (as some have)

felixfontein:
@ailin-nemui thank you very much for fixing this!

PR #1188, opened 2020-05-06
fix npe on no text from format_get_text_theme_charargs

ailin-nemui:


ailin-nemui:
for example during /quit if buffer has already been unloaded

ailin-nemui:
@irssi/developers

dequis:
Is this because of https://github.com/irssi/irssi/pull/1079 ?

ailin-nemui:
yea it's a condition in this new code

PR #1191, opened 2020-05-08
fix NULL assertion in format args

ailin-nemui:


ailin-nemui:
I'll merge this as it is another fixup to #1188 and the new code from #1079

dequis:
Yeah sure

Issue #1189, opened 2020-05-08
new scrollback_format code breaks "print text" modifiers


ailin-nemui:
for example colorize_nicks.pl stops working :-(

ailin-nemui:
reported by @g4570n

ailin-nemui:
List of scripts that are on scripts.irssi.org which are (obviously) broken by this new feature

- [ ] nickcolor_gay.pl
- [ ] dim_nicks.pl
- [ ] hideshow.pl
      changes the text_dest_rec with format_create_dest :-/
- [ ] bitlbee_hide_password.pl
- [ ] mass_hilight_blocker.pl
      changes the text_dest_rec level
- [ ] munge_own_nickname_to_username.pl
- [ ] colorize_nicks.pl

ailin-nemui:
looks like mass_hilight_blocker.pl is broken anyway regardless of this code

PR #1192, opened 2020-05-12
add "print format" signal to perl

ailin-nemui:


ailin-nemui:
this will be needed to implement a workaround / new way to change printed text from scripts. scripts have to be adapted to change the text in the format arguments, instead of the final "print text".

#1189 -> #1079

ailin-nemui:
@irssi/developers 

need to also think about adapting python &| tcl...

PR #1194, opened 2020-05-14
add Irssi::UI::TextDest->set_level($bits) to change the level  of a "print (no)format" or "print text" in the old code

ailin-nemui:


ailin-nemui:
this can be used for hideshow.pl in #1189

ailin-nemui:
@irssi/developers

Issue #1193, opened 2020-05-13
Support numeric 489 as ERR_SECUREONLYCHAN


mh-source:
This was already discussed shortly on IRC with @ailin-nemui - and per his request i open this issue on it.

Numeric 489 is currently defined only as ERR_VOICENEEDED in Irssi but is used as an error reply (ERR_SECUREONLYCHAN) for non-secure clients trying to join a secure-only channel on several networks.

Right now 489 will be printed in the channel-window and then (with setting autoclose_windows = on) promptly lost with the closed window. It should probably go to the status-window instead/too, as it does with ban, inviteonly, etc.

For reference:
see 489 (and 480) at https://defs.ircdocs.horse/defs/numerics.html
see 489 at https://github.com/ircd-hybrid/ircd-hybrid/blob/master/src/numeric.c

The mode used for this differs between networks, if you are interested in that rabbithole see: https://github.com/ircv3/ircv3-ideas/issues/45 and https://github.com/ircv3/ircv3-ideas/issues/46

Perhaps it would be best to support a "free text" style numeric. but thats just IMHO and without having looked into how numerics are implemented in Irssi: ":%s 489 %s :%s"

falsovsky:
This problem also affects me with ircd-hybrid.

mh-source:
@falsovsky and/or anybody else who might be interested. I made a quick and dirty scripted fix for use in my own testing. while we wait... :-)

[mh_secureonlychan_489.pl.txt](https://github.com/irssi/irssi/files/4622730/mh_secureonlychan_489.pl.txt)

falsovsky:
Thank you, I'm still @ work. But later in the day I'll test and report.

falsovsky:
@mh-source It works!

mh-source:
_If_ this code gets approved (at some point) I am wondering about putting the secureonlychan script on Irssi-scripts so people who might be forced/prefer to use an older version of Irssi can still enjoy the wonderful world of secure only channels in their full glory. And adding a version check to the script, telling people if they don't need it. any opinions (for/against) this idea?

mh-source:
With both a source-code fix in git and a fall-back solution in the form of a script available on irssi.org, i think this issue can be close. Thanks to everyone for your patience and support.

PR #1196, opened 2020-05-15
Add better support for numeric 489 ERR_SECUREONLYCHAN along with a print format

mh-source:
This is a first attempt at adding support for numeric 489 as ERR_SECUREONLYCHAN in addition to the already supported ERR_VOICENEEDED. To fix my own submitted issue #1193 

It adds a new event_489 handler and the format string 'joinerror_secure_only'. but it is not unlikely i have missed some other (required) details for adding a numeric and format string.

It is ofcourse tested (at least for ERR_SECUREONLYCHAN, i dont know where it is used as ERR_VOICENEEDED so cant test triggering that) and "it works on my computer" :-)

modified:   src/fe-common/irc/fe-events-numeric.c
modified:   src/fe-common/irc/module-formats.c
modified:   src/fe-common/irc/module-formats.h


ailin-nemui:
thanks :-)

mh-source:
Now the failed tests have been fixed and I tested the ERR_VOICENEEDED on undernet with success. Hopefully no more loose ends.

ailin-nemui:
@irssi/developers

dequis:
Sure, looks solid. We should do something about those `clang-format off` at some point later (shouldn't block this, good to keep them as a marker)

PR #1197, opened 2020-05-21
use visible channel name in layout save

lotheac:
Hi,

there's something of a long standing problem with /layout save and !-channels (eg. on IRCnet). Consider the following initial config file, without channels or windows saved:

```
servers = (
  {
    address = "irc.cs.hut.fi";
    chatnet = "IRCnet";
    port = "6667";
    autoconnect = "yes";
  }
);

chatnets = {
  IRCnet = {
    type = "IRC";
    max_kicks = "1";
    max_msgs = "1";
    max_whois = "1";
  };
};

settings = {
  core = {
    real_name = "irssi-chansave";
    user_name = "irssi-chansave";
    nick = "irssi-chansave";
  };
  "fe-text" = { actlist_sort = "refnum"; };
};
windows = {
  1 = { immortal = "yes"; name = "(status)"; level = "ALL"; };
};
```

When irssi is started on this config file, and a !-channel is joined (`/join !!new-channel`, or `/join !existing-channel`), a new window is created and the channel item is added to that window, as expected. `/window` will report:
```
11:27 Window  : #3
11:27 Size    : 113x54
11:27 Level   : NONE
11:27 Server  : IRCnet
11:27 Items   : Name                           Server tag
11:27  CHANNEL: !existing-channel              IRCnet
```

Now, however, if `/layout save` is used by the user who wants this channel to stay in the same window next time they run irssi, a discrepancy is created:

```
11:29 Window  : #3
11:29 Size    : 113x54
11:29 Level   : NONE
11:29 Server  : IRCnet
11:29 Bounds  : Name                           Server tag
11:29         : !KYKAFexisting-channel         IRCnet          sticky
11:29 Items   : Name                           Server tag
11:29  CHANNEL: !existing-channel              IRCnet
```

When irssi is restarted, window #3 is again created with the bound:
```
11:29 Window  : #3
11:29 Size    : 113x54
11:29 Level   : NONE
11:29 Server  : IRCnet
11:29 Bounds  : Name                           Server tag
11:29         : !KYKAFexisting-channel         IRCnet          sticky
```

but when !existing-channel is rejoined, it does not match the bound on window 3, creating a new window instead and leaving window 3 empty.

So, I propose this patch to modify layout saving to create a binding for the visible_name of the channel instead, which makes this scenario of saving layouts for !-channels work correctly.

### DIFF
--- a/irssi.conf
+++ b/irssi.conf
@@ -107,7 +107,7 @@ channels = (
 
 aliases = {
   ATAG        = "WINDOW SERVER";
-  ADDALLCHANS = "SCRIPT EXEC foreach my \\$channel (Irssi::channels()) { Irssi::command(\"CHANNEL ADD -auto \\$channel->{name} \\$channel->{server}->{tag} \\$channel->{key}\")\\;}";
+  ADDALLCHANS = "SCRIPT EXEC foreach my \\$channel (Irssi::channels()) { Irssi::command(\"CHANNEL ADD -auto \\$channel->{visible_name} \\$channel->{server}->{tag} \\$channel->{key}\")\\;}";
   B           = "BAN";
   BACK        = "AWAY";
   BANS        = "BAN";
--- a/src/fe-common/core/windows-layout.c
+++ b/src/fe-common/core/windows-layout.c
@@ -169,12 +169,12 @@ static void sig_layout_save_item(WINDOW_REC *window, WI_ITEM_REC *item,
 		chat_protocol_find_id(item->chat_type);
 	if (proto != NULL)
 		iconfig_node_set_str(subnode, "chat_type", proto->name);
-	iconfig_node_set_str(subnode, "name", item->name);
+	iconfig_node_set_str(subnode, "name", item->visible_name);
 
 	if (item->server != NULL) {
 		iconfig_node_set_str(subnode, "tag", item->server->tag);
 		if (IS_CHANNEL(item)) {
-			rec = window_bind_add(window, item->server->tag, item->name);
+			rec = window_bind_add(window, item->server->tag, item->visible_name);
 			if (rec != NULL)
 				rec->sticky = TRUE;
 		}


ailin-nemui:
interesting! I wonder if there is any argument for this way or making sure that the idchan is matched by full ID on the bond when rejoining?

lotheac:
> interesting! I wonder if there is any argument for this way or making sure that the idchan is matched by full ID on the bond when rejoining?

I'm not sure. But actually it turns out I spoke a bit too soon - this doesn't fully work. It creates the binding correctly on !existing-channel, but on exit the full id is written to the config file anyway.

lotheac:
There's two arguments that come to my mind against storing the full timestamped name for !-channels in the configuration file:

1. if the !-channel to be joined does not exist the next time that irssi is started, a new incorrectly-named channel is automatically created. eg. when I had `!KYKAFexisting-channel` configured in both `channels` and `windows`, the correct saved window was used, but I ended up on the channel `!KYLPPKYKAFexisting-channel` instead. If the prefix is not present in the config file, a new channel is still created, but it will have the correct name (though the prefix changes, obviously).
2. `/channel add !existing-channel` is probably a more friendly user experience than having to figure out what the channel prefix before adding a channel -- I think that information isn't particularly easy to find in irssi.

lotheac:
Updated patch to now actually also write the visible_name to the config file. Now, the resulting config actually keeps the channel in the intended window, regardless of whether it was created by autojoining, and the binding matches the item in `/window`:

```
12:16 Window  : #2
12:16 Size    : 113x54
12:16 Level   : NONE
12:16 Server  : IRCnet
12:16 Bounds  : Name                           Server tag
12:16         : !existing-channel              IRCnet          sticky
12:16 Items   : Name                           Server tag
12:16  CHANNEL: !existing-channel              IRCnet
```

I also updated the ADDALLCHAN alias to add channels with their visible_name.

ailin-nemui:
@irssi/developers

PR #1199, opened 2020-06-02
Add support for building in Termux-Android in meson

ailin-nemui:
- introduce cross perl
- workaround for the android linker


ailin-nemui:
@irssi/developers 

can anyone comment especially on disable setuid/setgid on android; is it ok to make it like that or should this be explicitly requested with some flag?

PR #1204, opened 2020-07-04
fix extended bg colours

ailin-nemui:


ailin-nemui:
these got broken in #1079

ailin-nemui:
@irssi/developers

PR #1205, opened 2020-07-04
properly restore the saved text dest

ailin-nemui:
fixes %[-t] on saved prints


ailin-nemui:
related to #1079

ailin-nemui:
@irssi/developers

PR #1206, opened 2020-07-05
Fix crash from self-unloading script

aquanight:
Fixes a crash that can be caused by self-unloading scripts (or other cases of unloading scripts that also have active code), by guarding against multiple calls of perl_script_unload (by /script unload and by sig_script_error).

The particular test case that prompted this is http://urchlay.naptime.net/repos/misc-scripts/tree/selfunload.pl

The crash in the above test case was caused by the following:

1. The script unloads itself using Irssi::command("/script unload selfunload")
2. This leads to perl_script_unload which in turn destroys the script package and then calls perl_script_destroy (freeing the PERL_SCRIPT_REC).
3. Execution then returns to the perl script. An insertion of a print to STDERR can confirm that execution does in fact return to the script (the original version incorrectly believed the first command() never returned - it does).
4. However, the second command() refers to command() unqualified. Before this was imported via the use statement, but due to step 2, this import was undefined, so this second command causes a perl error.
5. The unhandled error triggers a "script error" signal, which sig_script_error() then calls perl_script_unload. On the same PERL_SCRIPT_REC that had already had perl_script_unload called upon it.
6. This leads to perl_script_unload executing against a free()'d memory block, and eventually leads to perl_script_destroy again which results in a double-free of the PERL_SCRIPT_REC and its constituent variables. And so everything explodes.

The resolution is to make PERL_SCRIPT_REC ref-counted. Some of the unload logic that previously existed in perl_script_destroy was moved to perl_script_unload: e.g. de-registration of signals and event handlers and removal from the active-scripts list. perl_script_destroy now is exclusively free()ing of the PERL_SCRIPT_REC (and still is responsible for generating the "script destroyed" signal. The signal and input/timeout handling trampolines will now ref/unref the PERL_SCRIPT_REC, so it will not be destroyed by perl_script_unload until these handlers complete.

### DIFF
--- a/src/perl/perl-core.c
+++ b/src/perl/perl-core.c
@@ -65,7 +65,6 @@ static void perl_script_destroy_package(PERL_SCRIPT_REC *script)
 
 static void perl_script_destroy(PERL_SCRIPT_REC *script)
 {
-	perl_scripts = g_slist_remove(perl_scripts, script);
 
 	signal_emit("script destroyed", 1, script);
 
@@ -274,6 +273,7 @@ static PERL_SCRIPT_REC *script_load(char *name, const char *path,
 	script->package = g_strdup_printf("Irssi::Script::%s", name);
 	script->path = g_strdup(path);
         script->data = g_strdup(data);
+	script->refcount = 1;
 
 	perl_scripts = g_slist_append(perl_scripts, script);
 	signal_emit("script created", 1, script);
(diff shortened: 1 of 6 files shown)


ailin-nemui:
nice!

src/perl/perl-core.c
```diff
@@ -311,10 +308,40 @@ PERL_SCRIPT_REC *perl_script_load_data(const char *data)
...
+	perl_source_remove_script(script);
+
+	link = g_slist_find(perl_scripts, script);
+	if (link != NULL) {
+		perl_scripts = g_slist_remove_link(perl_scripts, link);
+		g_slist_free(link);
+		perl_script_unref(script);
+	}
+
+/*	perl_scripts = g_slist_remove(perl_scripts, script);
```
ailin-nemui, line 326: can this commented code be removed entirely?

aquanight: Can and did.

ailin-nemui:
@irssi/developers

PR #1209, opened 2020-07-22
fix missing wrapping of line in signals

ailin-nemui:
related to #1079 /  57fb173130cec6e82bccfbe6ead09731b591c5a6

the signals that include LINE_REC*s do not have any possibility to wrap these, so we have to create some more code

otherwise we get a segv in perl_line_fill_hash, which is passed a raw LINE_REC* when called from plain_bless
 in perl-signals.c


ailin-nemui:
@irssi/developers

PR #1214, opened 2020-07-27
Update ignore help

vague666:


ailin-nemui:
I don't like the help yet but have yet to think of a better help text

`hide matching activity' sounds odd, we're hiding (lines?messages?text?)

Maybe something like?
> The special level 'HIDDEN' can be used to hide messages that can be revealed using WINDOW HIDELEVEL

ailin-nemui:
@irssi/developers

Issue #1220, opened 2020-09-08
The stock `colorless` theme does not de-color the vertical window split bar


nachanon:
_This is a theme bug with no effect on Irssi's functionality._

I have recently been checking out Irssi's default-installed alternative `colorless` theme as an alternative to default theme. However, as soon as I applied it to my horizontally-split Irssi session; I found that the theme does not correctly (de)color the vertical window split bar; which starkly contrasted with the rest of window.

This is more-or-less trivial to fix (with caveat), pull request will follow.

### Exact Steps to Reproduce ###

1. Run `irssi --noconnect`
1. Issue command: `/set theme colorless`
1. Issue command: `/quit`
1. Run `irssi --noconnect`
1. Issue command: `/window new -right split`
1. Issue command: `/1`
1. You would see that vertical window split bar is not colorless, which is against the point of using `colorless` theme...
	
	![(Screenshot)](https://u.cubeupload.com/xwindows/jSkFjv.png)

### Side Notes ###

- My terminal's default foreground/background color combo is `#FFFCDA`/`#000000`.
- This is not the only problem I found in stock `colorless` theme, other notable one is the lack of messages' bold highlights in several places.

### System Information ###

Irssi: 1.2.2 (20190829 0225)  
System: Fedora 32 (GNU/Linux x86_64)

ailin-nemui:
thanks for the issue report, this is an oversight. looking to your PR

PR #1221, opened 2020-09-08
Decolor the vertical window split bar in stock `colorless` theme

nachanon:
Current Irssi's stock `colorless` theme does not contain any code to de-color vertical window split bars (i.e. bug #1220), resulting in them being colored with default blue background; defeating the purpose of the theme.

This one-line change introduces following line into `themes/colorless.theme`:

	  window_border = "%8";

Which makes the vertical window split bars use reverse-video of default terminal foreground/background instead.

### Screenshots ###

Before the fix:
(Screenshot)

**After** the fix:
(Screenshot)

### Caveat ###

This fix brings out a quirk in Irssi's styling implementation&mdash; where vertical window split bar's styling overlapped with the bottom status bar's styling, resulting in a black cell at the middle of bottom status bar. (Status bar used reverse video, and split bar reversed it back)

Fixing the quirk itself might entail redefining the extent of of window split bar and/or status bar drawing inside Irssi's main codebase (with consideration on effects on other themes' compatibility); but that warrants a dedicated ticket for another day.

Having this fix is likely to be better than leaving color in `colorless` theme of course.

### System Information ###

Irssi: 1.2.2 (20190829 0225)<sup>\*</sup>  
System: Fedora 32 (GNU/Linux x86_64)

<sup>\*</sup> The change is tested by installing the new theme file as a user's own theme file.

### DIFF
--- a/themes/colorless.theme
+++ b/themes/colorless.theme
@@ -251,6 +251,7 @@ abstracts = {
   # default background for all statusbars. You can also give
   # the default foreground color for statusbar items.
   sb_background = "%8";
+  window_border = "%8";
 
   # default backround for "default" statusbar group
   #sb_default_bg = "%8";


ailin-nemui:
@irssi/developers

PR #1232, opened 2020-10-27
Implement priority on the matched lines for /hilight

### LINKED ISSUE #1228: hilight does not respect -priority arg
I have the following `hilight` configuration

```
 /hilight
...
19:34    3 ccc  -line -priority 20 -color %G -actcolor %g
19:34    4 ddd  -line -priority 1 -color %p -actcolor %p
19:34    5 fff  -line -priority 100 -color %k -actcolor %k
```

**Actual:**
After receiving a message containing all the matching words like `ccc ddd fff` the message always gets highlighted according to the first positional match in the config.

**Expected:**
highlight the message according to the `-priority` arg.

vague666:
Since the original use case wasn't broken(Fixes #1228) this adds more functionality instead :)

### DIFF
--- a/src/fe-common/core/hilight-text.c
+++ b/src/fe-common/core/hilight-text.c
@@ -273,14 +273,14 @@ HILIGHT_REC *hilight_match(SERVER_REC *server, const char *channel,
 	for (tmp = hilights; tmp != NULL; tmp = tmp->next) {
 		HILIGHT_REC *rec = tmp->data;
 
-		if (rec->priority > priority && !rec->nickmask &&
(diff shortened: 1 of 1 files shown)


ailin-nemui:
maybe you forgot to set some variable to NULL?

vague666:
Only possible var is tmprec, maybe I am remembering wrong that it is NULL when declared?

ailin-nemui:
now it doesn't crash anymore :-) only the code formatting is left to clean up to confirm to the new standard

vague666:
Maybe I do too much perl and forget that not all languages initialize variables :)

ailin-nemui:
@irssi/developers ?

PR #1234, opened 2020-11-11
Copy Perl signal arguments prior to running callbacks

### LINKED ISSUE #1233: memory corruption in Perl stack during signal emission
https://github.com/irssi/irssi/blob/master/src/perl/perl-signals.c#L209

crash on SvOK

reported by @RealKindOne

ailin-nemui:
Might fix #1233

### DIFF
--- a/.github/workflows/clangformat/clang-format-xs
+++ b/.github/workflows/clangformat/clang-format-xs
@@ -38,6 +38,7 @@ while :; do
 	    ;;
 	--lines)
 	    lines+=("$2")
+	    off_opts+=("$1" "$2")
 	    shift 2
 	    continue
 	    ;;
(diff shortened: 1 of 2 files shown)


dequis:
I guess the croaks around lines 100-102 might leak memory?

ailin-nemui:
that would seem likely...

ailin-nemui:
it would probably suffice to copy the arguments before the callback, what do you think would be better?

https://github.com/irssi/irssi/blob/e0292665c1adef5a1d8893555f4810dc5b72ffc6/src/perl/perl-signals.c#L206

dequis:
Yeah, maybe, if that makes the diff more minimal and you understand the scope of the issue. I'm not terribly confident about this one, the perl API is a lot.

ailin-nemui:
@irssi/developers

dequis (approved):
I hope there's no secret croaking anywhere

PR #1248, opened 2021-01-04
put proper versions in PACKAGE_VERSION

ailin-nemui:
we should ensure PACKAGE_VERSION aka `$J` resembles a version

could be broken if you tag something


ailin-nemui:
@irssi/developers

dequis:
What's the motivation for this one? What's a bad version, what's a good version, what happens if you have a bad version?

ailin-nemui:
an arbitrary tag such as "test" will get saved as the version (/eval echo $J), breaking scriptassist and possibly other things that want to use this value

ailin-nemui:
screwed up my git skills -> https://github.com/irssi/irssi/pull/1253

PR #1250, opened 2021-01-06
better account tracking

ailin-nemui:


ailin-nemui:
@irssi/developers

ailin-nemui:
https://github.com/mquin/irssi-scripts/blob/master/account-notify.pl

ailin-nemui:
if disable extended-join, the chase account code runs into `verne.freenode.net 263 Nei WHO :This command could not be completed because it has been used recently, and is rate-limited.` on freenode

at least make sure the chase is run only 1 time per user

ailin-nemui:
we're now only querying one user once even if they join multiple channels, but still:

I get the 263 error for something like `WHO losuler %tna`

the common issue seems to be that the person (nick) has left already (QUIT the network or changed nick)
at the time my queued who command makes it through

ailin-nemui:
now we purge the WHOs from the queue. hope that's good enough

ailin-nemui:
@irssi/developers

ailin-nemui:
here is a script that makes use of the account field http://anti.teamidiot.de/static/nei/*/Code/Irssi/account_expando.pl

ailin-nemui:
@irssi/developers

dequis:
"update clang-format to clang-format-11, fixes enum bug" what enum bug?

ailin-nemui:
clang-format < 11 will want to force-format  enums like this, iirc:
```
enum {	CHANNEL_QUERY_MODE,
		CHANNEL_QUERY_WHO,
		CHANNEL_QUERY_BMODE,

		CHANNEL_QUERIES
};
```

src/irc/core/channels-query.c
```diff
@@ -50,22 +50,7 @@
...
-
-#define CHANNEL_IS_MODE_QUERY(a) ((a) != CHANNEL_QUERY_WHO)
-
-typedef struct {
-	int current_query_type; /* query type that is currently being asked */
-        GSList *current_queries; /* All channels that are currently being queried */
-
-	GSList *queries[CHANNEL_QUERIES]; /* All queries that need to be asked from server */
-} SERVER_QUERY_REC;
+#define WHO_Ps_PPtna_745 "WHO %s %%tna,745"
```
dequis, line 53: what even is this name

src/irc/core/massjoin.c
```diff
@@ -39,12 +39,15 @@ static void event_join(IRC_SERVER_REC *server, const char *data,
...
 	g_return_if_fail(data != NULL);
 
 	if (g_ascii_strcasecmp(nick, server->nick) == 0) {
-		/* You joined, no need to do anything here */
-		return;
+		/* You joined, do not massjoin */
+		send_massjoin = FALSE;
+	} else {
+		send_massjoin = TRUE;
 	}
```
dequis, line 51: Hmm, can you explain this part? The return not being here means we call irc_nicklist_insert, which we previously didn't at this point. What is this for?

I think I understand the massjoin part but don't care much about it.

src/irc/core/servers-redirect.c
```diff
@@ -442,7 +442,7 @@ static void redirect_abort(IRC_SERVER_REC *server, REDIRECT_REC *rec)
 		g_free(str);
 
 		if (rec->failure_signal != NULL)
-			signal_emit(rec->failure_signal, 1, server);
+			signal_emit(rec->failure_signal, 3, server, rec->cmd->name, rec->arg);
```
dequis, line 445: Is this safe?

src/irc/core/irc-nicklist.c
```diff
@@ -510,7 +601,10 @@ void irc_nicklist_init(void)
 {
 	signal_add_first("event nick", (SIGNAL_FUNC) event_nick);
 	signal_add_first("event 352", (SIGNAL_FUNC) event_who);
+	signal_add_first("event 354", (SIGNAL_FUNC) event_whox_743);
 	signal_add("silent event who", (SIGNAL_FUNC) event_who);
+	signal_add("silent event whox", (SIGNAL_FUNC) event_whox_743);
+	signal_add("silent event whox useraccount", (SIGNAL_FUNC) event_whox_745);
```
dequis, line 607: What do 743 and 745 mean here?

src/irc/core/channels-query.c
```diff
@@ -91,6 +78,8 @@ static void sig_disconnected(IRC_SERVER_REC *server)
 	rec = server->chanqueries;
 	g_return_if_fail(rec != NULL);
 
+	g_hash_table_foreach(rec->accountqueries, (GHFunc) g_free, NULL);
+	g_hash_table_destroy(rec->accountqueries);
```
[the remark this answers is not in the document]
ailin-nemui: good catch! double free here, I initially copied the code without a DestroyNotify!

but the values don't need to be freed here AFAIK (maybe you can double check for me) as we're using this table as a set only

src/irc/core/channels-query.c
ailin-nemui: the name is a direct copy of the string content... can you suggest a good name? The reason I have this  as a constant is that the similar string is used in more than 1 place and this way when editing the code later on I will be sure to notice that the string must be changed in more than 1 place

src/irc/core/massjoin.c
ailin-nemui: The goal is that we fill in our OWN account here as well, if we receive it. Otherwise we would need more special code for that

ailin-nemui: should I expand the comment?

src/irc/core/servers-redirect.c
ailin-nemui: It's an API change, we add 2 more arguments to the failure signal. As far as I know existing code is also sloppy on arguments so this should not be more unsafe than what we have in other places.

(In general the whole signal emission system is a bit unsafe)

src/irc/core/irc-nicklist.c
ailin-nemui: Good point! These are arbitrary 3-digit numerics that the client should choose for their WHOX queries. I threw dice to get them. Make them a constant? Choose different numbers?

src/irc/core/channels-query.c
dequis: oooh lol, P means percentage, now i get it. But okay, just add a comment next to it explaining what it means? WHOX is cryptic

src/irc/core/irc-nicklist.c
ailin-nemui: I thought about encoding the selected fields in the number but 1000 numbers is not enough for all combinations

src/irc/core/massjoin.c
dequis: I think the comment is fine for the code, I'm just wondering about the diff. So it's fine to just do the irc_nicklist_insert here? I guess something else might do that later with fewer details and end up as a no-op?

src/irc/core/irc-nicklist.c
dequis: Hahah dice, amazing.

How about this: Put both WHOX commands in some header, something like this

```
/* identifiers chosen by fair dice roll, guaranteed to be random */
#define WHOX_FULL_ID "743"
#define WHOX_USERACCOUNT_ID "745"

/* some explanation here about WHOX because it's a bullshit spec */
#define WHOX_FULL_CMD "WHO %s %%tcuhnfdar," WHOX_FULL_ID
#define WHOX_USERACCOUNT_CMD "WHO %s %%tna," WHOX_USERACCOUNT_ID
```

And rename the callbacks from event_whox_745 to event_whox_useraccount, etc.

src/irc/core/massjoin.c
ailin-nemui: normally our own nick gets inserted only as part of `event_names_list` in irc-nicklist.c when the NAMES are received after join (with a special case further down to insert our own nick in any case if it was NOT contained in NAMES). The code in massjoin.c handles all the JOINs that happen after we are joined to a channel.

When I relax that check, the (special?) self-JOIN that we get when joining a channel will also be fed through the JOIN code in massjoin.c, with the benefit that we can leech on the extended-account there.

As a consequence, I had to change irc-nicklist.c's NAMES handler to update our CHANNEL USER MODES from the NAMES.

I don't know if all of this is a very bad(tm) idea and if it would be better to do it in some other way.

What do you think? Can you follow my explanation?

src/irc/core/irc-nicklist.c
ailin-nemui: thanks, I implemented your suggestion

src/irc/core/channels-query.c
ailin-nemui: now resolved by your suggestion below

src/irc/core/channels-query.c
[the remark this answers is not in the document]
ailin-nemui: removed that foreach now

ailin-nemui:
@dequis friendly ping :-)

Issue #1239, opened 2020-12-09
Crashes with /connect 


ailin-nemui:
```
/connect /dev/null
/connect -ssl_cert ... /home/luser/sockets/freenode-irc-orwell
```

`stunnel4 stunnel_sock.conf`
```
pid = /home/luser/run/stunnel.pid
output = /home/luser/log/stunnel_sock.log
[freenode-irc-orwell]
client = yes
accept = /home/luser/sockets/freenode-irc-orwell
connect = orwell.freenode.net:6697
#checkHost = orwell.freenode.net
```

reported by Koffa

Issue #1200, opened 2020-06-12
xdg specification config file


LCRERGO:
I've been experimenting with putting irssi config file in $HOME/.config/irssi by altering:
IRSSI_DIR_FULL "%s/.irssi" to "%s/.config/irssi". Are there any other places where I should care for problems with this change?

ailin-nemui:
- scripts use IRSSI_DIR_FULL "/scripts" but they belong into $XDG_DATA_HOME/irssi/scripts
- themes use IRSSI_DIR_FULL, but they belong into $XDG_DATA_HOME/irssi/themes
- awaylog uses IRSSI_DIR_FULL "/away.log", but it (probably) belongs into $XDG_CACHE_HOME
- several scripts such as adv_windowlist or nicklist create pipes in IRSSI_DIR_FULL which belong into $XDG_RUNTIME_DIR

vague666:
@LCRERGO, why not just use `irssi --home ~/.config/irssi` ? You could even alias it so you don't have to think about it and just run `irssi`

LCRERGO:
> @LCRERGO, why not just use `irssi --home ~/.config/irssi` ? You could even alias it so you don't have to think about it and just run `irssi`

Because I'm using DWM and I'm having some problem in using bash aliases with it. So I thought it would be easier to alter irssi source code to support it, and indeed it is.

iambumblehead:
support for XDG_CONFIG_HOME would be great

Issue #157, opened 2014-10-21
Irssi should respect XDG Base Directory Specification


josteink:
As the title says: Irssi should respect [XDG Base Directory Specification](http://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html).

Long story short: Irssi (among tons of other apps) auto-sets its home-directory to `~/.irssi`. This contributes to a cluttered home-folder full of dot-files and dot-folders.

XDG Base Directory Specification says these should be moved from `~/.app` to `$XDG_CONFIG_HOME/app` which typically resolves to `~/.config/app`.

To integrate this with prior dot-file conventions irssi must do the following:

* When a config exists in the XDG_CONFIG_HOME-folder, irssi should prefer this one. 
* When no config exists in XDG_CONFIG_HOME, it should revert to searching ~/.irssi
* When no config ecists anywhere Irssi should default to the XDG_CONFIG_HOME.

Sounds doable? Pretty please? :)

BTW: Welcome back to the world of the living (software project wise)!

PR #1251, opened 2021-01-06
[WIP] add: xdg-support

realaltffour:
- config file XDG support inspired from #511
- added support to try loading scripts from XDG_DATA_HOME, if not present try
  loading from XDG_CONFIG_HOME, if not present try loading the usual way.

### DIFF
--- a/scripts/scriptassist.pl
+++ b/scripts/scriptassist.pl
@@ -204,7 +204,13 @@ sub get_names {
     $xname =~ s/\W/_/g;
     my $pname = "${xname}::";
     if ($xname ne $sname || $sname =~ /_/) {
-	my $dir = Irssi::get_irssi_dir()."/scripts/";
+	my $dir = "";
+	if (Irssi::is_xdg_supported()) {
+	    $dir = Irssi::get_runtime_dir()."/scripts/";
+	}
+	else {
+	    $dir = Irssi::get_irssi_dir()."/scripts/";
+	}
 	if ($db && exists $db->{"$sname.pl"}) {
 	    # $found = 1;
 	} elsif (-e $dir.$plname || -e $dir."$sname.pl" || -e $dir."autorun/$sname.pl") {
--- a/scripts/scriptassist.pl
+++ b/scripts/scriptassist.pl
@@ -204,13 +204,7 @@ sub get_names {
     $xname =~ s/\W/_/g;
     my $pname = "${xname}::";
     if ($xname ne $sname || $sname =~ /_/) {
-	my $dir = "";
-	if (Irssi::is_xdg_supported()) {
-	    $dir = Irssi::get_runtime_dir()."/scripts/";
-	}
-	else {
-	    $dir = Irssi::get_irssi_dir()."/scripts/";
-	}
+	my $dir = Irssi::get_irssi_dir()."/scripts/";
 	if ($db && exists $db->{"$sname.pl"}) {
 	    # $found = 1;
 	} elsif (-e $dir.$plname || -e $dir."$sname.pl" || -e $dir."autorun/$sname.pl") {
--- a/src/core/core.c
+++ b/src/core/core.c
@@ -367,9 +367,9 @@ void core_deinit(void)
 	modules_deinit();
 
 	if (irssi_dir != irssi_cache_dir) {
-	    g_free(irssi_config_dir);
-	    g_free(irssi_cache_dir);
-	    g_free(irssi_runtime_dir);
+		g_free(irssi_config_dir);
+		g_free(irssi_cache_dir);
+		g_free(irssi_runtime_dir);
 	}
 	g_free(irssi_dir);
         g_free(irssi_config_file);
--- a/src/core/core.c
+++ b/src/core/core.c
@@ -216,7 +216,7 @@ void core_preinit(const char *path)
 	if (irssi_dir == NULL) {
 		/* check if %XDG_CONFIG_HOME/irssi exists and use it */
 		char *dirp = g_build_filename(g_get_user_config_dir(), "irssi", NULL);
-		if (!g_file_test(dirp, G_FILE_TEST_EXISTS)) {
+		if (g_file_test(dirp, G_FILE_TEST_EXISTS)) {
 			use_xdg = 1;
 			irssi_dir = g_strdup_printf("%s/irssi", g_get_user_data_dir());
 			irssi_config_dir = g_strdup_printf("%s/irssi", g_get_user_config_dir());
--- a/src/core/core.c
+++ b/src/core/core.c
@@ -236,9 +236,8 @@ void core_preinit(const char *path)
 			irssi_dir[len - 1] = '\0';
 	}
 	if (irssi_config_file == NULL)
-		irssi_config_file = use_xdg ? g_build_filename(g_get_user_config_dir(), "irssi",
-		                                               IRSSI_HOME_CONFIG, NULL) :
-                                              g_strdup_printf("%s/" IRSSI_HOME_CONFIG, irssi_dir);
+	   irssi_config_file = use_xdg ? g_build_filename(g_get_user_config_dir(), "irssi", IRSSI_HOME_CONFIG, NULL) :
+	      g_strdup_printf("%s/" IRSSI_HOME_CONFIG, irssi_dir);
 	else {
 		str = irssi_config_file;
 		irssi_config_file = fix_path(str);
--- a/src/core/core.c
+++ b/src/core/core.c
@@ -189,26 +189,30 @@ void core_register_options(void)
 
 void core_preinit(const char *path)
 {
-	const char *home;
 	char *str;
 	int len;
 
 	if (irssi_dir == NULL) {
-		home = g_get_home_dir();
-		if (home == NULL)
-			home = ".";
+	   irssi_dir = g_build_filename(g_get_user_config_dir(), "irssi", NULL);
+	   if (home == NULL)
+	      if (g_file_test(irssi_dir, G_FILE_TEST_IS_DIR) == FALSE) {
+		 home = ".";
+		 g_free(irssi_dir);
 
-		irssi_dir = g_strdup_printf(IRSSI_DIR_FULL, home);
+		 const char *home;
+		 home = g_get_home_dir();
+		 irssi_dir = g_build_filename(home? home: ".", ".irssi", NULL);
+	      }
 	} else {
-		str = irssi_dir;
-		irssi_dir = fix_path(str);
-		g_free(str);
-		len = strlen(irssi_dir);
-		if (irssi_dir[len-1] == G_DIR_SEPARATOR)
-			irssi_dir[len-1] = '\0';
+	   str = irssi_dir;
+	   irssi_dir = fix_path(str);
+	   g_free(str);
+	   len = strlen(irssi_dir);
+	   if (irssi_dir[len-1] == G_DIR_SEPARATOR)
+	      irssi_dir[len-1] = '\0';
 	}
 	if (irssi_config_file == NULL)
-		irssi_config_file = g_strdup_printf("%s/"IRSSI_HOME_CONFIG, irssi_dir);
+	   irssi_config_file = g_build_filename(irssi_dir, IRSSI_HOME_CONFIG, NULL)
 	else {
 		str = irssi_config_file;
 		irssi_config_file = fix_path(str);
--- a/src/core/core.c
+++ b/src/core/core.c
@@ -222,7 +222,7 @@ void core_preinit(const char *path)
 			irssi_config_dir = g_strdup_printf("%s/irssi", g_get_user_config_dir());
 			irssi_cache_dir = g_strdup_printf("%s/irssi", g_get_user_cache_dir());
 			irssi_runtime_dir =
-			    g_strdup_printf("%s/irssi", g_get_user_runtime_dir(), "irssi", NULL);
+			    g_strdup_printf("%s/irssi", g_get_user_runtime_dir());
 		} else { /* fallback to non-xdg location */
 			home = g_get_home_dir();
 			if (home == NULL)
--- a/src/core/core.c
+++ b/src/core/core.c
@@ -65,22 +65,23 @@ void log_away_deinit(void);
 void wcwidth_wrapper_init(void);
 void wcwidth_wrapper_deinit(void);
 
+int xdg_support;
 int irssi_gui;
 int irssi_init_finished;
 int sighup_received;
 time_t client_start_time;
 
-static char *irssi_dir, *irssi_xdg_dir, *irssi_config_file;
+static char *irssi_dir, *irssi_config_file;
 static GSList *dialog_type_queue, *dialog_text_queue;
+
+int is_xdg_supported()
+{
+	return xdg_support;
+}
 
 const char *get_irssi_dir(void)
 {
-        return irssi_dir;
-}
-
-const char *get_irssi_xdg_dir(void)
-{
-	return irssi_xdg_dir;
+	return (xdg_support = 1) ? irssi_xdg_dir : irssi_dir;
 }
 
 /* return full path for ~/.irssi/config */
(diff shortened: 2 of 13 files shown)


ailin-nemui:
please consider

- https://github.com/irssi/irssi/pull/51#issuecomment-60132604
  > I agree that XDG makes sense however I worry implementation of it would be incredibly confusing.
  > 
  > For example (I haven't thought through all the aspects of this):
  > 
  > Currently scripts both live under `~/.irssi/scripts` and `~/.irssi/scripts/autorun` (although ideally a symlink to the parent), if we strictly interpret the xdg base directory spec the autorun aspect should be treated as configuration while the script itself should be treated as data.
  > 
  > This means scripts should go under `~/.local/share/irssi/scripts` but then with symlinks under `~/.config/irssi/scripts/autorun` (maybe -- maybe we switch to a different way of specifying autorun as this is a bit cumbersome).
  > 
  > For this particular thing that would mean changes in:
  > 
  > - API to add variants on `Irssi::get_irssi_dir();` (perl and C)
  > - scriptassist
  > - Docs
  > 
  > Additionally it would mean introducing a search order for scripts (and probably code to warn users if they have a script named the same thing in both `~/.local/share/irssi/scripts` and `~/.irssi/scripts`).
  > 
  > In general I don't think the cost of this is worth it. (Remember Irssi has users who are not hugely familiar with Unix and use this on a shell account; the cost of making a change and then obsoleting random tutorials on the web is very large).
- https://wiki.archlinux.org/index.php/XDG_Base_Directory#Specification
- https://github.com/irssi/irssi/issues/1200#issuecomment-643310943
  > - scripts use IRSSI_DIR_FULL "/scripts" but they belong into $XDG_DATA_HOME/irssi/scripts
  > - themes use IRSSI_DIR_FULL, but they belong into $XDG_DATA_HOME/irssi/themes
  > - awaylog uses IRSSI_DIR_FULL "/away.log", but it (probably) belongs into $XDG_CACHE_HOME
  > - several scripts such as adv_windowlist or nicklist create pipes in IRSSI_DIR_FULL which belong into $XDG_RUNTIME_DIR
- `--config` and `--home` must take precedence
- never look in multiple places in the user's folders, use the first folder that exists
- only look in XDG_DATA_HOME for scripts, and only look there if using XDG_CONFIG_HOME
- **only** fix clang-format errors of the code that you touched, this is important for code review. use the `git-clang-format` command to do this, or download the "artefact" from github actions to manually patch the requested diff. also see https://github.com/irssi/irssi/wiki/clang-format#manual-check-after-git-add 

in general it will require some work before we can accept this feature. I haven't found a definite answer for the script autorun mechanism in particular

ailin-nemui:
about mixed declarations, just declare your variables at the beginning of a block! Instead of 

    ...
    const char *x = y();
    ...

write

    {
        const char *x;
        ...
        x = y();

realaltffour:
@ailin-nemui added fallback which sets `irssi_dir` with either `~/.irssi` or `~/.config/irssi`, depending on `XDG_CONFIG_HOME` existence. Then it enables `is_xdg_supported()`. Making the need to check for XDG support only required when writing to outside of `XDG_CONFIG_HOME`.

Points that I need more clarification on:
> - `--config` and `--home` must take precedence

Correct me if I am wrong, but the `irssi_dir` check in `core/core.c` should do this automatically.
> - scriptassist

? grepped the code base with `gete_irssi_dir` and couldn't find this, and I have reviewed all occurrences.
> - several scripts such as adv_windowlist or nicklist create pipes in IRSSI_DIR_FULL which belong 
into $XDG_RUNTIME_DIR

I'd assume that these scripts are the defaults and present in `/scripts`?

scripts/scriptassist.pl
```diff
@@ -204,7 +204,13 @@ sub get_names {
     $xname =~ s/\W/_/g;
     my $pname = "${xname}::";
     if ($xname ne $sname || $sname =~ /_/) {
-	my $dir = Irssi::get_irssi_dir()."/scripts/";
+	my $dir = "";
+	if (Irssi::is_xdg_supported()) {
+	    $dir = Irssi::get_runtime_dir()."/scripts/";
```
ailin-nemui, line 209: here it should also be the data_dir, no?

ailin-nemui:
thanks for taking a stab at this! After checking your suggested changes, it seems that making get_irssi_dir return the XDG_DATA_DIR would more or less be identical to your solution?

in other words, running `irssi --config="$XDG_CONFIG_HOME"/irssi/config --home="$XDG_DATA_HOME"/irssi `?

if that is correct, then the solution could be as simple as adding a single if CONFIG_HOME/irssi/ exists, set the config to that dir and the irssi_dir to the DATA_HOME/irssi.

@dequis ?

ailin-nemui:
what I just described seems to be exactly what was done in #51 but rejected at that time :thinking:

ailin-nemui:
In any case I feel we have to get rid of all the IFs in your code:

for example. instead of

```
	if (is_xdg_supported())
		awaylog_file = g_build_filename(g_get_user_cache_dir(), "irssi", "away.log", NULL);
	else
		awaylog_file = g_strconcat(get_irssi_dir(), "/away.log", NULL);
```

the code should just be

```
	awaylog_file = g_build_filename(get_irssi_cache_dir(), "away.log", NULL);
```

and so on

ailin-nemui:
another caveat is that we may now need to ensure that certain directories such as cache_dir / irssi are created if they do not exist

realaltffour:
> thanks for taking a stab at this! After checking your suggested changes, it seems that making get_irssi_dir return the XDG_DATA_DIR would more or less be identical to your solution?
> 
> in other words, running `irssi --config="$XDG_CONFIG_HOME"/irssi/config --home="$XDG_DATA_HOME"/irssi `?
> 
> if that is correct, then the solution could be as simple as adding a single if CONFIG_HOME/irssi/ exists, set the config to that dir and the irssi_dir to the DATA_HOME/irssi.
> 
> @dequis ?

From my understanding `irssi_dir` is the config directory location. Therefore it would make sense to return XDG_CONFIG_HOME or ~/.irssi. I don't get why you would return `XDG_DATA_DIR` with a variable that is meant to represent the configuration folder.

> In any case I feel we have to get rid of all the IFs in your code:
> 
> for example. instead of
> 
> ```
> 	if (is_xdg_supported())
> 		awaylog_file = g_build_filename(g_get_user_cache_dir(), "irssi", "away.log", NULL);
> 	else
> 		awaylog_file = g_strconcat(get_irssi_dir(), "/away.log", NULL);
> ```
> 
> the code should just be
> 
> ```
> 	awaylog_file = g_build_filename(get_irssi_cache_dir(), "away.log", NULL);
> ```
> 
> and so on

I am going to work on this implementation.

realaltffour:
> another caveat is that we may now need to ensure that certain directories such as cache_dir / irssi are created if they do not exist

We should keep in mind that my changes only assume XDG is being used if `XDG_CONFIG_HOME/irssi` is present. I believe I saw a function that ensures that `~/.irssi` is present, if `core_preinit()` runs before it, we can test to see if the user created XDG config directory, if they did, we check/create the XDG cache, runtime and data directories. I will investigate the changes I mentioned before.

scripts/scriptassist.pl
realaltffour: Yep, I missed it, my bad

ailin-nemui:
starting to look nicely!

I was mistaken about one thing,

looks like we need get_irssi_data_dir after all, and get_irssi_dir should be == irssi_config_dir, for the `startup` file

we also need the differentiation between data and config dir for some scripts, for example the friends* scripts need configs

only question is what to do with `get_irssi_dir`, should it default to config, data, or remove it entirely?

realaltffour:
> only question is what to do with `get_irssi_dir`, should it default to config, data, or remove it entirely?

Defaulting it to data would be a sane thing to do, this would allow backward compatibility with user made scripts, and since they are scripts, they would belong to XDG data

src/core/settings.c
```diff
@@ -778,6 +778,33 @@ static void init_configfile(void)
...
+				g_error("Couldn't create %s directory: %s",
+						get_irssi_runtime_dir(), g_strerror(errno));
+			}
+		}
+		else if (!S_ISDIR(statbuf.st_mode)) {
+			g_error("%s is not a directory.\n"
+					"You should remove it with command: rm %s",
+					get_irssi_runtime_dir(), get_irssi_runtime_dir());
+		}
+	}
```
ailin-nemui, line 808: maybe extract this repeating code to a helper function?

ailin-nemui:
you could also take a look at the compilation error here https://github.com/irssi/irssi/pull/1251/checks?check_run_id=1673055256

realaltffour:
> you could also take a look at the compilation error here https://github.com/irssi/irssi/pull/1251/checks?check_run_id=1673055256

I fixed that. Do you have an idea on why my commits are failing the clang-format checks? I installed the git hook, and ensured git clang-format is ran.

ailin-nemui:
I can't say I know for sure, maybe you hadn't installed it yet when that commit was created? Please manually apply the fix from the Artefact download on this page: https://github.com/irssi/irssi/actions/runs/474336250

There still seems to be some error condition of the new code, maybe you can investigate:
https://github.com/irssi/irssi/pull/1251/checks?check_run_id=1674123860#step:6:25

Once we're confident this code is good, you will have to increase our ABI constant in https://github.com/irssi/irssi/blob/master/src/common.h#L9

src/core/core.c
```diff
@@ -189,26 +202,44 @@ void core_register_options(void)
...
 
 	if (irssi_dir == NULL) {
-		home = g_get_home_dir();
-		if (home == NULL)
-			home = ".";
-
-		irssi_dir = g_strdup_printf(IRSSI_DIR_FULL, home);
+		/* check if %XDG_CONFIG_HOME/irssi exists and use to it */
+		char *dirp = g_build_filename(g_get_user_config_dir(), "irssi", NULL);
+		if (stat(dirp, &statbuf) == 0) {
```
LemonBoy, line 214:
```suggestion
		if (!g_file_test(dirp, G_FILE_TEST_EXISTS)) {
```

And `dirp` is leaked if the`if` body is executed.

ailin-nemui:
thanks for the review @LemonBoy

src/fe-common/core/fe-common-core.c
```diff
@@ -408,7 +408,7 @@ static void autorun_startup(void)
 	GString *buf;
 	gsize tpos;
 
-	/* open ~/.irssi/startup and run all commands in it */
+	/* open $CONFIG/startup and run all commands in it */
 	path = g_strdup_printf("%s/startup", get_irssi_dir());
```
ailin-nemui, line 412: don't we need get_irssi_config_dir here as well?

src/fe-text/irssi.c
```diff
@@ -281,7 +281,7 @@ static void check_files(void)
 	struct stat statbuf;
 
 	if (stat(get_irssi_dir(), &statbuf) != 0) {
-		/* ~/.irssi doesn't exist, first time running irssi */
+		/* irssi config folder doesn't exist, first time running irssi */
```
ailin-nemui, line 284: according to the comment, the firsttimer depends on the config dir. but for xdg, we want the user to create the xdg-related folder manually. maybe we should change the check to look for the config file? or some other check? what do you think

src/core/core.c
```diff
@@ -192,24 +205,47 @@ void core_preinit(const char *path)
...
+			    g_build_filename(g_get_user_runtime_dir(), "irssi", NULL);
+		} else { /* fallback to non-xdg location */
+			home = g_get_home_dir();
+			if (home == NULL)
+				home = ".";
+			irssi_dir = g_strdup_printf(IRSSI_DIR_FULL, home);
+			/* all special XDG paths are equal to irssi_dir if XDG is not
+			 * supported */
+			irssi_cache_dir = irssi_dir;
+			irssi_runtime_dir = irssi_dir;
```
ailin-nemui, line 227: all these memory allocations are never freed at the tear-down of this module. this should be added and maybe it's easier to strdup them instead of special-caseing the frees

realaltffour: I assume that special-casing frees is meant to refer to `g_free()`
If so, `g_strdup_printf()` would still need to be freed the same way.
Can you elaborate what do you mean with strdup them?

ailin-nemui: here these are aliases, and must only be (g_)freed once. which does happen at the end of core.c. But the same cannot be said for the variables initialised with g_build_filename, which must be g_freed individually (currently missing)

granted this is a very minor memory leak as core_deinit is called just before ending  irssi only

src/fe-text/irssi.c
realaltffour: It would solve an edge case.

We only use XDG paths when the user explicitly creates `$XDG_CONFIG_HOME/irssi`. If we had XDG enabled in an irssi run, this check would simply fail. The config directory is already there (as acknowledged by the first statement).

If we are not using the XDG directories, we default to `~/.irssi`. There is a chance that this directory does not exist, if so, it's a firstimer.

If the config file location is specified in the commandline, this would mean that the user is not a firsttimer. It's for this situation where the user might not have `$XDG_CONFIG_HOME` nor `~/.irssi/` (correct me if this is not possible), the code assumes they are a firsttimer. This is not certainly bad, as it can be used to indicate that the config file is specified and a custom directory structure is being used. This also can be subtle and maybe even misleading.

Checking the config file rather than the directory would solve this edge case.

src/core/core.c
realaltffour: Thanks for the clarification!

> granted this is a very minor memory leak as core_deinit is called just before ending irssi only

Well, it is still a memory leak. I am going to fix this.

ailin-nemui:
sorry, I didn't write clearly

it does NOT matter if you g_strdup_printf, or g_build_filename, the end result must be freed here:

https://github.com/irssi/irssi/blob/master/src/core/core.c#L325

If you later on copy a string (`y = x;`) they point to the same memory, so you must only `g_free` one of them

if you copy the string like `y = g_strdup(x);`, then you have two copies and both must be `g_free`d

again, it does not matter if you use g_strdup_printf or g_build_filename

hope it's more clear now :-)

ailin-nemui:
@OJFord @josteink @keis @wsdjeg @LCRERGO now is your chance to test this awesome feature you've been requesting. Looking forward to your test results

josteink:
As the one who filed [the original feature request](https://github.com/irssi/irssi/issues/157), back in 2015, I do appreciate the effort put into this by all you contributors. I really do!

That said, I've largely moved to using IRC via Matrix and IRC-bridges, instead of using a dedicated IRC-client for that purpose. As such, I don't really have a live environment I can use to provide test-feedback with.

Sorry :grimacing:

keis:
Hey, thanks for working on this. I really wasn't expecting some action on this after all this time.

However it does not seem to work as (I) expected.

- I moved my config to ~/.config/irssi/config and started irssi
- Looks like no config was detect and I just got the default instead
- A empty ~/.irssi directory was created for some reason
- I tried loading my config with -c and irssi segfaulted

Relevant bits of my env
```
env | grep XDG                                                                                                                                                                                                  
XDG_CONFIG_HOME=/home/keis/.config
XDG_DATA_DIRS=/home/keis/.local/share/flatpak/exports/share:/var/lib/flatpak/exports/share:/usr/local/share:/usr/share
XDG_DATA_HOME=/home/keis/.local/share
XDG_DOWNLOAD_DIR=/home/keis/Downloads
XDG_RUNTIME_DIR=/run/user/1000
XDG_SEAT=seat0
XDG_SESSION_CLASS=user
XDG_SESSION_DESKTOP=sway
XDG_SESSION_ID=2
XDG_SESSION_TYPE=wayland
XDG_VTNR=2
```

keis:
Oh derp, -c is not shorthand for config. loading with `--config` does work. Segfaulting is not great though and FWIW it does not do so on `irssi 1.2.2 (20190829 0225)`

ailin-nemui:
@keis thank you very much for testing, I guess there are still some bugs

@realaltffour can you check these issues?

src/core/core.c
```diff
@@ -192,24 +211,49 @@ void core_preinit(const char *path)
...
 
 	if (irssi_dir == NULL) {
-		home = g_get_home_dir();
-		if (home == NULL)
-			home = ".";
-
-		irssi_dir = g_strdup_printf(IRSSI_DIR_FULL, home);
+		/* check if %XDG_CONFIG_HOME/irssi exists and use it */
+		char *dirp = g_build_filename(g_get_user_config_dir(), "irssi", NULL);
+		if (!g_file_test(dirp, G_FILE_TEST_EXISTS)) {
```
ailin-nemui, line 219: could it be that this condition is inverted?

ailin-nemui:
@keis wrt -c I believe you are running into #1239

src/core/core.c
realaltffour: That's funny. It is inverted.

src/core/settings.c
```diff
@@ -762,21 +762,31 @@ static CONFIG_REC *parse_configfile(const char *fname)
...
+		        dname, dname);
+	}
+}
+
+static void init_configfile(void)
+{
+	char *str;
+
+	init_ensure_dir(get_irssi_dir(), 0700); /* XDG_DATA_HOME or ~/.irssi*/
+	if (!strcmp(get_irssi_dir(), get_irssi_cache_dir()) &&
```
ailin-nemui, line 786: is this condition also mixed up ?

src/fe-common/core/themes.c
```diff
@@ -932,7 +932,7 @@ THEME_REC *theme_load(const char *setname)
 	theme = theme_find(name);
 
 	/* check home dir */
-	fname = g_strdup_printf("%s/%s.theme", get_irssi_dir(), name);
+	fname = g_build_filename(get_irssi_dir(), g_strdup_printf("%s.theme", name), NULL);
```
ailin-nemui, line 935: here is a memory leak, g_strdup_printf result is not assigned and freed

src/core/settings.c
realaltffour: No, we call `init_ensure_dir()` only when `get_irssi_dir()` is different to `get_irssi_cache_dir()` and `get_irssi_runtim_dir()`

ailin-nemui: .... but `!strcmp` means **equal**

maybe write `strcmp.... != 0`?

realaltffour: I'm currently having trouble finding time to get back on this issue. If someone would like to make a commit, it would be much appreciated. Sorry! :cry:

ailin-nemui: anyway thank you for the contribution I think you got pretty far

PR #1252, opened 2021-01-06
fix autotools build and package the meson.build files

ailin-nemui:
- Check Irssi compilation using github actions
- move clang-format-xs into utils
- remove travis


ailin-nemui:
@irssi/developers

dequis:
I don't get the "package the meson.build files" part of this

ailin-nemui:
we use `make dist` to create our release. If we want the next release to be buildable using meson and configure, then we have to ship all the meson.build files in that release tarball. to do this, we need to add the meson files to the EXTRA_DIST variable of automake

dequis (approved):

PR #1258, opened 2021-01-14
add a limit to showing /NAMES on join

ailin-nemui:
only show the counts if too many nicks

### DIFF
--- a/src/fe-common/core/fe-channels.c
+++ b/src/fe-common/core/fe-channels.c
@@ -630,7 +630,7 @@ void fe_channels_init(void)
 {
 	settings_add_bool("lookandfeel", "autoclose_windows", TRUE);
 	settings_add_bool("lookandfeel", "show_names_on_join", TRUE);
-	settings_add_int("lookandfeel", "show_names_on_join_limit", 18);
+	settings_add_int("lookandfeel", "show_names_on_join_limit", 0);
 	settings_add_int("lookandfeel", "names_max_columns", 6);
 	settings_add_int("lookandfeel", "names_max_width", 0);
 
--- a/src/fe-common/core/fe-channels.c
+++ b/src/fe-common/core/fe-channels.c
@@ -107,9 +107,14 @@ static void signal_window_item_changed(WINDOW_REC *window, WI_ITEM_REC *item)
 
 static void sig_channel_joined(CHANNEL_REC *channel)
 {
-	if (settings_get_bool("show_names_on_join") &&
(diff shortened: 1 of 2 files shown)


src/fe-common/core/fe-channels.c
```diff
@@ -625,6 +630,7 @@ void fe_channels_init(void)
 {
 	settings_add_bool("lookandfeel", "autoclose_windows", TRUE);
 	settings_add_bool("lookandfeel", "show_names_on_join", TRUE);
+	settings_add_int("lookandfeel", "show_names_on_join_limit", 18);
```
dequis, line 633: Yeah, nah, this won't do, I'm going to bikeshed the fuck out of this constant.

ailin-nemui: which constant would you prefer?

ailin-nemui: n.b this is 3 lines (3 × 6)

ailin-nemui:
@i/d

ailin-nemui:
@irssi/developers

src/fe-common/core/fe-channels.c
dequis: I guess I just disagree with the whole thing of "having too many nicks". If a channel is huge, the flood of names on join is useful information for me.

Default disabled?

ailin-nemui: I just think we should enable this. you still get the names count if the limit is exceeded and can easily type /n if you really need to see the list

but most of the time I'd argue the full list is useless on large channels

ailin-nemui: the full names list regularly hides the topic message and other more useful information

ailin-nemui: it's also a compromise to turning off show_names_on_join by default

dequis: I'm ok with the feature existing, but if people have a preference for a maximum number of nicks that are worth showing, they can set it in the setting. I don't like sneaky changes of default behavior, and I wouldn't want this changed in my own config.

ailin-nemui: I understand your point, but I think we should change the default settings somehow. The defaults represent what the user sees by default and I think we should make that better somehow.

I've opened #1276 to discuss this

ailin-nemui:
idea by dx: right before the count, add a message saying "number of users over show_names_on_join_limit, use /names to show full list"

PR #1259, opened 2021-01-15
do not beep on hidden lines with /set beep_msg_level ... -hidden

ailin-nemui:


ailin-nemui:
idea by xerox_

vague666:
Works as advertised if I understand the use case correctly, ie hidden lines trigger beep/activity hilight in terminal/wm and -HIDDEN prevents this. The question is should this have to be added? Shouldn't the original code already make sure activity is blocked

ailin-nemui:
should we change the defaults?
@irssi/developers

vague666:
If we go this route then yes

ailin-nemui:
by default beep_msg_level is empty so .. idk

src/fe-common/core/printtext.c
```diff
@@ -404,7 +405,18 @@ static void msg_beep_check(TEXT_DEST_REC *dest)
 	    (beep_when_away || (dest->server != NULL &&
 				!dest->server->usermode_away)) &&
 	    (beep_when_window_active || dest->window != active_win)) {
-                signal_emit("beep", 0);
+		if (beep_msg_level_ignore & MSGLEVEL_HIDDEN) {
```
vague666, line 408: We'd be able to skip this PR just by adding the check for MSGLEVEL_HIDDEN to line https://github.com/irssi/irssi/blob/80926476fc4c17dea9f24f9da753f7443b8c05e1/src/fe-common/core/printtext.c#L403
ie
`(dest->level & (MSGLEVEL_NO_ACT|MSGLEVEL_HIDDEN)) == 0`
if I remember my c correctly, is there any other use case for this PR?

ailin-nemui: we maybe don't want to abort the beep if the user has _not_ hidden the hidden lines (using /window hidelevel -hidden); and we may want to abort the beep if the user has hidden another line (e.g. using /window hidelevel +joins)

vague666: Hum hum, yeah

Issue #1267, opened 2021-02-01
header filename conflict in term.h


ailin-nemui:
The `src/fe-text/term.h` may conflict with `/usr/include/term.h` from XCurses.

~in my opinion the meson build should work, but @bsandro reported otherwise while investigating #1238~

The autotools build is currently prone to this issue, as it adds the current folder to `-I`. See #1268 for a fix

The easiest way would be to rename our term.h :/

ailin-nemui:
@dequis @irssi/developers any thoughts?

ailin-nemui:
this is when trying to `#include <term.h>` in order to use the right function declarations

mistydemeo:
Looks like this was fixed by #1268 - can confirm that fixed the conflict for me in #1290.

ailin-nemui:
fixed by #1268

PR #1268, opened 2021-02-01
configure automake with nostdinc

ailin-nemui:
the current directory won't be added to the compiler's -I path

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -5,7 +5,7 @@ AC_PREREQ(2.50)
 
 AC_CONFIG_HEADERS([irssi-config.h])
 AC_CONFIG_MACRO_DIR([m4])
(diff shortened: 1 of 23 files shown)


ailin-nemui:
this makes the build closer to meson. also it means `#include <term.h>` shouldn't include Irssi's own term.h in #1267

ailin-nemui:
@irssi/developers

dequis:
Did you check if this affects known modules? (Should probably include it in release notes as potentially breaking anyway)

ailin-nemui:
as it is only our own compilation, this should not be able to affect 3rd party modules in any way :-)

dequis (approved):
Oh okay I assumed it'd stop people from #including things in certain ways

PR #1269, opened 2021-02-05
fix invalid unsigned arithmetic.

ihsinme:
I believe your checks are not correct.
in my opinion they are equivalent to checks `!=`.
I suggest a simple fix.


ailin-nemui:
thanks for the report, would you be interested in following through and applying the style patch as well:

https://github.com/irssi/irssi/pull/1269/checks?check_run_id=1839878784#step:5:9

by the way did you or your automated checker find any other cases in the code? maybe there are more cases of signed -> unsigned

ihsinme:
> by the way did you or your automated checker find any other cases in the code? maybe there are more cases of signed -> unsigned

currently I see only this error. (

ailin-nemui:
@irssi/developers

PR #1270, opened 2021-02-08
correction of incorrect sequence of checks.

ihsinme:
I find it odd to first use `size` as a non-negative parameter, and then check for a value.

### DIFF
--- a/src/core/write-buffer.c
+++ b/src/core/write-buffer.c
@@ -63,13 +63,13 @@ int write_buffer(int handle, const void *data, int size)
         const char *cdata = data;
 	int next_size;
 
+	if (size <= 0)
+		return size;
+
 	if (write_buffer_max_blocks <= 0) {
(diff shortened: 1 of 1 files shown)


ailin-nemui:
@irssi/developers

dequis:
Sure. What motivates this?

ihsinme:
I'm sorry I didn't understand the question.

ailin-nemui:
they mean: why are you fixing these bugs for us?

ihsinme:
I am doing security analysis.
It is important for me to understand the representation of errors in real projects.
so I fix them, not only in your project.

and your project is open source, and it seems logical to me that anyone can suggest an improvement to you.)

ailin-nemui:
by the way, theoretically the behaviour could be different for the case size == 0 now (write with 0 does error check, if the buffering is turned off)

PR #1275, opened 2021-02-27
stop parsing on lone tag escape

ailin-nemui:
Credit to OSS-Fuzz


ailin-nemui:
@irssi/developers

dequis (approved):
Oh. Yep. Could have seen this one, but yeah i'm rusty and my mental model of for loops isn't quite right, it seems. Yay for oss-fuzz.

PR #1283, opened 2021-03-12
fail redirects on EFNet/freenode 263

ailin-nemui:


ailin-nemui:
if you redirect a /who and freenode cancels the who with 263, Irssi will use the response to the *next* who as that of the current one. This might also confuse some internal caches.

This code will send a redirect failure if a 263 is encountered, so I think it's better than before. The question is if we need even better handling (and if so, how to implement it?)

@irssi/developers

ailin-nemui:
to add to the insult, freenode **also** sends a 315, but it sends it __after__ the 263 and it sends it with argument `*` so Irssi has no way to match it against the _requested_ /who argument

ailin-nemui:
the same happens on EFNet, thanks @edk0

ailin-nemui:
@irssi/developers

PR #1288, opened 2021-03-17
add IRC_CHATNET_REC->usermode to perl chatnet hashes

jesopo:


jesopo:
the include reordering that clang wants breaks stuff >:O

ailin-nemui:
@irssi/developers

PR #1291, opened 2021-04-01
remove unused bool include

ailin-nemui:


ailin-nemui:
also fixes compilation on solaris

PR #1293, opened 2021-04-02
fix memory leak.

ihsinme:
It seems to me that you have a memory leak in this area.
however, it is easily accessible from outside, which makes it dangerous.
I suggest a simple fix.


ailin-nemui:
thanks

PR #1295, opened 2021-04-03
autoload modules from default setting

ailin-nemui:


ailin-nemui:
a replacement for the startup file when it comes to /load

ailin-nemui:
@irssi/developers

PR #1296, opened 2021-04-03
autoload the IRC module

ailin-nemui:


ailin-nemui:
a tracking issue

currently, Irssi creates dummy protocols which crash everything. The config loading is done into a dummy protocol (for example CHATNET_REC) which is later cast to the IRC_CHATNET_REC when the IRC protocol is loaded, but now of course these objects have the wrong memory

ailin-nemui:
Ticho> IIRC the exact problem I was was that my matrix server setup struct was created as an irc server setup struct, with all the irc module specific extra members set and initialized, so I was accessing wrong memory locations when I casted it to MATRIX_SERVER_SETUP_REC and reached inside

ailin-nemui:
merged into ailin-nemui/irssi#42

PR #1299, opened 2021-04-07
fail on empty text

ailin-nemui:
reported by @dwfreed


dwfreed:
lastlog reuses the same GString with every call of textbuffer_line2text, and depends on this function truncating the GString, otherwise when this particular edge case triggers, you'll end up with multiple copies of the same line printed by lastlog.  Either test after the truncation, or make lastlog do the truncation itself before calling textbuffer_line2text.

(Originally reported to me by @mnordhoff)

ailin-nemui:
moved it to after the truncation. the actual bug (reason unclear) is the line not having a text

ailin-nemui:
#1281

PR #1305, opened 2021-04-24
Fix compat with NetBSD libcurses by using tiparm() instead of tparm()

niacat:
Previously, irssi defined its own `tparm()` prototype, so passing a variable number of arguments worked.

However, since https://github.com/irssi/irssi/commit/0b82f14151dad5f106c208d25818816a8f782a54, <term.h> in included, which on systems using NetBSD's libcurses instead of ncurses, gives the X/Open compliant definition of tparm, which does not work with a variable number of arguments, resulting in a build error.

`tiparm()` is a variable-argument version of `tparm()` supported by both ncurses and NetBSD's libcurses. It was added to the X/Open Curses standard in Issue 7, the final version of which was released in November 2009.

from the ncurses man page:

>X/Open Curses  prototypes tparm with a fixed number of parameters,
>rather than a variable argument list.
>
>This implementation uses a variable argument list, but can be  con-
>figured  to  use  the  fixed-parameter list.  Portable applications
>should provide 9 parameters after the format; zeroes are  fine  for
>this purpose.
>
>In  response  to review comments by Thomas E. Dickey, X/Open Curses
>Issue 7 proposed the tiparm function in mid-2009.

### DIFF
--- a/src/fe-text/terminfo-core.c
+++ b/src/fe-text/terminfo-core.c
@@ -17,7 +17,7 @@ inline static int term_putchar(int c)
 #else
 /* Don't bother including curses.h because of these -
    they might not even be defined there */
-char *tparm();
+char *tiparm();
 int tputs();
 
 int setupterm();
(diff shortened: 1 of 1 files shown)


ailin-nemui:
that's what you get from trying to fix apple :(

ailin-nemui:
nb we don't have tiparm on Solaris 5.10

niacat:
@ailin-nemui does irssi compile using Solaris curses? i'd expect it not to because their tparm is also fixed-argument, but I guess that can be "fixed" by not including term.h on __sun.

ailin-nemui:
maybe we have to revert the apple patch and go back to not including term.h

niacat:
that's fine by me.

ailin-nemui:
@mistydemeo any ideas?

niacat:
FWIW my first version of this patch just added the missing arguments to tparm(), but it does add a lot of noise to the code.

ailin-nemui:
for the time being I have attached a [no-term.h.patch](https://github.com/irssi/irssi/releases/download/1.2.3/no-term.h.patch) to the releases page. sorry for the inconvenience

mistydemeo:
I wonder if we can add a check for `tiparm` and set a macro which decides whether to use `tparm` or `tiparm` based on that?

> FWIW my first version of this patch just added the missing arguments to tparm(), but it does add a lot of noise to the code.

How ugly is it? Sounds like it might actually be a good solution.

ailin-nemui:
@irssi/developers

PR #1308, opened 2021-04-26
Fix build on operating systems with X/Open Curses, version 2

niacat:
Add the missing arguments to tparm. X/Open Curses specifies tparm takes a fixed number of 10 arguments, while ncurses has implemented it as a varargs function. tiparm is a standardized version of varargs tparm, support in both NetBSD libcurses and ncurses, but not by older versions of Solaris.

This is an alternate fix to the one proposed in #1305 that should keep compatibility with older versions of Solaris by avoiding tiparm.


ailin-nemui:
@irssi/developers

dequis (approved):

PR #1314, opened 2021-05-10
fix build with meson 0.58.0

ailin-nemui:


ailin-nemui:
See https://github.com/mesonbuild/meson/issues/8700

thanks @eli-schwartz

PR #1315, opened 2021-05-15
Use g_strndup, strndup is not available on old platforms

pekdon:
strndup is not available on Solaris 10, to ensure building is ok use
glib function.


ailin-nemui:
@irssi/developers

dequis (approved):

PR #1316, opened 2021-05-20
Add aliases for ns, ms, os, cs shorthand commands supported by many servers to message services easily

FeepingCreature:
Right now, you have to either do `/quote cs ...` or `/msg chanserv ...`, raising potential for leaking information via typos.

Tested on Esper, Freenode and Libera.


src/irc/core/irc-commands.c
```diff
@@ -1045,6 +1045,11 @@ void irc_commands_init(void)
 	command_bind_irc("knockout", NULL, (SIGNAL_FUNC) cmd_knockout);
 	command_bind_irc("setname", NULL, (SIGNAL_FUNC) command_1self);
 	command_bind_irc("server purge", NULL, (SIGNAL_FUNC) cmd_server_purge);
+	/* common shorthands for MSG NickServ, ChanServ, OperServ, MemoServ :parameter */
+	command_bind_irc("ns", NULL, (SIGNAL_FUNC) command_self);
+	command_bind_irc("cs", NULL, (SIGNAL_FUNC) command_self);
+	command_bind_irc("os", NULL, (SIGNAL_FUNC) command_self);
+	command_bind_irc("ms", NULL, (SIGNAL_FUNC) command_self);
```
vague666, line 1052: You need to unbind the commands too, but I'm wondering if it wouldn't be better to add those as default aliases?

vague666: Place to add the aliases: https://github.com/irssi/irssi/blob/master/irssi.conf
But lets let the maindevs have the final say

FeepingCreature: Added unbind calls for now, until someone tells me how to add default aliases. :)

ailin-nemui:
Thanks for this pull request, however I believe that the Irssi project prefers to do these commands yourself. Every user can set up their own aliases for example `/alias ns quote ns`
@irssi/developers ?

vague666:
I don't mind adding default aliases, those commands are often referenced in support articles so it would be good if irssi knew what to do with them without the user having to ask for help when stuff fail

FeepingCreature:
Switched to aliases.

dequis:
I'm also ok with these as aliases

PR #1322, opened 2021-05-26
Fix cursor getting stuck for auto completions that changes case

### LINKED ISSUE #1176: cursor not advancing after auto-completion
Reported by Kryten

| | |
--- | ---
kryten  |  Also, I may have mentioned this before, and wonder if anybody else can reproduce this: 1.) Type a ":D", 2.) try putting more than one space after it, 3.) whitness how the cursor doesn't advance further than one space, while any additional ones have indeed been added after it, so you can press End and jump at the end of all those spaces.
vague  | Sounds like utf8 problems combined with /completion -auto ?
vague  | But there might be a bug in the completion code too, the cursor should advance normally after completion
kryten | Lemme check my completion list..
kryten | Wow, I indeed had an auto-completion of ':d' to ':D' there, but didn't think that'd affect it.. O_o
kryten |Yeah, and it does so on other completions, but also I'd think the lowercase should matter - since that's exactly the reason why I added it to begin with. >_<
kryten | Irssi v1.3dev-245-g8b7bf1f0 (20200403)

ffrogman:
Fixes #1176

So I realized autocomplete code is case insensitive - I did not change that behavior. For example if you only define the following autocomplete: ` abccd -> ABCCD`. `abccd`, `aBccD`, `ABcCD` all will convert to `ABCCD`. This also means `ABCCD` would convert to the identical string `ABCCD`. This is the source of the bug

This bug was reproducible for any complete that exclusively changes the case of letters and no other character changes; `:d` -> `:D`, `Ffrogman` -> `ffrogman`, `FfroGmAn` -> `ffrogman`.. 

The autocomplete would return a match for these and the replacement would happen every time space was pressed - even after the completion was already made. After an auto completion the cursor jumps to one character after the completion. So every space would trigger another completion and reset the cursor even though there would be no change.

Fix is to stop trying to replace these identical strings if they are already. There's no reason to do the work of swapping out identical strings.

### DIFF
--- a/src/fe-common/core/completion.c
+++ b/src/fe-common/core/completion.c
@@ -104,7 +104,7 @@ char *auto_word_complete(const char *line, int *pos)
 
 	/* check for words in autocompletion list */
 	replace = completion_find(word, TRUE);
-	if (replace == NULL) {
+	if (replace == NULL || (!g_strcmp0(replace, word))) {
 		ret = NULL;
 		g_string_free(result, TRUE);
 	} else {


ailin-nemui:
merged into ailin-nemui/irssi#8

PR #1337, opened 2021-06-23
Allow tag/* in activity_hide_targets

vague666:


ailin-nemui:
merged into ailin-nemui/irssi#11

PR #1342, opened 2021-08-27
Implement the UTF8ONLY IRCv3 specification

progval:
https://ircv3.net/specs/extensions/utf8-only


progval:
ping?

ailin-nemui:
sorry, this is currently out of scope. thanks again for your efforts and apologies for the bad news

progval:
Sorry but I don't understand. The purpose of this is to make sure messages sent by irssi are not rejected by servers like Ergo. Why is it out of scope?

PR #1368, opened 2022-03-08
Clarify the help for /ignore -time

vague666:


ailin-nemui:
maybe `<time>` instead of `<str>`? compare https://irssi.org/documentation/help/knockout/

vague666:
Does knockout accept time units or is seconds the only valid unit?

ailin-nemui:
iirc it uses the same "time value" parsing function

ailin-nemui:
they surely can't be caused by fixing the docs ;)

ailin-nemui:
merged into ailin-nemui/irssi#77

PR #1369, opened 2022-03-22
Irssi nei patches

tomaw:
Let's merge this back in.


vague666:
We need to just press the merge button I think :)

PR #1371, opened 2022-04-01
Format RPL_HELPSTART/RPL_HELPTXT/RPL_ENDOFHELP

progval:
They are typical replies when using the HELP command.

Tested on Ergo.

Before:
screenshot-2022-04-01_15-41-13

After:
screenshot-2022-04-01_15-49-08


ailin-nemui:
524 is already used to notify about Quarantined channels, it would not be good to bind it as "help not found"

progval:
Yes, but ERR_HELPNOTFOUND is more widespread:

* 524 is ERR_HELPNOTFOUND on irc-ratbox and derivatives (Hybrid, Solanum, Plexus, ...), Inspircd, and Ergo. This covers 6 of the top 10 networks (according to netsplit.de).
* UnrealIRCd may also start using it as ERR_HELPNOTFOUND in a few years depending on client support.
* 524 is ERR_QUARANTINED on ircu2 derivatives (Nefarious, snircd, ...) except UnrealIRCd. This covers 2 of the top 10.

What do you usually do in this kind of situation?

ailin-nemui:
we would need to detect which one it is, probably by using a strcmp and see if the response contains :Help, so that we can display the right one. quarantined channels also contain the quarantine reason, wouldn't want to hide that for sure

progval:
Hmm that might fail because `:Help` is not guaranteed to be a substring of the ERR_HELPNOTFOUND message.

However, we could check if the second parameter starts with one of the ISUPPORT `PREFIX` character; it's unlikely to be a valid topic name. Agreed?

ailin-nemui:
sadly that's not so cool either, if we do /shelp #xxx then it would trigger the quarantine code path??

maybe just leave the 624 unhandled (sad but less work)?

progval:
Ah, sad. I'll remove it, then.

src/fe-common/irc/module-formats.c
```diff
@@ -89,6 +89,9 @@ FORMAT_REC fecommon_irc_formats[] = {
 	{ "invitelist_long", "{channel $0}: invite {ban $1} {comment by {nick $2}, $3 secs ago}", 4, { 0, 0, 0, 1 } },
 	{ "no_such_channel", "{channel $0}: No such channel", 1, { 0 } },
 	{ "channel_synced", "Join to {channel $0} was synced in {hilight $1} secs", 2, { 0, 2 } },
+	{ "irctxt_help_start", "$1", 2, { 0, 0 } },
+	{ "irctxt_help_txt", "$1", 2, { 0, 0 } },
+	{ "irctxt_end_of_help", "$1", 2, { 0, 0 } },
```
ailin-nemui, line 94: no need for irctxt_ prefix here, only for the constants. although I think I would rename it to server_help_start /IRCTXT_SERVER_HELP_START so it won't be confused with client-internal help

ailin-nemui:
thanks for the improvement, I copied this to ailin-nemui#82

PR #1377, opened 2022-06-11
fix yet another meson regression

ailin-nemui:
broken by meson 0.60.0

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -58,6 +58,9 @@ if fs.exists('config.status') or fs.exists('irssi-version.h') or fs.exists('defa
   error('this tree has been configured with autotools, cannot proceed')
 endif
 
+UNSET = '=INVALID='
+UNSET_ARR = [UNSET]
+
 chat_modules = ['irc']
 
 run_command('mkdir', meson.current_build_dir() / incdir, check : false)
@@ -334,8 +337,8 @@ if want_perl
   perl_rpath = ''
 
   #### ccopts ####
-  perl_ccopts = meson.get_cross_property('perl_ccopts', false)
-  if perl_ccopts == false
+  perl_ccopts = meson.get_cross_property('perl_ccopts', UNSET_ARR)
+  if perl_ccopts == UNSET_ARR
     res = run_command(cross_perl, '-MExtUtils::Embed', '-e', 'ccopts', check : true)
     perl_ccopts = res.stdout().strip().split()
   endif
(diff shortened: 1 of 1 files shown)


eli-schwartz:
For context, the issue that this PR fixes only happens when meson.get_cross_property() finds an existing value.

In all such cases, versions of Meson before 0.60.0 would print the following warning:
```
meson.build:338: WARNING: Trying to compare values of different types (list, bool) using ==.
The result of this is undefined and will become a hard error in a future Meson release.
Compiler for C supports arguments -fPIC: YES 
meson.build:352: WARNING: Trying to compare values of different types (list, bool) using ==.
The result of this is undefined and will become a hard error in a future Meson release.
```

Tested with the following cross file:
```ini
[binaries]
c = 'gcc'
cpp = 'g++'
pkgconfig = 'pkg-config'

[properties]
perl_ccopts = ['-D_REENTRANT', '-D_GNU_SOURCE', '-fwrapv', '-fno-strict-aliasing', '-pipe', '-fstack-protector-strong', '-I/usr/local/include', '-D_LARGEFILE_SOURCE', '-D_FILE_OFFSET_BITS=64', '-D_FORTIFY_SOURCE=2', '-I/usr/lib/perl5/5.34/core_perl/CORE']
perl_ldopts = ['-Wl,-E', '-Wl,-rpath,/usr/lib/perl5/5.34/core_perl/CORE', '-Wl,-O1,--sort-common,--as-needed,-z,relro,-z,now', '-fstack-protector-strong', '-L/usr/local/lib', '-L/usr/lib/perl5/5.34/core_perl/CORE', '-lperl', '-lpthread', '-ldl', '-lm', '-lcrypt', '-lutil', '-lc']
```

Note: the warning has been there since 0.44.0, and as I pointed out in IRC:
- the intended goal of this sentinel value was to check if a cross property is *defined*, but it was wrong if someone decided to define it in their cross file as `perl_ccopts = false`
- Meson 0.58 deprecates get_cross_property and replaces it with get_external_property... which comes with a has_external_property that does what you *actually* want.

Rather than bumping the minimum version of Meson to 0.58, it may be practical to wrap this in meson.version().version_compare() to use the new mechanism (the old mechanism can use whichever sentinel value it likes, since `false` will just be a Meson warning).

eli-schwartz:
Also relevant :(
https://github.com/Perl/perl5/issues/14030

PR #1403, opened 2022-07-19
handle SIGTERM in fe-text

lnde:
Exit gracefully on SIGTERM in fe-text by emitting "gui exit" and
quitting.


src/fe-text/irssi.c
```diff
@@ -331,6 +342,11 @@ int main(int argc, char **argv)
 	/* Does the same as g_main_run(main_loop), except we
 	   can call our dirty-checker after each iteration */
 	while (!quitting) {
+		if (sigterm_received) {
+			signal_emit("gui exit", 0);
+			quitting = TRUE;
```
dwfreed, line 347: This is redundant (which is why the `quit_on_hup` path doesn't do it)

lnde: Thanks for the feedback and this has been addressed in the new proposal.

lnde:
> @lnde could you squash your commits?

Sure thing, and I've now done that.

PR #1410, opened 2022-09-09
Restore program to working state

jonsykkel:


jonsykkel:
this hack breaks previously working setups and i have to spend 1hour of my life remotely debuging why my budy cant conect to a particular server
if ppl want the weird tls stuff they can enable it explicitly

vague666:
@jonsykkel I've updated the help for /server and /connect with -cap/-nocap details
Please check if #1411 or #1415 helps

jonsykkel:
my server dose not suport irc3 or starttls or anything like that and dose not respond to weird comands sent before USER
is this compliant with rfc482929abcef? who knows
the point is that this "JOIN :" thing is a hack and breaks working setups (it breaks my setup, which was workign, and now is not working) and the only thing that can fix this problem is a patch reverting the hack
i would also like to see hand written personal e-apologys extended to everyone who installed irsi1.4.1 and had their setup broken or was negatively affected by somone else installing irsi1.4.1
now of course it is inpossible to know in practice how many or who these ppl are, which means they will never receive their e-apology, which is unfortunate

jonsykkel:
that is wat i did, after debuging why concetion was not working then diging throuhg release notes and randomly discoverign this option
my conection now works, the problem is not that my conection is not wokring, the problem is that software that was working stopped working and has not resumed working

PR #1425, opened 2022-11-21
Add -comment attribute to hilights & ignores

### LINKED ISSUE #1423: RFE: comment attribute for hilight tracking
Turns out that I have hilights.  A lot of hilights.  I use them to keep track of problem users and problematic speech.  Unfortunately, my memory not being what it once was, I sometimes lose track of what I added specific hilight patterns for.

My thought was a -comment attribute that could be set when a hilight was added; this would be a free form text field that could be used as a description of the user/behavior which necessitated the hilight.  Could also use it for date specs to know how timely / pertinent a specific hilight may be.

gerdesas:
Add -comment attribute to hilights to assist those with many hilights in tracking what they are used for.  Useful for channel ops or network staff that often have many hilights set.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -3352,7 +3352,7 @@ v0.7.1 1999-03-08  Timo Sirainen <a@sicom.fi> [unstable]
 
           You can use these levels, ALL is set by default:
           ALL, CRAP, PUBLIC, MSGS, NOTICES, WALLOPS, SNOTES, ACTIONS, DCC,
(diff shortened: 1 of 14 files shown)


ailin-nemui:
- added comments to /ignore

- the comments can be seen in `/hilight` and `/ignore`

- also support editing ignores with `/ignore ... <id> [<levels>]`

PR #1432, opened 2022-12-18
Check window logfile parameter

horgh:


horgh:
I'm trying to figure out the clang-format issues!

PR #1436, opened 2022-12-20
Fix generate tag from address (servers.c)

baytuch:
This commit fixed the generation of the tag from the address in such a way as to contain a name. For example: IRC.it-hobby.km.ua-> it-hobby.


baytuch:
My project is hosted in a second level domain so I would like the tag to be the project name and not the second level domain.

baytuch:
I tried to fix broken mechanism.

ailin-nemui:
sorry, we cannot change the automatic tag generation by default. please add a network to configure the tag you want

you can also manually specify it like so: `/connect -network it-hobby irc.it-hobby.km.ua`

PR #1442, opened 2023-01-28
Support receiving monospace

lheckemann:
0x11 (decimal 17) is used to denote monospace by a number of clients according to https://modern.ircdocs.horse/formatting.html


MichaelJTokarev (approved):
I just verified the change. The annoying "Inverse-Q" markup at the start of code-17 words disappeared after this change.

PR #1456, opened 2023-03-26
fix stale special collector use after free

ailin-nemui:
reported by ednash and investigated by @dwfreed

exposed by GLib update in Arch Linux https://gitlab.gnome.org/GNOME/glib/-/issues/2937


dwfreed:
This doesn't fix the issue (as just proven on IRC), but it doesn't hurt, either.

dwfreed (approved):
After further testing, I can reliably reproduce the issue without this change, and cannot with it, so this seemingly is the fix.  However, the explanation is not known.

PR #1457, opened 2023-03-26
properly format listmodes and their timestamps

examknow:
This calls `my_asctime()` to display a timestamp for listmodes that is actually meaningful to humans. It also provides proper handling for quiet list entries.


dwfreed:
Note that at present, OFTC uses 344/345 instead of 728/729 for quiet list; you may want to handle both

src/fe-common/irc/fe-events-numeric.c
```diff
@@ -138,20 +138,55 @@ static void event_end_of_who(IRC_SERVER_REC *server, const char *data)
...
+	seconds %= (86400*365);
+	weeks = seconds/604800;
+	seconds %= 604800;
+	days = seconds/86400;
+	seconds %= 86400;
+	hours = seconds/3600;
+	hours %= 3600;
+	minutes = seconds/60;
+	minutes %= 60;
+	seconds %= 60;
```
dwfreed, line 159:
```suggestion
	years = seconds / (86400 * 365);
	seconds %= (86400 * 365);
	weeks = seconds / 604800;
	days = (seconds / 86400) % 7;
	hours = (seconds / 3600) % 24;
	minutes = (seconds / 60) % 60;
	seconds %= 60;
```

PR #1485, opened 2023-08-14
build: fix linking error on arm64 Darwin

emilengler:
Currently, it is not possible to build git master on Darwin, as seen in several issues reported previously.

This commit fixes that issues, by initializing two variables to NULL. I have no idea, why this solves the linking error, as to my knowledge, global variables do not need to be initialized.

### DIFF
--- a/src/fe-common/core/themes.c
+++ b/src/fe-common/core/themes.c
@@ -34,8 +34,8 @@
 #include "default-theme.h"
 
 GSList *themes;
-THEME_REC *current_theme;
-GHashTable *default_formats;
+THEME_REC *current_theme = NULL;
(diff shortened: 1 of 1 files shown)


emilengler:
Hmmm. With this patch applied, irssi won't really initialize anymore. I am not sure if this is because of this patch though, or if the macOS version is simply broken.

emilengler:
I do not think that this bug comes from this patch, as applying it to 1.4.4 still works fine.

emilengler:
Sorry for being a bit inactive here. It seems like the issues are solved on the newest macOS release though (or maybe someone else fixed it here).

PR #1486, opened 2023-08-14
patch by petteri_ to enable bash-like editing of window history

vague666:


vague666:
Some comments by petteri_
```
(aug 14 11:03:14)      petteri_ │ vague, the patch may need more work for upstream. This works for me, I'll leave it at that.
(aug 14 11:03:17)      petteri_ │ For instance, I don't know how this interferes with HISTORY_ENTRY_REC.time that is used in command_history_{load,delete}_entry. There is also repetition in command_history_{next,prev}_int.
(aug 14 11:03:21)      petteri_ │ One more point, although the present behavior is consistent, it is at odds with many programs that use GNU Readline. Irssi already shares some key bindings with Readline.
```

PR #1489, opened 2023-08-21
Correct message level documentation

soliton-:
I'll also check level usage in examples.

In https://github.com/irssi/irssi/blob/07b971cf0c5a6607f9071f332ae27adb3bb4baa1/src/core/levels.c#L25-L49 there is also PUBNOTICES listed which is not mentioned. Not sure if that should also be added to the help.

### DIFF
--- a/docs/help/in/levels.in
+++ b/docs/help/in/levels.in
@@ -22,7 +22,7 @@
         NICKS           A nickname changes to another nickname.
         NOTICES         Private notices.
         PARTS           A nickname leaves a channel.
-        PUBLIC          Public messages in a channel.
+        PUBLICS         Public messages in a channel.
         PUBNOTICES      Public notices in a channel.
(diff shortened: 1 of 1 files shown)


soliton-:
Any unambiguous prefix works but I would expect the complete form in documentation.

soliton-:
https://github.com/irssi/irssi/blob/07b971cf0c5a6607f9071f332ae27adb3bb4baa1/docs/help/in/levels.in#L41-L42

If this means we need to pretend the two lists are the same or that this primarily describes the scripts interface then there should be some hint about that or that will keep tripping people up IMO.

soliton-:
Up to you of course what you want to commit but I would at least be consistent in the examples. It's confusing to see PUBLIC and PUBLICS in `/help ignore`.

ailin-nemui:
I believe this is now fixed with #1425

PR #1516, opened 2024-03-31
Add hidelevel changed signal

vague666:


vague666:
Does it matter where it's sent from? I was hoping to send a WINDOW_REC but had real trouble getting the sbar_item script to understand what came with the signal, rather, what was received was undef, no matter what I tried

vague666:
That's unfortunate. Maybe it can be automated somehow
Updates coming :)

PR #1519, opened 2024-04-01
deprecated openssl3 function

ailin-nemui:


src/core/network-openssl.c
```diff
@@ -754,41 +755,49 @@ static void set_server_temporary_key_info(TLS_REC *tls, SSL *ssl)
 
 	/* OPENSSL_NO_EC is for solaris 11.3 (2016), github ticket #598 */
 #ifndef OPENSSL_NO_EC
-	EC_KEY *ec_key = NULL;
 #endif
```
dwfreed, line 758: Remove empty `#ifndef`

dwfreed (changes requested):
Looks right to me, except for the one nit.

PR #1525, opened 2024-04-12
fix /ban command

ailin-nemui:
broken by https://github.com/irssi/irssi/pull/1457

reported by laama on IRC


laama1:
the fix is working where the affected version segfaulted when listing bans on certain channels or networks. but does not crash anymore after the fix

Issue #1531, opened 2024-07-14
fails to connect to an INSPIRCD 4.0.1. server 


fezie:
Irssi can't connect to an INSPIRCD 4.0.1 server:

```

19:19 -!- Irssi: Connecting to localhost [127.0.0.1] port 6667
19:19 Waiting for CAP LS response...
19:19 -!- Irssi: Connection to localhost established
19:19 !irc.example.com *** Looking up your hostname...
19:19 -!- JOIN Not enough parameters.
19:19 !irc.example.com *** Found your hostname (localhost)
19:19 -!- ERROR Closing link: (671AAAAAC@localhost) [Connection timeout]
19:19 -!- Irssi: Connection lost to localhost
```

rawlog shows:

```
<< CAP LS 302
<< JOIN :
>> :irc.example.com NOTICE * :*** Looking up your hostname...
>> :irc.example.com 461 * JOIN :Not enough parameters.
>> :irc.example.com NOTICE * :*** Found your hostname (localhost)

```
irc-servers.c has this:

```
        if (!conn->no_cap) {
                signal_emit("server waiting cap ls", 2, server, CAP_LS_VERSION);
                irc_send_cmd_now(server, "CAP LS " CAP_LS_VERSION);
                /* to detect non-CAP servers, send this bogus join */
                /* the : will make INSPIRCD respond with 451 instead of 461, too */
                irc_send_cmd_now(server, "JOIN :");
        }

```

But as you can see above it does return 461 and not 451

SadieCat:
If you really need to do this hack you should probably use something like `PING` that can't be used before connection registration rather than relying on undefined behaviour with malformed trailing parameters.

SadieCat:
I'm likely to revert my workaround in the next dev branch so please consider fixing this even though I added a workaround for this.

PR #1534, opened 2024-08-31
Fix non-CAP InspIRCd 4 servers not working

MarioMario456:
The InspIRCd quirk (bug?) Irssi uses to detect non-CAP servers no longer works with InspIRCd 4, resulting in non-CAP servers refusing to connect. This pull request fixes [issue 1531](https://github.com/irssi/irssi/issues/1531).

### DIFF
--- a/src/irc/core/irc-servers.c
+++ b/src/irc/core/irc-servers.c
@@ -313,9 +313,9 @@ static void server_init_1(IRC_SERVER_REC *server)
 	if (!conn->no_cap) {
 		signal_emit("server waiting cap ls", 2, server, CAP_LS_VERSION);
 		irc_send_cmd_now(server, "CAP LS " CAP_LS_VERSION);
-		/* to detect non-CAP servers, send this bogus join */
-		/* the : will make INSPIRCD respond with 451 instead of 461, too */
-		irc_send_cmd_now(server, "JOIN :");
+
+		if (!server->cap_supported)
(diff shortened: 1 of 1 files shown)


src/irc/core/irc-servers.c
```diff
@@ -313,9 +313,9 @@ static void server_init_1(IRC_SERVER_REC *server)
 	if (!conn->no_cap) {
 		signal_emit("server waiting cap ls", 2, server, CAP_LS_VERSION);
 		irc_send_cmd_now(server, "CAP LS " CAP_LS_VERSION);
-		/* to detect non-CAP servers, send this bogus join */
-		/* the : will make INSPIRCD respond with 451 instead of 461, too */
-		irc_send_cmd_now(server, "JOIN :");
+
+		if (!server->cap_supported)
```
dwfreed, line 317: `server->cap_supported` will always be `NULL` at this point, this is entirely wrong.

dwfreed (changes requested):
This fix is entirely wrong. Also, please use proper commit messages.

src/irc/core/irc-servers.c
MarioMario456: Tried to fix it. Sorry if it's still "not quite right"; on my system and setup it "just works".

https://github.com/MarioMario456/irssi/commit/d7e1bdba8202a0f5f5259152a05c1ca61042cb63

PR #1536, opened 2024-09-22
core/chatnets : Make object local to the object file

ReDressCodes:
This object is not exposed via extern and therefore isn't accessible to other source files except through another extern which, even then conflict with the coding style in the rest of irssi.

 This makes it safe to assume that it should be local to this file.

There probably is an isolated occurrence and even if other such issues exist in the rest of the code, it would be great to get some comments on this one first.


ReDressCodes:
This seems to be talking too long. Closing it

PR #1539, opened 2024-10-10
[WIP] Add server_redirect for NAMES

### LINKED ISSUE #1537: irssiproxy: hide NAMES from client and replies to client
irssiproxy seems to have code for various other commands, for example WHO. TheLounge (which I connect to irssiproxy for having IRC on the phone) polls NAMES instead of WHO, leaving random output in all windows.

This output already is displayed differently from NAMES requests sent by the user. Since irssiproxy contains such code for other commands, it would only be consistent to fix it there rather than by script.

wilhelmy:
Then, use this redirect to make irssi not print NAMES commands sent by proxy clients.

Closes https://github.com/irssi/irssi/issues/1537


wilhelmy:
Apparently this patch doesn't work on hackint, which sends timestamped messages such as 
`@time=2024-10-10T15:50:57.611Z :vindobona.hackint.org 353 foo @ #bar :...`
But it works on all other networks for hiding these /NAMES replies. Maybe someone here is more familiar with irssi's codebase than I am.

wilhelmy:
Hmm, after more prolonged testing with all of my regular networks connected, IRCnet and OFTC also still sometimes get /NAMES replies. I guess this fix is incomplete. Does anyone here know about this redirections mechanism? I can do my own digging of course but I'm not sure when.

wilhelmy:
I made a mistake in my original patch handling the `NAMES` numeric reply, which has a different format than other numeric replies, with the channel in the second parameter of the reply. I'm running this newly patched irssi and the bug no longer appears.

You can use irssiproxy and connect TheLounge as the proxy client to reproduce all this.

wilhelmy:
I guess I should clarify a bit more what the exact problem with my previous patch was. To quote from RFC 2812:

```
       353    RPL_NAMREPLY
              "( "=" / "*" / "@" ) <channel>
               :[ "@" / "+" ] <nick> *( " " [ "@" / "+" ] <nick> )
         - "@" is used for secret channels, "*" for private
           channels, and "=" for others (public channels).
```

`argpos=2` is the correct position of the `<channel>` that skips this flag. `argpos=1` is what most other redirects use. RFC 1459 in the section that discusses `RPL_NAMREPLY` makes no mention of this, however I've seen this `=` flag in all telnet sessions with non-IRCnet ircd implementations as well... guess it's an oversight in the original RFC.

wilhelmy:
Seems I still didn't completely fix it. There are still spurious `NAMES` responses in my irssi windows, although more rarely. Mostly IRCnet is affected, and usually it happens right after `MODE` changes but I'm suspecting that this is just because that triggers TheLounge to request `NAMES`. I need to investigate whether it has something to do with the `!<prefix>` part of `!channel`s, such as qualifying `!channel` to its real collisionproof name `!f8fdzchannel` and back. Marking as WIP again, sorry for the noise.

PR #1540, opened 2024-11-01
Include <sys/select.h> to get select(3).

sortie:
This is the header required by POSIX-1.2024.

This change is needed to build irssi on operating systems such as Sortix that have a strict libc without obsolete behaviors.


sortie:
Hi, what's the status on this PR?

PR #1541, opened 2024-11-01
Replace obsolescent inet_addr(3)/inet_aton(3) with inet_pton(3).

sortie:
inet_addr has become obsolescent as of POSIX-1.2024 and is not available on strict POSIX 2024 libc implementations.

inet_pton(3) is the standard and portable replacement available on all POSIX-1.2001 systems.

--

I'm unclear if you want to accept this change as-is depending on what platforms irssi supports. inet_pton should be on every relevant Unix today since it was added to POSIX 23 years ago. I need this little change or something like it to compile irssi with a strict modern libc. :)

### DIFF
--- a/src/core/network.c
+++ b/src/core/network.c
@@ -491,8 +491,6 @@ int net_ip2host(IPADDR *ip, char *host)
 
 int net_host2ip(const char *host, IPADDR *ip)
 {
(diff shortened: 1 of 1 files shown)


sortie:
Hi, what's the status on this PR?

PR #1570, opened 2025-06-29
feat: add support for JOIN 0 to leave all channels

cseriildi:
This PR adds support for the `JOIN 0` command, to leave all currently joined channels as defined in [RFC 2812 §3.2.1](https://www.rfc-editor.org/rfc/rfc2812.html#section-3.2.1).
- Prevents automatic `#` prefixing for 0 when no other parameter is given. 
- Updates `/help join` with this behavior (syntax, description, and example).

### DIFF
--- a/docs/help/in/join.in
+++ b/docs/help/in/join.in
@@ -13,7 +13,7 @@
 
 %9Description:%9
 
-    Joins the given channels.
+    Joins the given channels, or use `JOIN 0` to leave all currently joined channels.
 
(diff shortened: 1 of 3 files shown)


Steering7253:
In my experience the only time anyone uses JOIN 0 is when someone trolls them into it.

cseriildi:
Appreciate the feedback! 🙏 I hadn’t noticed the existing documentation — since this is intentional, I’m closing the PR.

PR #1578, opened 2025-07-25
make compilation work on cygwin

ailin-nemui:


RealKindOne:
Perl is not being compiled into irssi?

https://gist.github.com/RealKindOne/1216b44908bba5b39a7def3da3e092ac

RealKindOne:
Applied patch and rebuilt/installed `glib2.0-2.85.2-1` package, rebuilt/installed irssi git. No perl.

PR #1580, opened 2025-07-27
use GIO resolver

ailin-nemui:


src/core/network.c
```diff
@@ -392,95 +392,68 @@ int net_getsockname(GIOChannel *handle, IPADDR *addr, int *port)
...
-	}
-	freeaddrinfo(ailist);
-	return 0;
+	return iprec;
 }
 
-/* Get name for host, *name should be g_free()'d unless it's NULL.
-   Return values are the same as with net_gethostbyname() */
-int net_gethostbyaddr(IPADDR *ip, char **name)
+int net_gethostbyname_first_ips(const char *addr, GResolverNameLookupFlags flags, IPADDR *ip4,
```
thiagomacieira, line 425: This function is throwing away important information:
1. the relative order between IPv4 and IPv6 (which one comes first)
2. the second and further addresses of each family.

*Most* DNS resolvers will change the order of the results if you look up the same thing again, but that's not required. So by throwing away the second and later results, we could end up stuck trying to connect to a server that fails to respond. This is a pre-existing problem.

src/core/servers.c
```diff
@@ -276,48 +272,37 @@ static void server_real_connect(SERVER_REC *server, IPADDR *ip,
...
-	g_io_channel_unref(server->connect_pipe[1]);
-
-	server->connect_pipe[0] = NULL;
-	server->connect_pipe[1] = NULL;
+	server->connect_cancellable = NULL;
 
 	/* figure out if we should use IPv4 or v6 address */
-	if (iprec.error != 0) {
-                /* error */
+	if (iprec->error != NULL) {
```
thiagomacieira, line 283: This comment here is a symptom of the problem. irssi code should not need to figure this out: the resolver gives an order to try and we should try it.

src/core/network.c
```diff
@@ -392,95 +371,87 @@ int net_getsockname(GIOChannel *handle, IPADDR *addr, int *port)
...
+                                IPADDR *ip6)
+{
+	RESOLVED_IP_REC *iprec;
+
+	iprec = net_gethostbyname(addr, flags);
+	if (iprec->error == NULL) {
+		GList *curr;
+
+		for (curr = iprec->ailist; curr->next; curr = curr->next) {
```
horgh, line 434: Would we want this to be this instead?

```c
for (curr = iprec->ailist; curr; curr = curr->next) {
```

Wondering if we'd miss the last element right now.

src/core/servers.c
```diff
@@ -276,48 +272,47 @@ static void server_real_connect(SERVER_REC *server, IPADDR *ip,
...
+				/* retry resolve */
+				server_start_connect_resolve(server);
+				return;
+			}
+		}
+		if (curr != NULL) {
+			GInetAddress *addr;
+			addr = curr->data;
+			server->connrec->last_connected = i + 1;
+			ip = g_new0(IPADDR, 1);
```
horgh, line 311: Do we leak the memory allocated to `ip`? Maybe we need a free after calling `server_real_connect()` with it.

src/core/servers.c
```diff
@@ -326,28 +321,63 @@ static void server_connect_callback_readpipe(SERVER_REC *server)
...
+		server->connrec->resolved_host = NULL;
 		server_connect_failed(server, errormsg);
+
+		resolved_ip_unref(iprec);
+	}
+}
+
+static void server_connect_callback_resolved(RESOLVED_IP_REC *iprec, SERVER_REC *server)
+{
+	server->connect_cancellable = NULL;
```
horgh, line 345: I'm less confident about this one, but I'm wondering if we need to cancel and unref this.

src/core/network.c
```diff
@@ -81,27 +81,6 @@ int i_io_channel_read_block(GIOChannel *channel, void *data, int len)
 	return received < len ? -1 : 0;
 }
 
-IPADDR ip4_any = {
```
horgh, line 84: It looks like this is still used in dcc code. I get unknown command when running `/dcc` now, so I think the dcc module is failing to load. Not too sure about this one either :-).

src/core/network.c
```diff
@@ -392,95 +371,87 @@ int net_getsockname(GIOChannel *handle, IPADDR *addr, int *port)
...
+	GList *ailist;
+	GError *error;
+	GResolver *resolver;
+	RESOLVED_IP_REC *iprec;
 
-	g_return_val_if_fail(ip != NULL, -1);
-	g_return_val_if_fail(name != NULL, -1);
+#ifdef HAVE_CAPSICUM
+	if (capsicum_enabled())
+		return (capsicum_net_gethostbyname(addr, flags));
```
horgh, line 405: It looks like `capsicum_net_gethostbyname()` takes 3 params still, so will this compile in that case?

horgh (commented):
Looks like a nice improvement!

src/core/network.c
ailin-nemui: this is correct! oops :-)

src/core/network.c
ailin-nemui: yes! duh

src/core/servers.c
ailin-nemui: yes!

src/core/servers.c
ailin-nemui: yes! we need to unref it. I don't think we have to cancel it though since we do not want to cancel the operation

src/core/network.c
ailin-nemui: yes! it won't compile! I think I found a solution for this line of code (move it to `net_gethostbyname_first_ips`), unfortunately I cannot test it. ... but then I think there will be other places where net_gethostbyname_nonblock are used that will be a problem ...

PR #1581, opened 2025-08-27
feat: Add Sidepanels, Nick Alignment, and Active Window Whois

kofany:
This pull request introduces three new major features to enhance the user experience:

### 1. Native Sidepanels with Mouse Support

This feature adds native sidepanels for channel and user lists with full mouse support.

**Configuration:**

*   `sidepanel_left`: Enable/disable left sidepanel (default: `TRUE`)
*   `sidepanel_right`: Enable/disable right sidepanel (default: `TRUE`)
*   `sidepanel_left_width`: Width of left sidepanel (default: `18`)
*   `sidepanel_right_width`: Width of right sidepanel (default: `18`)
*   `sidepanel_right_auto_hide`: Auto-hide right panel when empty (default: `TRUE`)
*   `sidepanel_debug`: Enable debug logging (default: `FALSE`)

**Theme Format Settings:**

*   `/format sidepanel_header`
*   `/format sidepanel_item`
*   `/format sidepanel_item_selected`
*   `/format sidepanel_item_activity`
*   `/format sidepanel_item_events`
*   `/format sidepanel_item_highlight`
*   `/format sidepanel_item_nick_mention`
*   `/format sidepanel_item_query_msg`
*   `/format sidepanel_nick_op`
*   `/format sidepanel_nick_voice`
*   `/format sidepanel_nick_normal`

### 2. Nick Alignment and Truncation

This feature provides perfectly aligned nicknames for enhanced readability and consistent formatting across different chat windows.

**Configuration:**

*   `nick_column_enabled`: Enable/disable native nick column (default: `FALSE`)
*   `nick_column_width`: Set fixed width for nick column (default: `12`)

**Expandos:**

*   `$nickalign`: Align nicknames within specified width
*   `$nicktrunc`: Truncate nicknames that exceed column width

### 3. Whois in Active Window

This feature displays `whois` responses directly in the current chat window, keeping the workflow smooth and uninterrupted.

**Configuration:**

*   `print_whois_rpl_in_active_window`: Enable/disable whois responses in the active window (default: `FALSE`)

PR #1582, opened 2025-08-27
feat: Add option to print WHOIS replies in active window (#1)

kofany:
# Description
This pull request introduces a new feature that allows WHOIS replies to be printed directly in the active window.  
This helps maintain a smooth workflow by reducing the need to switch between windows.

It addresses the feedback from [PR #1581](https://github.com/irssi/irssi/pull/1581) and isolates the WHOIS functionality into a dedicated pull request.  
The code has also been formatted to match the project’s coding style.

# Configuration
- `print_whois_rpl_in_active_window`: Boolean setting to enable or disable this feature (default: `FALSE`).

# To enable
```irc
/set print_whois_rpl_in_active_window ON
```

### DIFF
--- a/src/fe-common/irc/fe-whois.c
+++ b/src/fe-common/irc/fe-whois.c
@@ -12,17 +12,25 @@
 
 #include <irssi/src/fe-common/core/printtext.h>
 
+/* When enabled, prints WHOIS replies into active window while preserving
+ * formatting and theming. */
+#define WHOIS_PRINT(server, nick, level, formatnum, ...)                                           \
+	do {                                                                                       \
+		if (settings_get_bool("print_whois_rpl_in_active_window") && active_win != NULL)   \
+			printformat_window(active_win, level, formatnum, __VA_ARGS__);             \
+		else                                                                               \
+			printformat(server, nick, level, formatnum, __VA_ARGS__);                  \
+	} while (0)
+
 static void event_whois(IRC_SERVER_REC *server, const char *data)
 {
 	char *params, *nick, *user, *host, *realname, *recoded;
 
 	g_return_if_fail(data != NULL);
 
-	params = event_get_params(data, 6, NULL, &nick, &user,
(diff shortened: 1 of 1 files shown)


vague666:
Thanks :) Can you fix the white space so it's the same as in the original? Because the white space fixes makes the PR messy to read

kofany:
@vague666 so Im little confused with those whitespaces as they was made after using clang format from your repo, my original code without this formatter is here https://github.com/irssi/irssi/commit/49f819ea8a09d4fe0ab4197fd02df7682239aff9 

So what exactly should I remove?

src/fe-common/irc/fe-whois.c
```diff
@@ -12,6 +12,16 @@
 
 #include <irssi/src/fe-common/core/printtext.h>
 
+/* When enabled, prints WHOIS replies into active window while preserving
+ * formatting and theming. */
+#define WHOIS_PRINT(server, nick, level, formatnum, ...) \
+	do { \
+		if (settings_get_bool("print_whois_rpl_in_active_window") && active_win != NULL) \
```
dwfreed, line 19: The setting value should be cached, as getting it for every print is not free.

src/fe-common/irc/fe-whois.c
```diff
@@ -406,6 +416,7 @@ static void event_whois_default(IRC_SERVER_REC *server, const char *data)
 void fe_whois_init(void)
 {
 	settings_add_bool("lookandfeel", "whois_hide_safe_channel_id", TRUE);
+	settings_add_bool("lookandfeel", "print_whois_rpl_in_active_window", FALSE);
```
dwfreed, line 419: I'm not a fan of the setting name.  Perhaps `whois_reply_active_window`

```suggestion
	settings_add_bool("lookandfeel", "whois_reply_active_window", FALSE);
```

dwfreed (changes requested):

Issue #31, opened 2014-06-15
Bug in the hilight code in combination with ACTIONS and color codes


incognico:
When you are hilighted in an ACTION and there are color codes before your nickname and there is a comma after your nickname, weird stuff happens: the text is grey after the hilighed part. I don't exactly remember if bold formatting plays a role but I remember that I discussed this issue with someone on IRC and he told me that this is a very ugly bug in the very tricky hilight code :-)

![capture](https://cloud.githubusercontent.com/assets/2098066/3280794/75c9ce76-f47e-11e3-9a19-d84e20877189.PNG)

ailin-nemui:
closing as duplicate of #95

Issue #95, opened 2014-07-13
hilight-text improperly restores previous color


Zannick:
The hilight algorithm inserts color control codes to turn the desired text into the hilight color. Unfortunately, in retrieving what it thinks is the previous color control code, it ignores several codes (such as ^o (15) which resets colors, similar to the control sequence ^dg which **is** properly handled). Thus, when in the presence of another script or a specially-formatted message (eg. from a bot that reports commits to a repo, giving different colors for username, branch, message, etc), it restores the previously used color.

ie.

^c5color^o plain Zannick should be plain
becomes
^c5color^o plain ^c8Zannick^c5 should be plain
where "should be plain" is colored red

Examples (where a script colors other users' nicks and irssi hilights mine): http://imgur.com/MdgxbMX,0SN8TuO

A simplistic way to fix this for ^o would be to expand the else case in strip_real_length (which is where irssi finds the last color control code) to something like

```
if (IS_COLOR_CODE(*str)) {
  *last_color_pos = (int)(str-start);
  *last_color_len = 1;
} else {
  [previous stuff...]
}
```

though this will still be wrong in the general case, since some of the control characters (eg. ^b ^f ^g ^v ^7) are toggles (and they do not remove the color from the hilight, either).

I'd suggest something more complicated where the toggles are tracked, and then hilight-text adds a ^o followed by the appropriate control characters:
^bbold Zannick still bold -> ^bbold ^c8Zannick^o^b still bold

(On the other hand, I fixed this for myself simply by adding a case to handle specifically ^o.)

hloeung:
Within scripts, is there a way to get the current color codes? It would be useful to get the current color codes, make the color/style changes, and then restore the color codes (cc @kot0dama).

kot0dama:
> Within scripts, is there a way to get the current color codes? It would be useful to get the current color codes, make the color/style changes, and then restore the color codes (cc @kot0dama).

I think there would be a way with `gui print text` signal, I'm close to being able to apply foreground and background color after a color change we do in the script. But then I'm not sure how to use $fg and $bg values in such a callback.

PR #1602, opened 2026-01-02
Restore the colours to the previous one instead of the theme default

### LINKED ISSUE #1592: Text style reset overrides theme colors
IRC clients typically clear all text formatting with `^O`, while in irssi this action also sets the line’s color to the theme’s `default_color` rather than the one applied before styling for that event (public/private message, quit/part, etc.), so it acts like `%N` instead of `%n`. This behavior can be reproduced even when `hide_colors` and `hide_text_style` are both enabled.

ailin-nemui:
Adds a history of maximum 10 colours which irssi can revert back to.

Fixes https://github.com/irssi/irssi/issues/1592


ailin-nemui:
@soulseller you're welcome to test it

@zannick I added ^O to strip_real_length like you suggested in #95 

@incognico see if this fixes issue #31

src/core/special-vars.c
```diff
@@ -27,6 +27,7 @@
 #include <irssi/src/core/signals.h>
 #include <irssi/src/core/special-vars.h>
 #include <irssi/src/core/utf8.h>
+#include <irssi/src/fe-common/core/formats.h>
```
horgh, line 30: Should we be depending on fe-common in core? Maybe we could move the constants if this is the right spot to be using them.

src/fe-common/core/formats.c
```diff
@@ -1237,17 +1392,17 @@ static const char *get_ansi_color(THEME_REC *theme, const char *str,
...
-	bg = bg_ret == NULL ? -1 : *bg_ret;
+	fg = -1;
+	bg = -1;
 
 	if (!i_isdigit(**str)) {
 		/* turn off color */
-		fg = -1;
-		bg = -1;
+		fg = 99;
+		bg = 99;
```
horgh, line 1405: What does 99 correspond to? I wonder if naming it would help. I see it's also in lines 1678 and 1682 below.

horgh (commented):
Cool!

PR #1604, opened 2026-01-24
make Irssi::Irc usable without dcc

ailin-nemui:
moved the Perl dcc glue to a separate module Irssi::Irc::Dcc, allowing to use Irssi::Irc with dcc unloaded


horgh (commented):
👍

PR #1605, opened 2026-01-24
add compatibility code for older GResolver

ailin-nemui:
IPv4-only flags were added in GLib 2.60

### DIFF
--- a/src/core/net-nonblock.c
+++ b/src/core/net-nonblock.c
@@ -26,6 +26,7 @@
 #include <irssi/src/core/net-nonblock.h>
 
 typedef struct {
+	GResolverNameLookupFlags flags;
 	NetGethostbynameContinuationFunc cont;
 	void *cont_data;
 } NET_GETHOSTBYNAME_CALLBACK_DATA;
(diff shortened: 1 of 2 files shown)


src/core/net-nonblock.c
```diff
@@ -39,7 +40,39 @@ static void net_gethostbyname_callback(GResolver *resolver, GAsyncResult *result
 	RESOLVED_IP_REC *iprec;
 
 	error = NULL;
+#if GLIB_CHECK_VERSION(2, 59, 0)
 	ailist = g_resolver_lookup_by_name_with_flags_finish(resolver, result, &error);
+#else
+	/* compatibility code for old GLib */
```
horgh, line 46: I wonder what distros have these old versions, as they seem quite old. All in-support ubuntu and debian versions have newer glib it looks like for example. Maybe bumping the version irssi requires would be an option? Anyway the change looks good!

ailin-nemui: yea... maybe we should up the GLib version requirement... I don't even know what the minimum required version is (the readme says 2.32, but idk if that still works). The version I tested is 2.58.3

PR #1609, opened 2026-01-25
run meson formatter

ailin-nemui:


horgh (approved):
Looks good!

Would running this in a github action to make sure things stay formatted make sense?

Would it make sense to run it on the root meson.build too?

horgh:
Oh I see the root meson.build did change. It changed again when I ran it locally so I thought it wasn't changed. My command must be wrong.

ailin-nemui:
no, I forgot to commit the root meson.build. also, a github action would be good...

horgh:
I added a github action in https://github.com/irssi/irssi/pull/1611!

PR #1610, opened 2026-01-25
Gio channel rename

ailin-nemui:
This is a huge rename from `handle` fields to `channel` and suffix\`ing of all `GIOChannel`-related functions with `_channel`. There should be no functional changes, otherwise that would be a bug. The purpose is to prepare for the migration of server connections to GSocketConnection/GIOStream, which will be done by the addition of a `stream` field and creating `_stream` and `_source` functions to deal with all kinds of GIOStream IO in a later PR

### DIFF
--- a/src/core/net-disconnect.c
+++ b/src/core/net-disconnect.c
@@ -19,6 +19,7 @@
 */
 
 #include "module.h"
+#include <irssi/src/core/net-sendbuffer.h>
 #include <irssi/src/core/network.h>
 
 #include <sys/select.h>
@@ -31,7 +32,7 @@
 
 typedef struct {
 	time_t created;
-	GIOChannel *handle;
+	GIOChannel *channel;
 	int tag;
 } NET_DISCONNECT_REC;
 
@@ -39,12 +40,18 @@ static GSList *disconnects;
 
 static int timeout_tag;
 
+void net_disconnect_any(NET_DISCONNECT_REC *rec)
+{
+	if (rec->channel != NULL)
+		net_disconnect_channel(rec->channel);
+}
+
 static void net_disconnect_remove(NET_DISCONNECT_REC *rec)
 {
 	disconnects = g_slist_remove(disconnects, rec);
 
 	g_source_remove(rec->tag);
-        net_disconnect(rec->handle);
+	net_disconnect_any(rec);
 	g_free(rec);
 }
 
@@ -57,7 +64,7 @@ static void sig_disconnect(NET_DISCONNECT_REC *rec)
 	   if server just keeps sending us stuff we won't get stuck */
 	count = 0;
 	do {
-		ret = net_receive(rec->handle, buf, sizeof(buf));
+		ret = net_receive_channel(rec->channel, buf, sizeof(buf));
 		if (ret == -1) {
 			/* socket was closed */
 			net_disconnect_remove(rec);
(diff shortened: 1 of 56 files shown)


src/core/net-sendbuffer.c
```diff
@@ -112,15 +115,16 @@ static int buffer_add(NET_SENDBUF_REC *rec, const void *data, int size)
...
+	int ret = 0;
 
 	g_return_val_if_fail(rec != NULL, -1);
 	g_return_val_if_fail(data != NULL, -1);
 	if (size <= 0) return 0;
 
 	if (rec->buffer == NULL || rec->bufpos == 0) {
                 /* nothing in buffer - transmit immediately */
-		ret = net_transmit(rec->handle, data, size);
+		if (rec->channel != NULL)
```
horgh, line 126: I wonder if this could be a behaviour change. Previously we'd end up setting `ret = -1` if `rec->channel` is `NULL` it looks like.

src/core/net-disconnect.c
```diff
@@ -92,15 +99,17 @@ static int sig_timeout_disconnect(void)
 
 /* Try to let the other side close the connection, if it still isn't
    disconnected after certain amount of time, close it ourself */
-void net_disconnect_later(GIOChannel *handle)
+void net_disconnect_later(NET_SENDBUF_REC *handle)
 {
 	NET_DISCONNECT_REC *rec;
 
 	rec = g_new(NET_DISCONNECT_REC, 1);
```
horgh, line 106: This is a pre-existing issue, but I wonder if this should be `g_new0`.

src/core/servers.c
```diff
@@ -195,25 +195,25 @@ static void server_connect_callback_init_ssl(SERVER_REC *server, GIOChannel *han
...
+	g_return_if_fail(ip != NULL || unix_socket != NULL || host != NULL);
 
 	if (ip != NULL) {
 		server->connrec->chosen_family = ip->family;
 		net_ip2host(ip, ipaddr);
 		server->connrec->ipaddr = g_strdup(ipaddr);
 	}
 
-	signal_emit("server connecting", 2, server, ip);
+	signal_emit("server connecting", 3, server, ip, ip != NULL ? ipaddr : host);
```
horgh, line 216: Is having an additional argument okay?

src/core/net-disconnect.c
horgh: It might actually be relevant to this PR given we conditionally don't set the fields now, so they'd be uninitialized.

src/core/net-disconnect.c
```diff
@@ -136,15 +145,16 @@ void net_disconnect_deinit(void)
...
-                fd = g_io_channel_unix_get_fd(rec->handle);
-		FD_ZERO(&set);
-		FD_SET(fd, &set);
-		tv.tv_sec = first ? 0 : max-now;
-		tv.tv_usec = first ? 100000 : 0;
-		if (select(fd+1, &set, NULL, NULL, &tv) > 0 &&
-		    FD_ISSET(fd, &set)) {
-			/* data coming .. check if we can close the handle */
-			sig_disconnect(rec);
+		if (rec->channel != NULL) {
```
horgh, line 148: Hmm, I wonder if this one could be a behaviour change too. Previously the else if was connected to the `select`.

src/irc/core/irc-servers.c
```diff
@@ -695,23 +695,24 @@ void irc_server_send_away(IRC_SERVER_REC *server, const char *reason)
...
-	/* A bit kludgy way to do the flood protection. In ircnet, there
-	   actually is 1sec / 100 bytes penalty, but we rather want to deal
-	   with the max. 1000 bytes input buffer problem. If we send more
-	   than that with the burst, we'll get excess flooded. */
-	if (len < 100 || server->cmd_queue_speed <= 10)
-		server->wait_cmd = 0;
-	else {
-		server->wait_cmd = server->last_cmd;
-		server->wait_cmd += (2 + len / 100) * G_USEC_PER_SEC;
+	if (server->handle != NULL) {
```
horgh, line 698: I'd probably return early here if it was null, saving a level of nesting!

PR #1625, opened 2026-07-01
irc-nicklist: avoid reading past end of parsed token in event_userhost

hyder365:
When processing RPL_USERHOST (302), the code finds '=' in the host string, replaces it with NUL, advances ptr past it, and dereferences *ptr to check for '-'. If '=' is the last character in the string (e.g. 'nick=' sent by a malicious server), ptr advances to the NUL terminator. While this is still within the allocated buffer, it is defensively safer to guard the '-' check with a NUL check on *ptr first.

### DIFF
--- a/src/irc/core/irc-nicklist.c
+++ b/src/irc/core/irc-nicklist.c
@@ -531,7 +531,7 @@ static void event_userhost(SERVER_REC *server, const char *data)
 			oper = 0;
 		*ptr++ = '\0';
 
-		nicklist_update_flags(server, *pos, *ptr == '-', oper);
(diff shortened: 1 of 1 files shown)


dgl:
> While this is **still within the allocated buffer**, it is defensively safer to guard the '-' check with a NUL check on *ptr first.

*defensively safer* ?!

It's basic maths to remove the expression you've added here. In case you don't believe me I went to the trouble of checking that gcc does indeed optimise this out: https://godbolt.org/z/j7v48xe1v

This is pointless, please don't waste people's time. Particularly if this is an LLM assisted report, use your own brain.

PR #1626, opened 2026-07-01
irc-nicklist: initialize nickrec->hops when sscanf fails in fill_who

hyder365:
When fill_who() is called from event_whox_channel_full, the hops parameter may be an empty string if the server sent a malformed WHOX reply with fewer tokens than expected. sscanf() on an empty string returns 0 (no items matched), leaving nickrec->hops at its prior value rather than a clean default.

Fix: check the sscanf return value and set hops to 0 if parsing failed.


dgl:
While checking the return of sscanf is probably a good idea, this isn't going to change anything as the value is already 0. `g_new0` is used to allocate `NICK_REC` in the IRC code.

PR #1627, opened 2026-07-06
ctcp: cap CTCP body length used in dynamic signal name (IRC path)

acts-1631:
The body of a CTCP message received over IRC (PRIVMSG/NOTICE) is concatenated into a dynamic signal name and emitted via signal_emit in ctcp_msg/ctcp_reply (ctcp.c) and the DCC sub-dispatch ctcp_msg_dcc/ctcp_reply_dcc (dcc.c). The first whitespace-separated token becomes the CTCP command and is interned as a signal name; the remainder is delivered to any matching handler, including loaded perl scripts.

signal_emit interns unknown signal names permanently in the global signal-name registry (module_get_uniq_id_str g_strdup's the name into two hash tables that are freed only at exit). With no length or count cap on the IRC-CTCP path, a remote user sending many distinct CTCP command names grows the registry without bound, exhausting memory and crashing the victim's irssi.

### DIFF
--- a/src/irc/core/ctcp.c
+++ b/src/irc/core/ctcp.c
@@ -234,6 +234,15 @@ static void ctcp_msg(IRC_SERVER_REC *server, const char *data,
 	if (ignore_check(SERVER(server), nick, addr, target, data, MSGLEVEL_CTCPS))
 		return;
 
+	/* Cap the CTCP body length that is concatenated into the signal name.
+	 * The body comes from a remote user over IRC (PRIVMSG) and is passed
+	 * as the second argument to any matching signal handler, including
+	 * loaded perl scripts. A modest cap limits the attack surface for the
+	 * dynamic-signal-name dispatch and aligns with the convention used for
+	 * CTCP-over-DCC-chat (dcc-chat.c). */
+	if (strlen(data) > 256)
+		return;
+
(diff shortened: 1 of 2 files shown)


vague666:
Thank you for your effort. How much of this is LLM assisted?

acts-1631:
All the bugs so far were found with LLMs and a custom multi-step audit process.

vague666:
> With no length or count cap on the IRC-CTCP path, a remote user sending many distinct CTCP command names grows the registry without bound, exhausting memory and crashing the victim's irssi.

This PR doesn't solve the issue sadly. The implementation would maybe need a list of valid commands or some other mechanism. Just limiting the command length isn't enough.

acts-1631:
Indeed I think you're right. I can work on a more complete fix but I want to get some of the other ones in first if we can. Right now I'm kinda piling up branches and it's getting disorganized.

dgl:
I don't think this PR does anything useful, an IRC message is already capped on RFC compliant networks at around 480-490 characters (512 - lengths of command, sender's hostmask and target's name). This just limits any potential memory leak to about half that. #1632 could fix this entire class of issue instead, so I recommend this particular PR is just closed, there's no value in trying to patch this.

vague666:

PR #1628, opened 2026-07-06
otr: fix three bugs in multi-fragment message reassembly

acts-1631:
enqueue_otr_fragment() in src/otr/otr.c reassembles ?OTR: messages that
arrive split across multiple PRIVMSGs (bitlbee/znc and some ircds do
this). It's reachable for any private message a peer sends, with no
OTR session required, since sig_message_private feeds every query
PRIVMSG straight into it. Three separate bugs there:

1. Empty fragment causes a size_t underflow (msg_len - 1 wraps to
   SIZE_MAX) and reads off the page, crashing irssi. An empty PRIVMSG
   body reaching this path is enough to trigger it.

2. The buffer growth check uses '>' instead of '>=' when deciding
   whether to realloc. When a fragment's length exactly matches the
   remaining space, no realloc happens, but the NUL terminator is
   still written one byte past the end of the buffer, corrupting the
   heap. Fragment lengths and the running remaining-space counter are
   both attacker controlled, so this is reachable remotely too.

3. There's no cap on the total reassembled size, so a peer can open a
   reassembly and never send the terminating '.', growing the buffer
   with every PRIVMSG until the process gets OOM-killed.

Each commit fixes one of these independently. Added a 256 KiB cap for
#3, well above any real OTR message. Existing test suite passes, and I
tested each bug/fix manually with ASan against the reassembly function
in isolation (crashes without the fix, no crash with it, and normal
multi-fragment reassembly still reconstructs the message correctly).

### DIFF
--- a/src/otr/otr.c
+++ b/src/otr/otr.c
@@ -605,6 +605,15 @@ static enum otr_msg_status enqueue_otr_fragment(const char *msg, struct otr_peer
 	/* We are going to use it quite a bit so ease our life a bit. */
 	msg_len = strlen(msg);
 
+	/* An empty fragment carries no data and no end tag. Without this
+	 * guard, the end-tag check on the open-reassembly path indexes
+	 * msg[msg_len - 1]; when msg_len == 0 (size_t) this underflows to
+	 * SIZE_MAX and reads off the page, crashing irssi. */
+	if (msg_len == 0) {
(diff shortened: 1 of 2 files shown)


vague666:
@ahf mind looking at this since you wrote the otr module?

vague666:

acts-1631:
As long as the issues get fixed, it's ok if you don't want to take the pull requests. These are real bugs. I emailed staff@irssi.org and got no reply.

PR #1629, opened 2026-07-06
irc-proxy: guard NULL chatnet in CTCP forwarding

acts-1631:
sig_server_event() in src/irc/proxy/listen.c forwards CTCP PRIVMSGs to
proxy clients that have CTCP forwarding enabled, using
strstr(rec->proxy_address, server->connrec->chatnet). If the server
connection has no chatnet (e.g. connected via /server without a
network definition, which is a normal thing to do), connrec->chatnet
is NULL and strstr() dereferences it, crashing irssi. Any remote user
can trigger it by sending the victim a CTCP.

connrec->chatnet is nullable in general, and event_connected() in the
same file already guards the same field at an equivalent strstr() call
a bit further down. This path was just missing that guard.


vague666:

PR #1632, opened 2026-07-09
signals: avoid interning unknown emitted names

acts-1631:
Unknown remote-controlled signal names are interned by `signal_emit()` even when no handlers have registered for them. A private name index for active signals lets `signal_emit()` return `FALSE` without creating an ID for unknown input.

This preserves CTCP fallback handling and also covers CAP and other dynamic emitters.

Prepared with Codex assistance and reviewed by the contributor.

### DIFF
--- a/src/core/signals.c
+++ b/src/core/signals.c
@@ -46,6 +46,7 @@ typedef struct {
 void *signal_user_data;
 
 static GHashTable *signals;
+static GHashTable *signals_by_name;
 static Signal *current_emitted_signal;
 static SignalHook *current_emitted_hook;
 
@@ -65,6 +66,7 @@ static int signal_unref_full(Signal *rec, int remove)
 			signal_get_id_str(rec->id));
 	}
 
+	g_hash_table_remove(signals_by_name, signal_get_id_str(rec->id));
 	if (remove)
 		g_hash_table_remove(signals, GINT_TO_POINTER(rec->id));
         g_free(rec);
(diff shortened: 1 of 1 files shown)


dgl:
Don't open multiple pull requests for the same underlying issue, this just fragments the discussion. (Particularly with a random commit from another PR included, this isn't helpful to anyone.)

I think this PR does hint at a better approach, but it breaks the encapsulation of signals. `signal_emit` has nothing in its API contract that returns the signal ID; so it should be able to do similar within the signals code and not have this be CTCP specific.

There is already #1624 which also covers the CAP response and has a strangely identical commit for the CTCP part. Fixing the underlying issue in `signal_emit` definitely feels like a better approach as it would solve all these reports in one place.

(I'm guessing this is two people using the same LLM model? Please disclose your LLM use up front, I know you're trying to help here, but there's not many contributors here and LLMs are very good at generating a lot of noise and questionable code.)

acts-1631:
Thanks, I traced the overlap and agree the correct scope is `signal_emit()` itself. I’ve replaced the two CTCP-specific mitigation commits with one central fix: `signal_emit()` now looks up an existing signal ID and returns `FALSE` for an unknown name instead of interning it. This preserves the existing fallback behavior and covers CTCP, CAP, and other dynamic emitters without exposing signal-registry details to their callers.

src/core/signals.c
```diff
@@ -278,7 +278,9 @@ int signal_emit(const char *signal, int params, ...)
 
 	g_return_val_if_fail(params >= 0 && params <= SIGNAL_MAX_ARGUMENTS, FALSE);
 
-	signal_id = signal_get_uniq_id(signal);
+	signal_id = module_find_uniq_id_str("signals", signal);
```
dgl: Maybe add a wrapper macro like https://github.com/irssi/irssi/blob/master/src/core/signals.h#L70-L71 -- then we again don't leak encapsulation.

acts-1631:
Updated the PR with a signals-internal lookup, so unknown emitted names no longer create IDs.

dgl:
That is not what I meant, the previous version was basically there but needed a macro to wrap `module_find_uniq_id_str`. Given the lack of understanding now shown with the latest PR I can't really trust this PR as anything other than reporting the issue.

vague666:

PR #1633, opened 2026-07-09
dcc: enforce declared receive size

acts-1631:
DCC GET trusts the sender-provided file size for UI display and auto-get policy, but continues writing data until the peer closes the connection. A sender can advertise a small file and exhaust the receiver's disk.

Limit each read to the remaining declared byte count and close the transfer once the full size has been received.


vague666:

PR #1634, opened 2026-07-09
irc-proxy: expire unauthenticated clients

acts-1631:
The IRC proxy allocates a client record before PASS is processed. Silent peers can keep unauthenticated connections open indefinitely, even when a proxy password is configured, exhausting file descriptors and memory.

Add a configurable authentication timeout (default 1 minute, via irssiproxy_timeout setting) and cancel it after successful registration.


vague666:

PR #1635, opened 2026-07-09
dcc-server: expire idle accepted clients

acts-1631:
DCC SERVER accepts and retains every peer connection, while the generic DCC timeout explicitly excludes server records. Silent peers can exhaust file descriptors and memory.

Apply the existing DCC timeout to accepted DCC SERVER clients until they send their first protocol line.


vague666:

PR #1636, opened 2026-07-29
Typo Fix: manual.txt Chapter 5.1

jremes-foss:
Good afternoon.

While studying through Irssi documentation and code, I noticed a small typo in Chapter 5.1.

I decided to fix it.

Thank you.

PR #1637, opened 2026-07-29
Bug fix: Solaris VM environment GCC15 Keyword Check

jremes-foss:
My earlier typo fix in #1636 had a failed pipeline due to Solaris VM.

I investigated the issue:
- Looks like the Solaris VM environment is using GCC 15.
- GCC 15 defaults to C23 mode.
- In standard C23, `bool` is now a built-in standard than a macro or a typedef.

To solve this issue, I switched the pipeline to older standard (gnu11). 

This should make the Solaris pipeline pass.

Thank you.


vague666:
We need to verify that change doesn't affect anything else but thank you :)

jremes-foss:
@vague666 You're welcome. :)

dgl (approved):
It's strange the header files (from CSW?) aren't compatible with the default C std the compiler uses. Maybe meson isn't using the right include paths or something (https://www.opencsw.org/manual/for-developers/index.html).

However given this is a solaris workflow specific change this seems fine as a workaround for now and is better than every PR being red.

jremes-foss:
Thank you for approval @dgl. I do not have authorisation to push to main, so please merge when you're able to do so.

PR #1639, opened 2026-08-23
Answer IRCv3 CAP commands to fix client connect hang in the proxy module

hloeung:
When running `irssi` as a proxy, then having another `irssi` instance connect to that `irssi` proxy, the `irssi` instance hangs on `Waiting for CAP LS response...`. This change fixes this by answering to CAP LS with an empty list and NAK REQs so capability negotiation can finish cleanly.

On connecting to the `irssi` acting as a proxy, we now get this:

```
|08:48 -!- Irssi: Connecting to localhost [127.0.0.1] port 6669
|08:48 Waiting for CAP LS response...
|08:48 -!- Irssi: Connection to localhost established
```

### DIFF
--- a/src/irc/proxy/listen.c
+++ b/src/irc/proxy/listen.c
@@ -164,6 +164,44 @@ static void grab_who(CLIENT_REC *client, const char *channel)
 	g_string_free(arg, TRUE);
 }
 
+static void handle_client_cap(CLIENT_REC *client, const char *args)
+{
+	char *subcmd, *cap_args;
+	const char *target_nick;
+
+	subcmd = g_strdup(args);
+	cap_args = strchr(subcmd, ' ');
+	if (cap_args != NULL) {
+		*cap_args++ = '\0';
+		while (*cap_args == ' ')
+			cap_args++;
+	} else {
+		cap_args = "";
+	}
+
+	target_nick = client->nick != NULL ? client->nick : "*";
+
+	if (g_ascii_strcasecmp(subcmd, "LS") == 0) {
+		proxy_outdata(client, ":%s CAP %s LS :\r\n", client->proxy_address, target_nick);
+	} else if (g_ascii_strcasecmp(subcmd, "REQ") == 0) {
+		const char *req = cap_args;
+
+		while (*req == ' ' || *req == ':')
+			req++;
+
+		if (*req != '\0') {
+			proxy_outdata(client, ":%s CAP %s NAK :%s\r\n", client->proxy_address,
+			              target_nick, req);
+		}
+	} else if (g_ascii_strcasecmp(subcmd, "LIST") == 0) {
+		proxy_outdata(client, ":%s CAP %s LIST :\r\n", client->proxy_address, target_nick);
+	} else if (g_ascii_strcasecmp(subcmd, "END") == 0) {
+		/* CAP negotiation complete */
+	}
+
+	g_free(subcmd);
+}
+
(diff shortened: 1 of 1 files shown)


hloeung:
Fixed the `clang-format` failures, but the Solaris VM one still exists. It might be fixed by https://github.com/irssi/irssi/pull/1637

src/irc/proxy/listen.c
```diff
@@ -164,6 +164,48 @@ static void grab_who(CLIENT_REC *client, const char *channel)
...
+		const char *req = cap_args;
+
+		if (*req == ':')
+			req++;
+
+		if (g_ascii_strcasecmp(req, "message-tags") == 0) {
+			proxy_outdata(client, ":%s CAP %s ACK :message-tags\r\n",
+			              client->proxy_address, target_nick);
+		} else {
+			proxy_outdata(client, ":%s CAP %s NAK :%s\r\n", client->proxy_address,
```
dgl, line 193: This NACKs an empty list, which seems wrong.

dgl (commented):
I think more is needed to properly support CAP and inparticular `message-tags` via the proxy module.

Currently `sig_server_event` will send the server's line on verbatim, which depending what capabilities Irssi has negotiated could not match what the client expects. However the proxy intercepts CTCP, so those will then lose the tags.

Just claiming to negotiate the capability isn't really quite right.

> When running irssi as a proxy, then having another irssi instance connect to that irssi proxy, the irssi instance hangs on Waiting for CAP LS response....

Isn't the actual bug here that the proxy module in `handle_client_connect_cmd` doesn't reply with anything if the command isn't recognised?

Possibly part of a >20 year old attempt at security through obscurity, as the proxy doesn't make it obvious what it is until you get the `PASS` right.

It's amusing Irssi is being caught out by its own spec violation (the [capability negotiation spec](https://github.com/ircv3/ircv3-specifications/blob/master/extensions/capability-negotiation.md#connection-registration) says you should send CAP LS + NICK + USER together, rather than trying to figure out via numerics that the server doesn't support `CAP`, there's already a hack for another ircd here: https://github.com/irssi/irssi/blob/43f1727ef9fe1cb51d5031ee2dd0e1965217e4cb/src/irc/core/irc-servers.c#L317)

dgl:
So there's two things here:

* More IRCv3 support in the proxy and what made me actually look at this, as the title claimed. That might be cool, although I don't know whether many people use the proxy anymore...

* Otherwise the `CAP` behaviour is actually noted here:
https://github.com/irssi/irssi/blob/master/NEWS#L149 with a workaround. It mentions STARTTLS as the reason for that and IRCv3 has deprecated the tls extension (https://ircv3.net/specs/deprecated/tls) so that could be revisited potentially.

hloeung:
Apologies for claiming to support IRCv3 CAP `message-tags` when it did not, and probably not even close.

I've updated the PR so instead, answer or reply to a `CAP LS` with an empty list of capabilities supported to allow the negotation to complete and for the connection to be established. Would this work?

Thanks and appreciate the time taken to review this change!

dgl:
I would expect at least that the proxy would strip CAPs it has negotiated with the server if the client doesn't support them.

Thanks for the work, but I don't think there's value in merging this as-is, because it's no better than connecting with `-nocap`. #1643 has also been filed which can track the work to make proxy more CAP aware.

hloeung:
Okay then, thanks for your time with this one.

